PR #28, opened 2015-03-11
Add TEEC_ErrorOrigin typedef

emontellese:
Add TEEC_ErrorOrigin typedef, which is referred to in the TEEC_ORIGIN_XX comments

Greetings!
My name is Eric Montellese, and I'm working on integrating OPTEE for an ST project.  Noticed this small omission.

Thanks!
Eric

### DIFF
--- a/public/tee_client_api.h
+++ b/public/tee_client_api.h
@@ -232,6 +232,7 @@
 #define TEEC_PARAM_TYPE_GET(p, i) (((p) >> (i * 4)) & 0xF)
 
 typedef uint32_t TEEC_Result;
+typedef uint32_t TEEC_ErrorOrigin;
 
 /**
  * struct TEEC_Context - Represents a connection between a client application


emontellese:
TEEC_Operation is missing comments for 'memRefs' and 'flags' as well.  Could you add this and/or send me what those parameters are used for?

jenswi-linaro:
Hi Eric,

The comment in the header file is wrong. TEE Client API Specification doesn't define this type. You can find a link to this specification at http://www.globalplatform.org/specificationsdevice.asp

Regards,
Jens

emontellese:
Thank you, Jens.  I'll read through that.

emontellese:
I see some other differences (for example the missing 'imp' container-type variable for the implementation-specific parts).  Is this intentional, or due to a spec update?  (looking to determine what I should code towards, since we do not yet have the official release - trying to be sure i have everything covered).  Thanks!

emontellese:
thanks, btw, that document you pointed me to is immensely helpful!

jenswi-linaro:
The imp part of the different structs are implementation specific, if it exists or not doesn't matter, portable code can't make any assumptions about it.

emontellese:
I guess I read the spec that if there were implementation-specific parts to the structs that they would go under an 'imp' container.  But, regardless, I understand now what you're saying and that these fields can be ignored however they're listed.  Thank you.

emontellese:
One other note:  is it possible to change the spec and optee header to use a named struct for the various typedefs?  With the unnamed struct, a forward declaration of the struct is not possible.
So, like this:

struct Foo;     /* Forward declaration */
typedef struct Foo /* The name is needed here */
{
} Foo; /* Definition and typedef combined */

Discussion of this problem here:
http://stackoverflow.com/questions/7256436/forward-declarations-of-unnamed-struct

pascal-brand-st-dev:
Unfortunately, this is not possible to change the Global Platform specifications.

emontellese:
Ok, understood.

One other minor note: I see the GP Spec has a void type for TEEC_FinalizeContext() but OPTEE has a TEEC_Result

pascal-brand-st-dev:
Right. We'll fix it.

emontellese:
Cool -- thanks!

pascal-brand-st-dev:
Eric, can we close this pull-request?

emontellese:
Sure -- thanks!

PR #33, opened 2015-07-07
Storage: Added create hard-link operation

m943040028:
This patch should work with the commit in optee_os:

a5c22091cd5aaee9a8da2d87f6937a46714e7a11

Signed-off-by: SY Chiu <sy.chiu@linaro.org>
Tested-by: SY Chiu <sy.chiu@linaro.org> (QEMU)


pascal-brand-st-dev:
I would change the commit message
- remove the sha1, as this will change when rebasing
- explain that link is required for atomic operation from optee_os

    Reviewed-by: Pascal Brand <pascal.brand@linaro.org>

PR #45, opened 2016-02-04
RPMB: fix read_cid() and data access with a real eMMC device

jforissier:
Prior to this commit, the RPMB code was tested in emulation mode only
(RPMB_EMU=1). This fixes a couple of issues found when testing with a
real eMMC device on a HiKey board. The eMMC is a Hardkernel 8GB module
plugged into the microSD port. Kernel is from Linaro's 96boards
repository [1], slightly modified so that the RPMB can be used on the
microSD connector (MMC_CAP_CMD23 is added to the capability flags of
the second MMC controller).

[1] https://github.com/96boards/linux branch hikey commit 54e6524885a9
("Merge pull request #178 from xin3liang/hikey-3.18-720p-boot-issue").

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -37,6 +37,8 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <stdbool.h>
+#include <stdlib.h>
+#include <unistd.h>
 
 #ifdef RPMB_EMU
 #include <stdarg.h>
@@ -110,7 +112,6 @@ struct rpmb_data_frame {
 
 /* mmc_ioc_cmd.opcode */
 #define MMC_SEND_EXT_CSD		 8
-#define MMC_SEND_CID			10
 #define MMC_READ_MULTIPLE_BLOCK		18
 #define MMC_WRITE_MULTIPLE_BLOCK	25
 
@@ -121,7 +122,6 @@ struct rpmb_data_frame {
 #define MMC_RSP_OPCODE	(1 << 4)	/* Response contains opcode */
 
 #define MMC_RSP_R1      (MMC_RSP_PRESENT|MMC_RSP_CRC|MMC_RSP_OPCODE)
-#define MMC_RSP_R2      (MMC_RSP_PRESENT|MMC_RSP_136|MMC_RSP_CRC)
 
 #define MMC_CMD_ADTC	(1 << 5)	/* Addressed data transfer command */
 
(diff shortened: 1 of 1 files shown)


tee-supplicant/src/rpmb.c
```diff
@@ -144,17 +151,64 @@ static int mmc_rpmb_fd(uint16_t dev_id)
 		fd = open(path, O_RDWR);
 		if (fd < 0) {
 			EMSG("Could not open %s (%s)", path, strerror(errno));
-			return -1;
+			return TEEC_ERROR_ITEM_NOT_FOUND;
```
jenswi-linaro: OK to rely on TEEC_ERROR-code happen to map as negative ints?

jforissier: Hmmm, no, it's definitely not a good idea and what's more completely useless here. I'll fix.

jforissier:
Updated.

tee-supplicant/src/rpmb.c
```diff
@@ -583,9 +623,19 @@ static uint32_t rpmb_data_req(int fd, struct rpmb_data_frame *req_frm,
 
 		/* Send write request frame(s) */
 		cmd.write_flag |= MMC_CMD23_ARG_REL_WR;
+		/*
+		 * Black magic: tested on a HiKey board with a HardKernel eMMC
```
pascal-brand-st-dev: Can `Black magic` be understood by non-french people :smile: ? Can it be replaced with `Hack`

tee-supplicant/src/rpmb.c
```diff
@@ -641,17 +691,34 @@ static uint32_t rpmb_data_req(int fd, struct rpmb_data_frame *req_frm,
...
+
 	info->rel_wr_sec_c = ext_csd[222];
 	info->rpmb_size_mult = ext_csd[168];
 	info->ret_code = RPMB_CMD_GET_DEV_INFO_RET_OK;
 
-	return TEEC_SUCCESS;
+err:
+	if (fd) {
+		/* 0 means emulation mode */
+		close(fd);
```
pascal-brand-st-dev: Can't we have `fd==0` in non-emulation mode too?

tee-supplicant/src/rpmb.c
jforissier: I think it is pretty much universal:
https://en.wikipedia.org/wiki/Black_magic
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/log/?id=master&qt=grep&q=black+magic

All things considered, this is in fact "white magic", because there is no evil purpose here... (a bit of selfishness perhaps?)
;-)

tee-supplicant/src/rpmb.c
jforissier: No, because file descriptor 0 is always standard input. Unless tee-supplicant closes all its file descriptors on startup (0, 1, 2), it is impossible to have `fd == 0` here.

pascal-brand-st-dev: I would prefer

    #ifndef RPMB_EMU
    close(fd);
    #endif

or a new function `rpmb_close_mmc` which close(fd) or does nothing depending on RPMB_EMU

jforissier: OK, I will add a new function.

jforissier:
Updated.

pascal-brand-st-dev:
Reviewed-by: Pascal Brand <pascal.brand@linaro.org>

jforissier:
@jenswi-linaro are you OK with this updated patchset?

jenswi-linaro:
Looks good.

jforissier:
Thanks, commits squashed and review tags added.

PR #49, opened 2016-04-25
tee-supplicant: prevent double load

vchong:
Tested-by: Victor Chong <victor.chong@linaro.org> (HiKey)

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -360,11 +360,79 @@ out:
 	OUTMSG();
 }
 
+/*
+ * http://stackoverflow.com/questions/6898337/determine-programmatically-if-a-program-is-running
+ */
+static int proc_find(const char* name)
+{
+	DIR* dir;
+	struct dirent* ent;
+	char* endptr;
+	char buf[512];
+	int found = 0;
+
+	if (!(dir = opendir("/proc")))
+	{
+		EMSG("can't open /proc");
+		return -1;
+	}
+
+	while((ent = readdir(dir)) != NULL)
+	{
+		/*
+		 * if endptr is not a null character, the directory is not
+		 * entirely numeric, so ignore it
+		 */
+		long lpid = strtol(ent->d_name, &endptr, 10);
+		if (*endptr != '\0')
+		{
+			continue;
+		}
+
+		/* try to open the cmdline file */
+		snprintf(buf, sizeof(buf), "/proc/%ld/cmdline", lpid);
+		FILE* fp = fopen(buf, "r");
+
+		if (fp)
+		{
+			if (fgets(buf, sizeof(buf), fp) != NULL)
+			{
+				/* check the first token in the file, the program name */
+				char* first = strtok(buf, " ");
+				if (!strcmp(first, name))
+				{
+					found++;
+
+					if (found == 2)
+					{
+						fclose(fp);
+						closedir(dir);
+						return found;
+					}
+				}
+			}
+			fclose(fp);
+		}
+	}
+
+	closedir(dir);
+	return found;
+}
+
 int main(int argc, char *argv[])
 {
 	int fd;
 	union tee_rpc_invoke request;
 	int ret;
+	int num_proc;
+
+	/* If tee-supplicant already running, let user know and exit without loading. */
+	num_proc = proc_find(argv[0]);
+	if (num_proc == 2)
+	{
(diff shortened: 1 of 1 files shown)


pascal-brand-st-dev:
Having 2 supplicants opened at the same time, but on different TEE, is a valid use-case. Is it possible to check the arguments of the instances of the tee-supplicant to tell it is really called twice or not?

tee-supplicant/src/tee_supplicant.c
```diff
@@ -360,11 +360,79 @@ static void process_rpmb(union tee_rpc_invoke *inv)
...
+		snprintf(buf, sizeof(buf), "/proc/%ld/cmdline", lpid);
+		FILE* fp = fopen(buf, "r");
+
+		if (fp)
+		{
+			if (fgets(buf, sizeof(buf), fp) != NULL)
+			{
+				/* check the first token in the file, the program name */
+				char* first = strtok(buf, " ");
+				if (!strcmp(first, name))
```
jenswi-linaro: This will not work if tee-supplicant is started as `tee-supplicant` in one case and as `/bin/tee-supplicant` in the other case.

jforissier:
Checking arguments seems a bit fragile. IMO that should be the driver that returns an error if it already has a supplicant process bound for a given TEE. It also would also prevent a TOCTOU bug.

jenswi-linaro:
This is a bit of a poor mans solution, even with the enhancement proposed by Pascal. A proper solution would need some support in the kernel API.

jbech-linaro:
It seems quite commit that daemons etc writes a "pid" file (like /var/run/tee-supplicant.pid) when first created and then subsequent runs checks if that file exists.

jforissier:
@jbech-linaro you're right, and that would be useful for the "stop" script, too. Beware of Android specific aspects, though (is `/var/run/` the proper path for Android? Is the file locking API (`fcntl(F_SETLK)`) available?. But I think it is a good solution.

jenswi-linaro:
If we going down that path, aren't the init scripts supposed to keep track of which daemons are already started and not? An advantage of not dealing with this in tee-supplicant is that it will not be required to run as root.

jforissier:
@jenswi-linaro yes, but we would depend on the init scripts doing the right thing, i.e. this is more or less self-contained. With the pid file, the daemon has all the knowledge it needs to exit if another instance is already running.

As for having tee-supplicant not running as root: I think it would be both a good and a bad thing. Good because the less privileges the better, but still it would bring extra security if we can prevent any user process from binding to the supplicant device. So, I think `/dev/teeprivX` should be open by root only, then tee-supplicant should drop its root privileges and run as a regular user once it has obtain a file descriptor to the device.

So that would be:
- Start as root
- Do the pid file stuff
- Open the TEE supplicant device
- Drop root privileges

jenswi-linaro:
With not running as root I didn't mean running as any uid, it's not uncommon to have designated uids for some daemons. What you suggests will work, but we'll need to drop privileges to an uid that still may read and write the secure storage files.

Long term there's one more thing to address, it would be preferable if tee-supplicant was started when the instance of a suitable driver is announced with an uevent.

jforissier:
@jenswi-linaro I didn't mean *any* uid either ;-)

uevent? Sounds good, although I don't know much about this technique. I would still recommend the .pid file as a quick fix.

jenswi-linaro:
@jforissier, I agree, let's do the quick fix first. The other requires quite a lot of investigation before it can be implemented.

d3zd3z:
/var/run won't work at all on Android. I really don't think this code should be in the supplicant at all. Should we not, instead, make the teeprivx driver only allow a single client?

jforissier:
@d3zd3z that was my initial proposal, but didn't Jens say it would require some significant work? Or I misunderstood?

d3zd3z:
@jforissier it probably is some work. I think it would be better to leave things as is, though, than add a fragile /proc hack. Programs that do this kind of thing tend to fail in unexpected places.

jenswi-linaro:
Enforcing only one open file on /dev/teeprivX isn't that hard to implement. I suppose we'd like to do it in the OP-TEE driver instead of in the framework to be flexible.

What's a bit more complicated is auto starting based on uevents, but that can be done at a later stage and should not require changes in the driver.

jforissier:
@d3zd3z I never recommended doing anything with `/proc` ;)
My preferences were (and still are): 1. Have the driver reject the second open, 2. Use a properly locked PID file like most daemons (used to) do.

PR #50, opened 2016-05-09
SQLFS

jforissier:
Adding support for a file system in a SQLite database. Will allow simplifying the REE-based Trusted Storage implementation  in OP-TEE, since SQLite provides atomicity thanks to its transaction-oriented interface.


Makefile
```diff
@@ -14,6 +14,8 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 
+WITH_SQLFS ?= y
```
pascal-brand-st-dev: Missing the update of `Android.mk`

libsqlite3/COPYING
```diff
@@ -0,0 +1,2 @@
+SQLite is public domain
```
pascal-brand-st-dev: Update `LICENSE`

Makefile
```diff
@@ -126,11 +126,11 @@ copy_export: build
 	cp public/*.h ${EXPORT_DIR}/include
 
 ifeq ($(WITH_SQLFS),y)
-.PHONY: build-libsqlite3
+.PHONY: build-libsqlite3 build-libsqlfs
```
pascal-brand-st-dev: Update of `Android.mk`

libsqlfs/COPYING
```diff
@@ -0,0 +1,506 @@
+		  GNU LESSER GENERAL PUBLIC LICENSE
```
pascal-brand-st-dev: Update `LICENSE` file too

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,491 @@
...
+{
+	struct stat st;
+	int rc;
+
+	mkdir("/data", 0700);
+	mkdir("/data/tee", 0700);
+	rc = stat("/data/tee", &st);
+	if (rc < 0)
+		return rc;
+	rc = sqlfs_open("/data/tee/sstore.db", &db);
```
pascal-brand-st-dev: What if this file has not been correctly cleaned-up? any risk?

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,491 @@
...
+struct dir_entry {
+	char *name;
+	TAILQ_ENTRY(dir_entry) link;
+};
+
+static pthread_mutex_t dir_db_mutex = PTHREAD_MUTEX_INITIALIZER;
+static struct handle_db dir_db =
+		HANDLE_DB_INITIALIZER_WITH_MUTEX(&dir_db_mutex);
+
+static sqlfs_t *db;
```
pascal-brand-st-dev: No risk as `db` is not protected by a mutex?

tee-supplicant/src/sql_fs.c
jforissier: What do you mean?
If the file is corrupt, I suppose it's possible to get an error code here and tee-supplicant will exit with "sql_fs_init() failed". Maybe I should add an EMSG here to print the file name.
Also if the DB is corrupt I guess we may get an error later when trying to access the contents. Hopefully error codes are properly handled in OP-TEE.

tee-supplicant/src/sql_fs.c
jforissier: `tee-supplicant` is single-threaded so no special protection is needed, I think.
Which brings the question: why use `HANDLE_DB_INITIALIZER_WITH_MUTEX` above? I have copied this from `tee_supp_fs.c` but it seems totally superfluous. Am I missing something?

libsqlite3/COPYING
jforissier: Would it be OK if I append the following to `LICENSE`:
```
===

SQLite is public domain. See libsqlite3/COPYING.

===

Libsqlfs is distributed under the GNU Lesser General Public License, version 2
or later. See libsqlfs/COPYING.
```

Makefile
jforissier: True. I'll work on it.

libsqlite3/COPYING
pascal-brand-st-dev: Yes, this looks good

tee-supplicant/src/sql_fs.c
pascal-brand-st-dev: You're right. Forget about my comment.

tee-supplicant/src/sql_fs.c
pascal-brand-st-dev: That's right, I can't see now why we're using a mutex. @jenswi-linaro any idea?

jenswi-linaro: In the log I can see that I introduced that. I don't know why, I will be needed when we introduce multithreading in supplicant. The GP socked API will most likely require a multithreaded supplicant.

jforissier: OK, I think I should at least make things consistent here. Since we don't currently run multi-threaded, I am inclined to simply remove `fd_db_mutex` and `dir_db_mutex`.

The other option is also simple enough: just drop `db` and pass `NULL`, letting `libsqlfs` use it own per-thread internal context (see https://github.com/guardianproject/libsqlfs/blob/master/README#L127-L138). It also means replacing `sqlfs_open()` with `sqlfs_init()`. But I've not tested that.

Just let me know what you prefer.

jenswi-linaro: I'm OK with either way, as long as things are consistent.

jforissier:
Updated.

pascal-brand-st-dev:
`Reviewed-by: Pascal Brand <pascal.brand@linaro.org>`

jforissier:
Updated.

libsqlfs/src/sqlfs.c
```diff
@@ -2755,17 +2755,22 @@ int sqlfs_proc_read(sqlfs_t *sqlfs, const char *path, char *buf, size_t size, of
         return -EBUSY;
     }
 
-    value.data = buf;
-    value.size = size;
-    r = get_value(get_sqlfs(sqlfs), path, &value, offset, offset + size);
-    if (r != SQLITE_OK)
-        result = -EIO;
-    else if ((size_t) offset > existing_size) /* nothing to read */
+    if ((size_t) offset >= existing_size) /* nothing to read */
```
pascal-brand-st-dev: I'd like https://github.com/guardianproject/libsqlfs/pull/10 is first merged

jbech-linaro:
Same here Jerome, I've had a look at this, but again, let's talk more about details next Tuesday.

Android.mk
```diff
@@ -38,5 +38,41 @@ LOCAL_EXPORT_C_INCLUDE_DIRS := $(LOCAL_PATH)/public
 
 include $(BUILD_SHARED_LIBRARY)
 
+ifeq ($(WITH_SQLFS),y)
```
jbech-linaro: Should we use the `CFG_xyz` naming convention instead?

Android.mk
```diff
@@ -38,5 +38,41 @@ LOCAL_EXPORT_C_INCLUDE_DIRS := $(LOCAL_PATH)/public
...
+                          -Wdiscarded-qualifiers \
+                          -Wmissing-declarations \
+                          -Wwrite-strings -Wstrict-prototypes
+LOCAL_CFLAGS += $(filter-out $(libsqlfs_CFLAGS_REMOVE),$(optee_CFLAGS)) \
+                -Wno-suggest-attribute=format
+LOCAL_STATIC_LIBRARIES := libsqlite3
+LOCAL_LDLIBS := -llog
+include $(BUILD_SHARED_LIBRARY)
+
+endif # WITH_SQLFS == y
```
jbech-linaro: If you change the name of the flag don't forget this.

Makefile
```diff
@@ -14,6 +14,12 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 
+WITH_SQLFS ?= n
+
+ifeq ($(WITH_SQLFS),y)
```
jbech-linaro: And this

Makefile
```diff
@@ -14,6 +14,12 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 
+WITH_SQLFS ?= n
```
jbech-linaro: And this

Makefile
```diff
@@ -122,3 +128,28 @@ copy_export: build
 	cp ${O}/libteec/libteec.so* ${EXPORT_DIR}/lib
 	cp ${O}/tee-supplicant/tee-supplicant ${EXPORT_DIR}/bin
 	cp public/*.h ${EXPORT_DIR}/include
+
+ifeq ($(WITH_SQLFS),y)
```
jbech-linaro: And here ...

libsqlfs/Makefile
```diff
@@ -0,0 +1,60 @@
+# !! Please note: !!
+#
+# The license for libsqlfs is LGPLv2, so be careful about changing the code
+# and/or linking it statically! See COPYING section 6.
```
jbech-linaro: It might even be worth adding a `$(info Warning: ...)` when trying to do so?

tee-supplicant/Makefile
```diff
@@ -22,11 +22,16 @@ TEES_SRCS	:= tee_supplicant.c \
 ifeq ($(RPMB_EMU),1)
 TEES_SRCS	+= sha2.c hmac_sha2.c
 endif
+ifeq ($(WITH_SQLFS),y)
```
jbech-linaro: One more `WITH_SQLFS`.

tee-supplicant/Makefile
```diff
@@ -39,6 +44,13 @@ endif
 TEES_FILE	:= $(OUT_DIR)/$(PACKAGE_NAME)
 TEES_LDFLAGS    := -L$(OUT_DIR)/../libteec -lteec
 
+ifeq ($(WITH_SQLFS),y)
```
jbech-linaro: And here.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	ds->handle = handle;
+	return ds;
+err:
+	put_dir(ds);
+	return NULL;
+}
+
+static int sql_fs_open(struct tee_fs_rpc *fsrpc)
+{
+	char *path = (char *)(fsrpc + 1);
```
jbech-linaro: It seems a bit fragile to just use the string, would it be possible to also use a length variable to tell the size of the string. In `new_file` below for example, we will just use the string and then do a `strdup`. I would prefer use `strndup` there instead.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+static int sql_fs_unlink(struct tee_fs_rpc *fsrpc)
+{
+	char *path = (char *)(fsrpc + 1);
+
+	return sqlfs_proc_unlink(db, path);
+}
+
+static int sql_fs_rename(struct tee_fs_rpc *fsrpc)
+{
+	char *paths = (char *)(fsrpc + 1); /* "old.txt\0new.txt\0" */
```
jbech-linaro: Here we not only rely on one null character, here we rely on two null characters. It would be good to see two size's attached to the data which describes the expected length of both of them.

Thanks for adding the comment! :+1:

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+static int sql_fs_truncate(struct tee_fs_rpc *fsrpc)
+{
+	struct file_state *fs = handle_lookup(&fd_db, fsrpc->fd);
+
+	return sqlfs_proc_truncate(db, fs->path, fs->pos);
+}
+
+static int sql_fs_mkdir(struct tee_fs_rpc *fsrpc)
+{
+	char *path = (char *)(fsrpc + 1);
```
jbech-linaro: Once more, length/size checks.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	return sqlfs_proc_mkdir(db, path, mode);
+}
+
+static int fill_dir(void *buf, const char *name, const struct stat *statp,
+		    off_t off)
+{
+	struct dir_state *ds = (struct dir_state *)buf;
+	struct dir_entry *de = NULL;
+	char *dname;
+	(void)statp;
```
jbech-linaro: Not a big deal, but we have moved away from casting to instead use `__unused` or `__maybe_unused`.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	(void)off;
+
+	if (!strcmp(name, ".") || !strcmp(name, ".."))
```
jbech-linaro: Use `strncmp` instead?

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+		return 0;
+
+	dname = strdup(name);
```
jbech-linaro: `fill_dir` is a function "defined" by sqlfs? I.e, we can't change it (easily) and therefore we can't check the length or `name` here in a good way. Otherwise I'd like to use `strndup` instead.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	if (!de)
+		goto err;
+
+	de->name = dname;
+	TAILQ_INSERT_TAIL(&ds->dir_entries, de, link);
+	return 0;
+err:
+	free(dname);
+	free(de);
+	return 1;
```
jbech-linaro: `1` stands for? Seems like a magic number, can we define it to something more understandable?

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+
+	TAILQ_REMOVE(&ds->dir_entries, de, link);
+	memcpy(outname, name, len);
+
+	return 0;
+}
+
+static int sql_fs_rmdir(struct tee_fs_rpc *fsrpc)
+{
+	char *path = (char *)(fsrpc + 1);
```
jbech-linaro: In this function it would probably also be good to check length for the string before calling the sql function.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+
+	return sqlfs_proc_rmdir(db, path);
+}
+
+static int sql_fs_access(struct tee_fs_rpc *fsrpc)
+{
+	char *path = (char *)(fsrpc + 1);
```
jbech-linaro: Same here.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	int flags = tee_fs_conv_accessflags(fsrpc->flags);
+
+	return sqlfs_proc_access(db, path, flags);
+}
+
+static int sql_fs_begin(void)
+{
+	if (sqlfs_begin_transaction(db))
+		return 0;
+	return -1;
```
jbech-linaro: Magic number?

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+static int sql_fs_closedir(struct tee_fs_rpc *fsrpc)
+{
+	struct dir_state *ds = handle_lookup(&dir_db, fsrpc->arg);
+
+	if (!ds)
+		return -EBADF;
+
+	put_dir(ds);
+
+	return 0;
```
jbech-linaro: Same for the functions returning 0. Maybe define it to something with "SUCCESS" or "OK" or something like that in the name.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	return -1;
+}
+
+static int sql_fs_end(struct tee_fs_rpc *fsrpc)
+{
+	int rollback = fsrpc->arg;
+
+	if (sqlfs_complete_transaction(db, rollback))
+		return 0;
+	return -1;
```
jbech-linaro: And here.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+}
+
+int sql_fs_init(void)
+{
+	const char *db_path = "/data/tee/sstore.db";
```
jbech-linaro: I think on some systems (AOSP I think) you need to select another location for the secure storage files. So maybe we should make the `/data/tee/` path a compile time flag? The default value could be `/data/tee/`. What do you say?

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+ */
+int sql_fs_process(void *cmd, size_t cmd_size)
+{
+	struct tee_fs_rpc *fsrpc = cmd;
+	int ret = -1;
+
+	if (cmd_size < sizeof(struct tee_fs_rpc))
+		return ret;
+
+	if (cmd == NULL)
```
jbech-linaro: `!cmd`

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,486 @@
...
+	case TEE_FS_LINK:
+		ret = -ENOTSUP;
+	case TEE_FS_BEGIN:
+		ret = sql_fs_begin();
+		break;
+	case TEE_FS_END:
+		ret = sql_fs_end(fsrpc);
+		break;
+	default:
+		break;
```
jbech-linaro: Debug print to catch unsupported rpc calls?

tee-supplicant/src/sql_fs.h
```diff
@@ -0,0 +1,56 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef SQL_FS_H
+#define SQL_FS_H
+
+#ifdef WITH_SQLFS
```
jbech-linaro: One more of those.

tee-supplicant/src/tee_fs.h
```diff
@@ -0,0 +1,184 @@
...
+#include <fcntl.h>
+#include <stdint.h>
+
+/*
+ * Structure for file related RPC calls
+ *
+ * @op     The operation like open, close, read, write etc
+ * @flags  Flags to the operation shared with secure world
+ * @arg    Argument to operation
+ * @fd     NW file descriptor
```
jbech-linaro: Please spell out to "Normal World" instead.

tee-supplicant/src/tee_fs.h
```diff
@@ -0,0 +1,184 @@
...
+#define TEE_FS_RMDIR     13
+#define TEE_FS_ACCESS    14
+#define TEE_FS_LINK      15
+#define TEE_FS_BEGIN     16 /* SQL FS: begin transaction */
+#define TEE_FS_END       17 /* SQL FS: end transaction */
+
+/*
+ * Open flags, defines shared with TEE.
+ */
+#define TEE_FS_O_RDONLY 0x1
```
jbech-linaro: This and the stuff below is basically bit's in a bitfield. Maybe time to add BIT(x) macro in optee_client?

tee-supplicant/tee_supplicant_android.mk
```diff
@@ -21,6 +21,16 @@ LOCAL_C_INCLUDES := $(LOCAL_PATH)/../public \
     $(LOCAL_PATH)/src
 
 LOCAL_SHARED_LIBRARIES := libteec
+
+ifeq ($(WITH_SQLFS),y)
```
jbech-linaro: One more `WITH_SQLFS`

jbech-linaro:
All in all I think this is good, I've added quite a few comments, but the majority of the is about a) WITH_SQLFS macro b) Checking size/length of path and filename strings. I think it would be worth trying to add checks for that, since we get all sorts of strange behavior when our secure storage implementation fails. Also, since this robustness in general for secure storage is highly anticipated.

Android.mk
jforissier: Yes, probably. Will fix.

libsqlfs/Makefile
jforissier: Not sure how to do that, but I can change the supplicant link line so that it explicitly looks for a shared object (hence, link would fail if one were to change the build flags in `libsqlfs/` to make it a static lib). Do you think the following would help?

```
diff --git a/tee-supplicant/Makefile b/tee-supplicant/Makefile
index 32c0a2c..0f06ae1 100644
--- a/tee-supplicant/Makefile
+++ b/tee-supplicant/Makefile
@@ -48,7 +48,10 @@ ifeq ($(CFG_SQL_FS),y)
 TEES_CFLAGS    += -DCFG_SQL_FS
 TEES_CFLAGS    += -Wno-strict-prototypes -Wno-implicit-function-declaration \
                   -Wno-nested-externs -Wno-unused-function
-TEES_LDFLAGS   += -L$(OUT_DIR)/../libsqlfs -lsqlfs -ldl
+# Note: explicitely requesting a shared library (.so) here, because statically
+# linking a LGPL library has licensing implications. Make sure you review
+# and comply with libsqlfs/COPYING section 6 before doing so.
+TEES_LDFLAGS   += -L$(OUT_DIR)/../libsqlfs -l:libsqlfs.so -ldl
 endif
 
 tee-supplicant: $(TEES_FILE)
```

tee-supplicant/src/sql_fs.c
jforissier: Why would you rely on the value of `fsrpc->len` being correct more than on the fact that the string is null-terminated? The lower-level functions still expect null-terminated strings.

tee-supplicant/src/sql_fs.c
jforissier: Agreed, it is ugly.
But, there is no `__unused` etc. defined in `optee_client` currently. Should we create a common include directory for private definitions that are common to all the binaries in `optee_client`? Such as, `optee_client/common/include/compiler.h` or somewhere else? Sounds a bit overkill for now maybe? Maybe not, there's already `__aligned(x)` in `tee_supplicant.c` that could be moved to the new header.
The other option is to `#define __unused __attribute__((unused))` locally in `sql_fs.c`.
Let me know what you prefer.

tee-supplicant/src/sql_fs.c
jforissier: Why?

tee-supplicant/src/sql_fs.c
jforissier: Correct. The prototype for `fill_dir` is defined by `libsqlfs` and it works with null-terminated strings.

tee-supplicant/src/sql_fs.c
jforissier: The doc [1] says "@return 1 if buffer is full, zero otherwise". Which I interpreted as "0: success, 1: error". The code in `sqlfs.c` indeed stops enumerating the directory entries when it gets a non-zero value [2].
I will `#define FILLER_SUCCESS 0` and `#define FILLER_ERROR 1` for clarity.

[1] https://github.com/OP-TEE/optee_client/pull/50/files#diff-81966fc05d81c061a15ec01b14f468a5R89
[2] https://github.com/jforissier/optee_client/blob/064239cd2037566a8adcffbaee77febf29867a91/libsqlfs/src/sqlfs.c#L2032-L2033

tee-supplicant/src/sql_fs.c
jforissier: No, this is the convention: <0 for error, >= 0 for success, now documented here: https://github.com/OP-TEE/optee_os/pull/797/commits/e453be8ceeecd804606a7f3c5569210e624fb79a#diff-5b5b7c2f5a37b83392407582ea1880c1R73 ;)

tee-supplicant/src/sql_fs.c
jforissier: Yeah, I could define generic codes `TEE_FS_OK` to 0 and `TEE_FS_ERROR` to -1 in `tee-supplicant/src/tee_fs.h`, and use that. But would that really make things more legible? By this logic, you could argue: "why `return -EBADF`, this inconsistent, you need another #define"... and also what about the place where we return a file descriptor?
I'd say you just need to keep in mind the convention I linked above and the code is OK as it is. But I may be biased, since I wrote the code ;)
Let me know what you prefer.

tee-supplicant/src/sql_fs.c
jforissier: Agreed. This has been partially taken care of in `tee_supp_fs.c` but is probably not AOSP-friendly since the `#define` is hard-coded and cannot be controlled from the build Makefile/scripts.
I'll change to:
```
#ifndef SQL_FS_DB_PATH
#define SQL_FS_DB_PATH "/data/tee/sstore.db"
#endif
/* ... */
const char *db_path = SQL_FS_DB_PATH;
```

tee-supplicant/src/sql_fs.c
jforissier: Yes sir!

tee-supplicant/src/sql_fs.c
jforissier: Yes, that should even be `EMSG()` I suppose.

tee-supplicant/src/tee_fs.h
jforissier: OK

tee-supplicant/src/tee_fs.h
jforissier: Yes, in a later PR maybe? Because we'd better have identical definitions in `optee_os`, and the `BIT()` macro was added only recently in `optee_os`.

jforissier:
Fixed some issues with compiler commands and flags, now Travis is OK.

tee-supplicant/src/sql_fs.c
d3zd3z: My personal preference is for the '0', although this one is a bit of a wash. At least with kernel programming and Posix APIs, 0 for success seems very ingrained, and rarely is a define used.

Oh if C only had real enums :-)

Makefile
```diff
@@ -122,3 +128,27 @@ copy_export: build
 	cp ${O}/libteec/libteec.so* ${EXPORT_DIR}/lib
 	cp ${O}/tee-supplicant/tee-supplicant ${EXPORT_DIR}/bin
 	cp public/*.h ${EXPORT_DIR}/include
```
etienne-lms: +	cp ${O}/libsqlfs/libsqlfs.so* ${EXPORT_DIR}/lib

jforissier: Good catch. I think I will rather:
```diff
+    [ "$(BUILD-LIBSQLFS)" ] && cp ${O}/libsqlfs/libsqlfs.so* ${EXPORT_DIR}/lib
```
...so that it will fail if SQL FS is enabled but the lib is not present.

jforissier:
Updated.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+	struct file_state *fs;
+	int fd;
+
+	fs = calloc(1, sizeof(*fs));
+	if (!fs)
+		return NULL;
+	fs->fd = -1;
+
+	fs->path = strdup(path);
+	if (!fs)
```
etienne-lms: `if (!fs->path)`

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+	char *from = paths;
+	char *to = paths + strlen(paths) + 1;
+
+	return sqlfs_proc_rename(db, from, to);
+}
+
+static int sql_fs_truncate(struct tee_fs_rpc *fsrpc)
+{
+	struct file_state *fs = handle_lookup(&fd_db, fsrpc->fd);
+
```
etienne-lms: `if (!fs) return -EBADF;`

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+	if (!ds)
+		return -EBADF;
+
+	de = TAILQ_FIRST(&ds->dir_entries);
+	if (!de)
+		return -ENOENT;
+
+	name = de->name;
+	len = strlen(name) + 1;
```
etienne-lms: rather use `strnlen()`:
```
len = strnlen(name, outlen);
if (len == outlen)
   return -ENAMETOOLONG;
memcpy(outname, name, len + 1);
```

Could also check `de->name != NULL` before using `name`.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+#define SQL_FS_DB_PATH "/data/tee/sstore.db"
+#endif
+
+int sql_fs_init(void)
+{
+	const char *db_path = SQL_FS_DB_PATH;
+	int rc;
+
+	mkdir("/data", 0700);
+	mkdir("/data/tee", 0700);
```
etienne-lms: those mkdir's here are legitimate only if SQL_FS_DB_PATH is located in /data/tee.

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+		break;
+	case TEE_FS_RMDIR:
+		ret = sql_fs_rmdir(fsrpc);
+		break;
+	case TEE_FS_ACCESS:
+		ret = sql_fs_access(fsrpc);
+		break;
+	case TEE_FS_LINK:
+		ret = -ENOTSUP;
+	case TEE_FS_BEGIN:
```
etienne-lms: missing `break`

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+		ret = sql_fs_readdir(fsrpc);
+		break;
+	case TEE_FS_RMDIR:
+		ret = sql_fs_rmdir(fsrpc);
+		break;
+	case TEE_FS_ACCESS:
+		ret = sql_fs_access(fsrpc);
+		break;
+	case TEE_FS_LINK:
+		ret = -ENOTSUP;
```
etienne-lms: missing `break;`

tee-supplicant/src/sql_fs.h
```diff
@@ -0,0 +1,56 @@
...
+{
+	return 0;
+}
+
+static inline int sql_fs_process(void *cmd, size_t cmd_size)
+{
+	(void)cmd;
+	(void)cmd_size;
+
+	return -1;
```
etienne-lms: maybe `-ENOTSUP` instead of `-1`

tee-supplicant/src/tee_fs.h
```diff
@@ -0,0 +1,184 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <sys/types.h>
```
etienne-lms: alphabetical sorting ?

tee-supplicant/src/tee_supplicant.c
```diff
@@ -170,6 +172,20 @@ static void process_fs(union tee_rpc_invoke *request)
...
+static void process_sql_fs(union tee_rpc_invoke *request)
+{
+	TEEC_SharedMemory shm;
+
+	if (request->recv.num_params != 1 || get_param(request, 0, &shm)) {
+		request->send.ret = TEEC_ERROR_BAD_PARAMETERS;
+		return;
+	}
+
+	sql_fs_process(shm.buffer, shm.size);
```
etienne-lms: `sql_fs_process()` return value is discarded ?

libsqlfs/Makefile
```diff
@@ -0,0 +1,63 @@
...
+SQLFS_CFLAGS_REMOVE := -Wall -Wunused-parameter \
+        -Wmissing-prototypes -Wdiscarded-qualifiers \
+        -Wunused-variable -Wmissing-declarations \
+	-Wwrite-strings -Wstrict-prototypes
+SQLFS_CFLAGS := $(addprefix -I, $(SQLFS_INCLUDES)) \
+	$(filter-out $(SQLFS_CFLAGS_REMOVE),$(CFLAGS)) \
+	-Wno-missing-format-attribute
+
+SQLFS_LFLAGS := -lpthread
+SQLFS_LIBS := $(CURDIR)/../$(O)/libsqlite3/libsqlite3.a
```
etienne-lms: use `$(OUT_DIR)` instead of `$(CURDIR)/../$(O)`

libsqlfs/Makefile
```diff
@@ -0,0 +1,63 @@
...
+SQLFS_LIBRARY := $(OUT_DIR)/$(LIB_MAJ_MIN)
+
+libsqlfs: $(SQLFS_LIBRARY)
+	$(VPREFIX)ln -sf $(SQLFS_LIBRARY) $(OUT_DIR)/$(LIB_MAJOR)
+	$(VPREFIX)ln -sf $(OUT_DIR)/$(LIB_MAJOR) $(OUT_DIR)/$(LIB_NAME)
+
+$(SQLFS_LIBRARY): $(SQLFS_OBJS)
+	@echo "  LINK    $@"
```
etienne-lms: minor comment:
optee_teec used to use `"  LD      $@"` as linker verbosity. Why changing to `"  LINK    $@"` here are in legacy optee_teec makefiles ?

tee-supplicant/src/sql_fs.c
etienne-lms: agree with jerome

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,494 @@
...
+		rc = sqlfs_proc_getattr(db, fs->path, &sb);
+		if (rc < 0)
+			return rc;
+		fs->pos = sb.st_size + fsrpc->arg;
+		}
+		break;
+	default:
+		return -EINVAL;
+	}
+
```
etienne-lms: lseek man page:
> The lseek() function allows the file offset to be set beyond the end of
> the file (but this does not change the size of the file).  If  data  is
> later written at this point, subsequent reads of the data in the gap (a
> "hole") return null bytes ('\0') until data is  actually  written  into
> the gap.

Do we support gaps ? or should we check `fs->pos` remains inside file current size ?

tee-supplicant/src/sql_fs.c
jforissier: Oops, will fix

tee-supplicant/src/sql_fs.c
jforissier: We do support gaps. See `test_file_hole()` in `xtest_6000.c` (that's `xtest 6007`).

tee-supplicant/src/sql_fs.c
jforissier: Yep, will fix.

tee-supplicant/src/sql_fs.c
jforissier: Agreed.

jforissier: Added `assert(de->name)` since the list is managed by us and should not be allowed to contain an entry with a NULL name.

tee-supplicant/src/sql_fs.c
jforissier: True. I'll rework this to read `mkdir_recursive(SQL_FS_DB_PATH, 0700);` instead or something like that.

tee-supplicant/src/sql_fs.c
jforissier: Ouch!

tee-supplicant/src/sql_fs.h
jforissier: OK

tee-supplicant/src/tee_fs.h
jforissier: Yes, will update

tee-supplicant/src/tee_supplicant.c
jforissier: That's a bug indeed. `sql_fs_process()` should return:
- >= 0 when it could invoke the requested operation (which will then store its own status, <0 or >=0, into `((struct tee_fs_rpc *)shm.buffer)->res`). In which case we should set `request->send.ret = TEEC_SUCCESS` to indicate the success of the RPC request.
- < 0 when it could not (invalid cmd or cmd_size), then we set `request->send.ret < 0`.

I will clean this in it's own commit to make it clear (hopefully).

libsqlfs/Makefile
jforissier: I'd rather use `LD $@` when the linker is invoked directly. Here we are invoking $(CC) to link, that's why (and I did not want to use `CC $@` obviously because the line would not stand out in the compile lines).
The linux kernel does a similar thing. It uses `LD` only if `$(LD)` is called directly, `AR` if it's `$(AR)`, and `LINK` for the kernel link step which actually calls a script (as far as I can tell).
```
$ grep -A 1  cmd_link.*= Makefile scripts/Makefile.build 
Makefile:      cmd_link-vmlinux = $(CONFIG_SHELL) $< $(LD) $(LDFLAGS) $(LDFLAGS_vmlinux)
Makefile:quiet_cmd_link-vmlinux = LINK    $@
Makefile-
--
scripts/Makefile.build:quiet_cmd_link_o_target = LD      $@
scripts/Makefile.build-# If the list of objects to link is empty, just create an empty built-in.o
scripts/Makefile.build:cmd_link_o_target = $(if $(strip $(obj-y)),\
scripts/Makefile.build-		      $(LD) $(ld_flags) -r -o $@ $(filter $(obj-y), $^) \
--
scripts/Makefile.build:quiet_cmd_link_l_target = AR      $@
scripts/Makefile.build:cmd_link_l_target = rm -f $@; $(AR) rcs$(KBUILD_ARFLAGS) $@ $(lib-y)
scripts/Makefile.build-
--
scripts/Makefile.build:quiet_cmd_link_multi-y = LD      $@
scripts/Makefile.build:cmd_link_multi-y = $(LD) $(ld_flags) -r -o $@ $(link_multi_deps) $(cmd_secanalysis)
scripts/Makefile.build-
scripts/Makefile.build:quiet_cmd_link_multi-m = LD [M]  $@
scripts/Makefile.build:cmd_link_multi-m = $(cmd_link_multi-y)
scripts/Makefile.build-
```

jforissier:
Updated.

tee-supplicant/src/sql_fs.c
d3zd3z: I wish I still had the rant I wrote about this while at Qualcomm.  Basically, there is no benefit to strncmp when comparing to a constant string. And, people often get it wrong, and it ends up being a string prefix comparison.

In summary, strncmp is almost never useful, unless you're actually trying to compare a prefix, or dealing with possibly non-terminated fixed-length strings.

In other cases, you'll need to do some kind of length check and make a decision based on that. This is only the case, though, when both strings are of unknown length. When one of the strings is constant, the length of the `strcmp` will be bound by that, regardless.

jforissier:
@etienne-lms and @jbech-linaro is this OK now? Can you give your R-b?

tee-supplicant/src/sql_fs.c
```diff
@@ -0,0 +1,524 @@
...
+		break;
+	case TEE_FS_END:
+		st = sql_fs_end(fsrpc);
+		break;
+	default:
+		EMSG("Unexpected SQL FS operation: %d", fsrpc->op);
+		break;
+	}
+
+	fsrpc->res = st;
```
etienne-lms: `res` is filled with unpredictable content and no error status is returned. 
(maybe it can be part of another change)

jbech-linaro:
Yes @jforissier :+1: 

tee-supplicant/src/tee_supplicant.c
etienne-lms: you updated the legacy fs return status but `sql_fs_process()` return value is still discarded.

tee-supplicant/src/sql_fs.c
jforissier: Good catch, I'll make another PR since it's not directly related to this one.

tee-supplicant/src/tee_supplicant.c
jforissier: Doh! I'll fix that. I will keep the legacy fs fix but I won't squash it, i.e., keep its own commit since it's unrelated to SQL.

jforissier:
Updated.

etienne-lms:

I was to add tested-by but i have a failure on qemu_v8. i have not tested same against qemu_v7.
i could not tell if it is a regression due to this PR or not.

etienne-lms:
@jforissier: I am testing the whole sqlfs on qemu_v8: this + optee_test  [pr113](https://github.com/OP-TEE/optee_test/pull/113) and [pr118](https://github.com/OP-TEE/optee_test/pull/118). i get some failures but i will report them on pr113.

jforissier:
@etienne-lms I also had some failures in xtest which were slightly dificult to reproduce, such as: xtest 6015 would fail when running the whole xtest and only on first try after a clean boot. The error was `malloc()` returning NULL. Increasing the TEE core heap size from 24 to 32 KiB fixed the problem. So it's possible other platforms may need to increase the heap to run the full `xtest` with several filesystems.

PR #51, opened 2016-05-24
Syslog

jforissier:


tee-supplicant/src/tee_supplicant.c
```diff
@@ -410,6 +431,9 @@ int main(int argc, char *argv[])
 			case RPC_CMD_SHM_FREE:
 				process_free(&request);
 				break;
+			case RPC_CMD_LOG:
+				process_log(&request);
+				break;
 			default:
 				EMSG("Cmd [0x%" PRIx32 "] not supported",
```
vchong: This is not part of the PR but is missing a format specifier.

jforissier: `PRIx32` *is* the format specifier ;)
```
$ git grep -n "define PRIx32"
lib/libutils/isoc/include/inttypes.h:66:#define PRIx32          "x"
```

vchong: Oops.. ok. Sorry for the noise.

jforissier:
Closing since this is not the correct approach (see comments in https://github.com/OP-TEE/optee_os/issues/810).

PR #52, opened 2016-06-05
Check in TEEC_AllocateSharedMemory if shm->flags is correctly set. As…

flowher:
… per "4.3.6. TEEC_SharedMemory" value of flags should be either TEEC_MEM_INPUT, TEEC_MEM_OUTPUT or both.

Signed-off-by: Krzysztof Kwiatkowski <kris@amongbytes.com>


jenswi-linaro:
So the spec says:
*All other bits in this field SHOULD be set to zero, and are reserved for future use*

I don't have a strong opinion on whether we should enforce the other bits to be clear or not. @jbech-linaro , @jforissier what do you think?

flowher:
The reason I've added it is issue #90 (it was suggested by Pascal). And indeed with this check there would be no issue #90.

jbech-linaro:
Since the wording in the spec is **SHOULD** then I think we should go a head with this patch. Even though it might not matter technically, it's more correct according to the spec.

Please just update the commit message a bit (too long rows, see my other related comment in [PR99](https://github.com/OP-TEE/optee_test/pull/99) you created in OP-TEE test). Otherwise:

jforissier:
I also think that checking the reserved bits is the way to go. Otherwise, it means that applications are free to set them to any value, and therefore the bits are no longer available for future use.

flowher:
Thanks ! 
I'll change a comment.

flowher:
That's also done

PR #53, opened 2016-07-27
Adjust error codes for TEEC_OpenSession() and TEEC_InvokeCommand()

jforissier:
When the ioctl() call inside TEEC_OpenSession() or TEEC_InvokeCommand()
fails, TEEC_ERROR_BAD_STATE is returned. Update the error path to
return more descriptive codes:
- TEEC_ERROR_OUT_OF_MEMORY if ioctl() has set errno to ENOMEM (with the
current driver, this can occur when the driver's private memory pool is
full).
- TEEC_ERROR_GENERIC otherwise.

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -45,7 +45,6 @@
 #include <linux/tee.h>
 
 #include <teec_trace.h>
-#include <err.h>
 
 /* How many device sequence numbers will be tried before giving up */
 #define TEEC_MAX_DEV_SEQ	10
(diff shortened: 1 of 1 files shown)


d3zd3z:
Aside from the error about `err` shadowing the global coming from `err.h`:
```
Reviewed-by: David Brown <david.brown@linaro.org>
```

jforissier:
That's interesting. I don't see any error when I build with `aarch64-linux-gnu-gcc` (Ubuntu 16.04). I wouldn't be surprised if https://github.com/OP-TEE/optee_client/pull/50/commits/a117d0d29f3073ed2d6801720019357e08ca95cb (part of https://github.com/OP-TEE/optee_client/pull/50) would 'fix' the issue.
Still, I will change `err` to something else.

d3zd3z:
I'm kind of wondering why we're including `<err.h>`, although I haven't tried removing it, it seems to mostly be function externs that we aren't using.

jforissier:
Removed `<err.h>` as suggested, all good!

PR #54, opened 2016-08-02
Try loading TAs from another writable path

d3zd3z:
To allow testing of corrupt TA loading on systems where the TA
directory is read-only, try loading TAs first from an alternate
`TEEC_TEST_LOAD_PATH` before the regular path.  This allows the test
to override the TA with one for testing purposes.

Although useful for testing, this should not be enabled for a
production system, since it facilitates replacing a TA with a corrupt
image.

Signed-off-by: David Brown <david.brown@linaro.org>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -32,6 +32,12 @@ CFG_TEE_SUPP_LOG_FILE ?= \"/data/teesupp.log\"
 # The location of the client library file.
 CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 
+# CFG_TA_TEST_PATH
+# Enable the tee test path.  When enabled, the supplicant will try
+# loading from a debug path before the regular path.  This allows test
+# such as 1008.5 that test loading of corrupt TAs.
+CFG_TA_TEST_PATH ?= 1
+
 # Default out dir.
 # Must be a relative path with respect to the op-tee-client root directory
 O               ?= out
--- a/tee-supplicant/tee_supplicant_android.mk
+++ b/tee-supplicant/tee_supplicant_android.mk
@@ -10,6 +10,10 @@ LOCAL_CFLAGS += -DDEBUGLEVEL_$(CFG_TEE_SUPP_LOG_LEVEL) \
                 -DBINARY_PREFIX=\"TEES\" \
                 -DTEEC_LOAD_PATH=\"$(CFG_TEE_CLIENT_LOAD_PATH)\" \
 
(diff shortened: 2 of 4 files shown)


d3zd3z:
Do we want a configurable way of turning this off?

jforissier:
Looks good, but yes, please add something to turn it off in the Makefile.

config.mk
```diff
@@ -32,6 +32,12 @@ CFG_TEE_SUPP_LOG_FILE ?= \"/data/teesupp.log\"
 # The location of the client library file.
 CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 
+# CFG_TA_TEST_PATH
+# Enable the tee test path.  When enabled, the supplicant will try
+# loading from a debug path before the regular path.  This allows test
+# such as 1008.5 that test loading of corrupt TAs.
+CFG_TA_TEST_PATH ?= 1
```
jforissier: '1' is OK but 'y' would be more consistent with the optee_os way.

tee-supplicant/tee_supplicant_android.mk
```diff
@@ -10,6 +10,10 @@ LOCAL_CFLAGS += -DDEBUGLEVEL_$(CFG_TEE_SUPP_LOG_LEVEL) \
                 -DBINARY_PREFIX=\"TEES\" \
                 -DTEEC_LOAD_PATH=\"$(CFG_TEE_CLIENT_LOAD_PATH)\" \
 
+ifeq ($(CFG_TA_TEST_PATH),1)
```
jforissier: `include ../config.mk` not needed?

jforissier:
With my comments addressed:

tee-supplicant/tee_supplicant_android.mk
d3zd3z: It doesn't seem to be needed in Android, I guess because all of the makefiles are processed by a single invocation of make, so these variables will have gotten set by the parent makefile. I can include it to be explicit, but I'm not sure what is the convention with Android.

jforissier: OK, if there's no problem let's not try to fix it ;) Thanks.

jforissier:
Also I don't think it's really useful to keep the two commits, I suggest you squash them.

d3zd3z:
Agree. I'll squash them.

d3zd3z:
Squashed.

PR #55, opened 2016-08-11
Remove the connection_method limit

prime-zeng:
Since we have to add support for login method other than TEE_LOGIN_PUBLIC
in the future, remove the limit now.

Signed-off-by: Zeng Tao <prime.zeng@huawei.com>

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -467,13 +467,6 @@ TEEC_Result TEEC_OpenSession(TEEC_Context *ctx, TEEC_Session *session,
 		goto out;
 	}
 
-	if (connection_method != TEEC_LOGIN_PUBLIC) {
-		eorig = TEEC_ORIGIN_API;
-		res = TEEC_ERROR_NOT_SUPPORTED;
-		goto out;
-	}
-
-
 	buf_data.buf_ptr = (uintptr_t)buf;
 	buf_data.buf_len = sizeof(buf);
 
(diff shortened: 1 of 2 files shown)


jforissier:

libteec/src/tee_client_api.c
```diff
@@ -467,13 +467,6 @@ TEEC_Result TEEC_OpenSession(TEEC_Context *ctx, TEEC_Session *session,
 		goto out;
 	}
 
-	if (connection_method != TEEC_LOGIN_PUBLIC) {
```
d3zd3z: Shouldn't we still be checking, at least somewhere, that the connection method is the expected one? We want to add other methods in the future, but I don't see why we want to remove the check until those are implemented.

d3zd3z:
I'm not sure we want to remove this check until other methods are implemented.

jforissier:
The GP _TEE Client API Specification v1.0_ seems to allow fallback to TEE_LOGIN_PUBLIC when nothing else is implemented:

> Note that this specification does not guarantee separation based on use of different login types – in many embedded platforms there is no notion of “group” or “user” so these login types may fall back to TEEC_LOGIN_PUBLIC – these details of generating the credential for each login type are implementation-defined.

In any case, I'd rather avoid double checks (libteec + TEE kernel) because it's difficult to maintain and error-prone as the implementation evolves (see for instance https://github.com/OP-TEE/optee_os/pull/938). So if we need to restrict which value we accept, I think it should occur in optee_os.

d3zd3z:
I agree with this, and would be happy with this change, if we verify optee_os will restrict to the implemented methods.

jforissier:
This? https://github.com/OP-TEE/optee_os/blob/master/core/kernel/tee_dispatch.c#L60-L81

d3zd3z:
Yeah that seems reasonable. I'm not sure why we were checking in on the user side to begin with.
```
Reviewed-by: David Brown <david.brown@linaro.org>
```

jforissier:
@prime-zeng would you please apply the R-b tags so that I can merge this PR? Thanks.

PR #56, opened 2016-08-17
Use "make V=1" for verbose build

jforissier:
Replace 'VERBOSE_BUILD' with 'V' for simplicity and consistency with
other projects.


jenswi-linaro:

PR #57, opened 2016-09-14
[RFC] libteec: add secure memref parameter type (SWG-186)

etienne-lms:
Introduce a new TEE parameter type identifier: TEEC_MEMREF_SECURE.

TEE Client API provides this type identifier for buffer that client shall
not attempt to access (secure memory buffer) but that are used as
input or output memory buffers.

As for legacy TEEC_MEMREF_xxx parameters, TEEC_SharedMemory structure
holds the memory reference. In case of "secure" memrefs:
- field 'id' provides the dma_buf file handle that refers the buffer
- field 'flag' must set the bit TEEC_MEM_SECURE

This PR requires few changes in the [optee_os PR1052](https://github.com/OP-TEE/optee_os/pull/1052) and tee/optee linux driver [linux/swg PR23](https://github.com/linaro-swg/linux/pull/23).

### DIFF
--- a/libteec/include/linux/tee.h
+++ b/libteec/include/linux/tee.h
@@ -144,6 +144,7 @@ struct tee_ioctl_buf_data {
  * These defines shared memory reference parameters (struct
  * tee_ioctl_param_memref)
  */
+#define TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_SECURE	4
 #define TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_INPUT	5
 #define TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_OUTPUT	6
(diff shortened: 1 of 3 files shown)


jenswi-linaro:
What does `TEEC_MEMREF_SECURE` correspond to in the TA?

etienne-lms:
> What does TEEC_MEMREF_SECURE correspond to in the TA?

in optee_os/lib/libutee/include/tee_api_defines.h
`#define TEE_PARAM_TYPE_MEMREF_SECURE    4
`

etienne-lms:
discard this PR. 
Current plans is that nonsecure world will NOT use a specific 'parameter type identifier' for secure memory buffer. Instead, a new TEE Client API will allow userland to register a memory buffer file descriptor and get a referenced shared memory object that can be used to invoke a TA.

PR #59, opened 2016-10-18
new API to register shared memory from a file descriptor (SWG-186)

etienne-lms:
This is a proposal for implementing the ability for a TA to safely handle memory buffer arguments related to secure buffers. See SWG-186.

This proposal defines a new API in the TEE Client API: `TEEC_RegisterSharedMemoryFileDescriptor()`.

The libteec requests a ioctl to the TEE driver to register the buffer and get a new file descriptor for the TEE shared memory structure. API will simply close this file descriptor to release the shared memory resource.

This PR is related to changes in the TEE linux driver: https://github.com/linaro-swg/linux/pull/26.

### DIFF
--- a/libteec/include/linux/tee.h
+++ b/libteec/include/linux/tee.h
@@ -38,7 +38,6 @@
  * data passed back and forth using TEE_IOC_CMD.
  */
 
-
 /* Helpers to make the ioctl defines */
 #define TEE_IOC_MAGIC	0xa4
 #define TEE_IOC_BASE	0
@@ -76,6 +75,7 @@ struct tee_ioctl_version_data {
 	__u32 impl_caps;
 	__u32 gen_caps;
 };
+
 /**
  * TEE_IOC_VERSION - query version of TEE
  *
@@ -100,6 +100,7 @@ struct tee_ioctl_shm_alloc_data {
 	__u32 flags;
 	__s32 id;
 };
+
 /**
  * TEE_IOC_SHM_ALLOC - allocate shared memory
  *
@@ -156,7 +157,6 @@ struct tee_ioctl_buf_data {
 	__u64 buf_len;
 };
 
-
 /*
  * Attributes for struct tee_ioctl_param, selects field in the union
  */
@@ -184,7 +184,7 @@ struct tee_ioctl_buf_data {
 
 /*
  * Matches TEEC_LOGIN_* in GP TEE Client API
- * Is only defined for GP compliant TEEs
+ * Are only defined for GP compliant TEEs
  */
 #define TEE_IOCTL_LOGIN_PUBLIC			0
 #define TEE_IOCTL_LOGIN_USER			1
@@ -216,6 +216,8 @@ struct tee_ioctl_param_memref {
  * @a: first value
  * @b: second value
  * @c: third value
+ *
+ * Value parameters are passed unchecked to the destination
  */
 struct tee_ioctl_param_value {
 	__u64 a;
@@ -248,7 +250,7 @@ struct tee_ioctl_param {
  * struct tee_ioctl_open_session_arg - Open session argument
  * @uuid:	[in] UUID of the Trusted Application
  * @clnt_uuid:	[in] UUID of client
- * @clnt_login:	[in] Login class of client, TEE_LOGIN_* above
+ * @clnt_login:	[in] Login class of client, TEE_IOCTL_LOGIN_* above
  * @cancel_id:	[in] Cancellation id, a unique value to identify this request
  * @session:	[out] Session id
  * @ret:	[out] return value
@@ -287,7 +289,8 @@ struct tee_ioctl_open_session_arg {
 				     struct tee_ioctl_buf_data)
 
 /**
- * struct tee_ioctl_invoke_func_arg - Invokes a function in a Trusted Application
+ * struct tee_ioctl_invoke_func_arg - Invokes a function in a Trusted
+ * Application
  * @func:	[in] Trusted Application function, specific to the TA
  * @session:	[in] Session id
  * @cancel_id:	[in] Cancellation id, a unique value to identify this request
@@ -332,6 +335,7 @@ struct tee_ioctl_cancel_arg {
 	__u32 cancel_id;
 	__u32 session;
 };
+
 /**
  * TEE_IOC_CANCEL - Cancels an open session or invoke
  */
@@ -345,6 +349,7 @@ struct tee_ioctl_cancel_arg {
 struct tee_ioctl_close_session_arg {
 	__u32 session;
 };
+
 /**
  * TEE_IOC_CLOSE_SESSION - Closes a session
  */
@@ -374,6 +379,7 @@ struct tee_iocl_supp_recv_arg {
 	 * struct tee_ioctl_param params[num_params];
 	 */
 } __aligned(8);
+
 /**
  * TEE_IOC_SUPPL_RECV - Receive a request for a supplicant function
  *
@@ -382,7 +388,6 @@ struct tee_iocl_supp_recv_arg {
  */
 #define TEE_IOC_SUPPL_RECV	_IOR(TEE_IOC_MAGIC, TEE_IOC_BASE + 6, \
 				     struct tee_ioctl_buf_data)
-
 
 /**
  * struct tee_iocl_supp_send_arg - Send a response to a received request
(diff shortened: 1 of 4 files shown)


public/tee_client_api.h
```diff
@@ -517,6 +518,23 @@ TEEC_Result TEEC_AllocateSharedMemory(TEEC_Context *context,
...
+ * @param sharedMem  pointer to the shared memory structure to register.
+ * @param fd         file descriptor of the target memory.
+ *
+ * @return TEEC_SUCCESS              The registration was successful.
+ * @return TEEC_ERROR_OUT_OF_MEMORY  Memory exhaustion.
+ * @return TEEC_Result               Something failed.
+ */
+TEEC_Result TEEC_RegisterSharedMemoryFileDescriptor(TEEC_Context *context,
+						    TEEC_SharedMemory *sharedMem,
+						    int fd);
```
jenswi-linaro: What does the life cycle of the supplied `fd` look like?
In `TEEC_ReleaseSharedMemory()` the file descriptor is closed. What if the file descriptor instead was closed as soon as it was successfully registered? Then we wouldn't need to keep track of it here.

jforissier: Since the new function is not defined by the GP specs, I think this should go into a separate header file: `tee_client_api_extensions.h` (same as we did for the key derivation extensions to the Internal Core API).

libteec/include/linux/tee.h
```diff
@@ -396,4 +396,33 @@ struct tee_iocl_supp_send_arg {
  * munmap(): unmaps previously shared memory
  */
 
+/**
```
etienne-lms: It is a matter of ioctl cmd IDs.

`TEE_IOC_SHM_ALLOC` uses `TEE_IOC_BASE + 1`.
I have incremented the cmd ID for  new `TEE_IOC_SHMREGISTER_FD`, so it is currently `TEE_IOC_BASE + 8`.

If I move this ioctl next to the 'shm alloc' one, i feel it would be better to change all other IDs:
* `TEE_IOC_VERSION` uses  `TEE_IOC_BASE + 0`        _(unchanged)_
* `TEE_IOC_SHM_ALLOC` uses `TEE_IOC_BASE + 1`      _(unchanged)_
* `TEE_IOC_SHM_REGISTER_FD` uses `TEE_IOC_BASE + 2`   _(new)_
* `TEE_IOC_OPEN_SESSION` uses `TEE_IOC_BASE + 3`         _(changed!)_
* `TEE_IOC_INVOKE` uses  `TEE_IOC_BASE + 4`         _(changed!)_
* an so on

This affects IDs that are already on their way to the upstream...

Would you rather want me to use `TEE_IOC_BASE + 2` or `TEE_IOC_BASE + 8` for the new `TEE_IOC_SHM_REGISTER_FD` ?

public/tee_client_api.h
etienne-lms: This expects `fd` is valid while client calls `TEEC_RegisterSharedMemoryFileDescriptor()`.
Once this API returns, the client can do whatever with `fd`. If `fd` gets closed, the TEE shared memory object remains valid in libteec and TEE driver as TEE driver has a ongoing refcount increment on the dmabuf object referred by `fd`. 

The registered shared memory object is released and the refcount on original dmabuf reference is decremented only once user client has closed the registered shared memory object (from `TEEC_ReleaseSharedMemory()` or from a crash).

For example, following as valid client implementations.
Immpl#1
- alloc memory from kernel and get a `fd`
- `TEEC_RegisterSharedMemoryFileDescriptor(..., shm, fd);
`- invoke a TA using `shm` as a memref parameter
- `TEEC_ReleaseSharedMemory(shm);`             <----- shm object is released here
- `close(fd);`                <---------------------------------------- physical buffer is released here

Impl#2
- alloc memory from kernel and get a `fd`
- `TEEC_RegisterSharedMemoryFileDescriptor(..., shm, fd);
`- close(fd);
- invoke a TA using `shm` as a memref parameter
- `TEEC_ReleaseSharedMemory(shm);`         <----------- shm object and physical buffer are freed here

etienne-lms: that's right, all 'secure shared buffer' handling is out of the GP TEE Client API (and TEE Core Internal API) specifications!

jenswi-linaro: Thanks for the explanation, a short version of this (that the client is expected to close the `fd` at some point, but could do it before calling `TEEC_ReleaseSharedMemory()`)  should be in the description of the function

etienne-lms: ok, i'll add a short description.

libteec/include/linux/tee.h
```diff
@@ -254,7 +283,7 @@ struct tee_ioctl_open_session_arg {
  * tee_ioctl_open_session_arg followed by any array of struct
  * tee_ioctl_param
  */
-#define TEE_IOC_OPEN_SESSION	_IOR(TEE_IOC_MAGIC, TEE_IOC_BASE + 2, \
+#define TEE_IOC_OPEN_SESSION	_IOR(TEE_IOC_MAGIC, TEE_IOC_BASE + 3, \
```
jenswi-linaro: No renumbering please, this is supposed to be a somewhat stable ABI

etienne-lms: I don't like having such magic number incremented in strange order.
So I will enumerate the cmd ids at begining of the header file and define the new ioct "register fd" (with newly supported id 8) next to the "allocate'shm" ioctl.

jenswi-linaro: Neither do I but the ABI has to be stable and as this is a read-mostly file it's better it's easy to read than to increase ids.

libteec/src/tee_client_api.c
```diff
@@ -630,9 +631,34 @@ TEEC_Result TEEC_RegisterSharedMemory(TEEC_Context *ctx, TEEC_SharedMemory *shm)
...
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	if (!shm->flags || (shm->flags & ~(TEEC_MEM_INPUT | TEEC_MEM_OUTPUT)))
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	memset(&data, 0, sizeof(data));
+	data.fd = fd;
+	rc = ioctl(ctx->fd, TEE_IOC_SHM_REGISTER_FD, &data);
+	if (rc < 0)
+		return rc;
```
jenswi-linaro: This is not a useful TEEC_Result value.

jbech-linaro: +1, also wouldn't it be more accurate to name "rc" "fd" instead? Since it's the file descriptor you get back?

etienne-lms:
fixes from comments.
added a commit for sync with linux tee.h (sync with v12)

libteec/src/tee_client_api.c
```diff
@@ -630,9 +631,34 @@ TEEC_Result TEEC_RegisterSharedMemory(TEEC_Context *ctx, TEEC_SharedMemory *shm)
...
+TEEC_Result TEEC_RegisterSharedMemoryFileDescriptor(TEEC_Context *ctx,
+						    TEEC_SharedMemory *shm, int fd)
+{
+	struct tee_ioctl_shm_register_fd_data data;
+	int rfd;
+
+	if (!ctx || !shm || fd < 0)
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	if (!shm->flags || (shm->flags & ~(TEEC_MEM_INPUT | TEEC_MEM_OUTPUT)))
```
etienne-lms: i would like to change something here.
Caller shall not set `flags`. API inits `flags` to IN+OUT.
Even if direction can be meaningless for such buffer, current API requires flags. This allows buffer to be used as IN, OUT and INOUT memref parameters.

jenswi-linaro: Why, wouldn't it be better be as like `TEEC_RegisterSharedMemory()` as possible? It would be a more homogenous interface at least.

etienne-lms:
It seems some comments from jens got lost.
Question was (more or less)

> shouldn't it be better to have "flags" defined by API caller, for consistency with `TEEC_AllocateSharedMemory()`?

My feeling: buffer is secure, caller may not know if it is in or out. And that i've stated in  the API description that the `TEEC_SharedMemory` structure memory area (arg#2) may be left uninitialized before calling the `TEEC_RegisterShared...()`. 
The fact is that at least one of the IN or OUT flags must be set otherwise the open-session and invoke-command APIs fail.
Either caller set them, or the api set them.

jenswi-linaro:
The client has to have some expectations on the shared memory since it's going to be used in invocations etc. If the client specifies required directions of the memory the function call can fail directly if the file descriptor can't support the requested directions.

etienne-lms:
ok, i will restore `flags` as being input arguments for the registering.
currently the client and driver will not check whether the buffer can be an 'input' and/or 'output' towards a TA.

etienne-lms:
squashed, rebased.

PR #62, opened 2016-11-02
build: fix the "make O=$(out_dir)" error

prime-zeng:
Signed-off-by: Zeng Tao <prime.zeng@hisilicon.com>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -39,7 +39,6 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 CFG_TA_TEST_PATH ?= 1
 
 # Default out dir.
-# Must be a relative path with respect to the op-tee-client root directory
 O               ?= $(CURDIR)/out
 export O
 
--- a/config.mk
+++ b/config.mk
@@ -40,7 +40,7 @@ CFG_TA_TEST_PATH ?= 1
 
 # Default out dir.
 # Must be a relative path with respect to the op-tee-client root directory
-O               ?= out
+O               ?= $(CURDIR)/out
 export O
 
 #########################################################################
--- a/libsqlfs/Makefile
+++ b/libsqlfs/Makefile
@@ -6,7 +6,7 @@
 include ../flags.mk
 
 O ?= out
-OUT_DIR	?= ${CURDIR}/../$(O)/libsqlfs
+OUT_DIR	?= $(O)/libsqlfs
 
 .PHONY: all libsqlfs clean
 
@@ -39,7 +39,7 @@ SQLFS_CFLAGS := $(addprefix -I, $(SQLFS_INCLUDES)) \
 	-Wno-missing-format-attribute
 
 SQLFS_LFLAGS := -lpthread
-SQLFS_LIBS := $(CURDIR)/../$(O)/libsqlite3/libsqlite3.a
+SQLFS_LIBS := $(O)/libsqlite3/libsqlite3.a
 SQLFS_LIBRARY := $(OUT_DIR)/$(LIB_MAJ_MIN)
 
 libsqlfs: $(SQLFS_LIBRARY)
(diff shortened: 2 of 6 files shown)


config.mk
```diff
@@ -40,7 +40,7 @@ CFG_TA_TEST_PATH ?= 1
 
 # Default out dir.
 # Must be a relative path with respect to the op-tee-client root directory
```
jforissier: Remove this comment, because with your changes it is an absolute path.

libsqlfs/Makefile
```diff
@@ -6,7 +6,7 @@
 include ../flags.mk
 
 O ?= out
-OUT_DIR	?= ${CURDIR}/../$(O)/libsqlfs
+OUT_DIR	?= $(O)/libsqlfs
```
jforissier: You're changing the default behavior (when O is not given on the command line). It seems you need to set `O ?= $(CURDIR)/../out`, no?
Same in the 3 other Makefiles below.

config.mk
prime-zeng: OK

libsqlfs/Makefile
prime-zeng: In fact, what i want to do is that when i use the O option for make, the build binaries will be put into that directory, relative path to the optee-client root directory.

prime-zeng: O ?= $(CURDIR)/out is set in the main Makefile and exported, so here is right.

jforissier: OK, so remove `O ?= out` above. It is confusing.

jenswi-linaro:
I don't agree that it's more flexible to convert supplied relative paths into absolute paths.
If you don't like relative paths, why use them to start with?

jforissier:
I think the point of this PR is simply to allow absolute paths. With the current code, `make O=/tmp/out` will fail because for `OUT_DIR` in the sub-makefiles we prepend stuff to `$(O)`.

I think what we want is:
```
$ cd /path/to/optee_client
$ make            # Shall create /path/to/optee_client/out and compile to /path/to/optee_client/out/{libteec,libsqlfs,...}
$ make O=foo      # Shall create /path/to/optee_client/foo and compile to /path/to/optee_client/foo/{libteec,libsqlfs,...}
$ make O=/tmp/out # Shall create /tmp/out and write output to /tmp/out/{libteec,libsqlfs,...}
```

The proposed patch does not pass all the above tests, so it's not good.

jenswi-linaro:
OK, misunderstanding by me then. As long as we still support relative paths without conversion into absolute paths I'm OK with it.

prime-zeng:
I have rename the outdir to support make O=/path/to/optee-client, if don't rename, the clean will remove the source. 
but still one problem: make O=~  can't be supported, any suggestion?

jforissier:
@prime-zeng renaming the directories is not the right solution. If the clean rules are broken (I think they are), they need to be fixed. I mean, "rm -rf" should probably never be used in a Makefile ; instead only object files etc. should be removed, and possibly the output directory if empty.

prime-zeng:
@jforissier agree, so we need to fix the clean behavior in the Makefile first, another patch to do this?

jforissier:
@prime-zeng yes, cleaning "make clean" has to be a separate commit. In this PR or in a different one, it doesn't matter.

prime-zeng:
The clean rules has been fixed, but still one problem, make O=~ is not supported, and i use the ifeq($(shell dirname$(O)) to distinguish relative and absolute path, i don't know if  there is a better way to do it.

jenswi-linaro:
Absolute paths starts per definition with `/`. `~` is supposed to be expanded by the shell. In `optee_os/mk/subdir.mk` we test for absolute path in `$(1)` with `ifneq ($(filter /%,$(1)),)`.

jforissier:
Created pull request https://github.com/OP-TEE/optee_client/pull/66. @prime-zeng your comments are welcome, let me know if it works for you.

PR #63, opened 2016-11-10
TEE Latency benchmark

igoropaniuk:
Add support for transparently bypassing ringbuffer as a fifth param

Relates to:
https://github.com/linaro-swg/linux/pull/27
https://github.com/OP-TEE/build/pull/81
https://github.com/OP-TEE/optee_test/pull/145
https://github.com/OP-TEE/optee_os/pull/1175


libteec/src/tee_client_api.c
```diff
@@ -573,6 +622,25 @@ TEEC_Result TEEC_InvokeCommand(TEEC_Session *session, uint32_t cmd_id,
...
+#ifdef CFG_TEE_PRINT_LATENCY_STAT
+	struct tee_ringbuf *ringb = (struct tee_ringbuf *)ringbuf_shm.buffer;
+
+	for (uint32_t ts_i = 0; ts_i < ringb->tm_ind; ts_i++)
+		printf("Cycle count = %"PRIu64"\tSrc = %s\tPC = 0x%"PRIx64"\n",
+				(ringb->stamps[ts_i].cnt),
+				bench_str_src(ringb->stamps[ts_i].src),
+				(ringb->stamps[ts_i].addr));
+#endif /* CFG_TEE_BENCHMARK_STAT */
+	if (allocated)
```
jenswi-linaro: This can't be anything but true at this stage, right?

public/tee_bench.h
```diff
@@ -0,0 +1,97 @@
...
+{
+
+	uintptr_t pc;
+
+	asm volatile("mov %0, r15" : "=r"(pc));
+	return pc;
+}
+
+/* Cycle counter */
+static inline __attribute__((always_inline)) uint64_t read_ccounter(void)
```
jenswi-linaro: Why the `__attribute__((always_inline))`?

public/tee_bench.h
```diff
@@ -0,0 +1,97 @@
...
+	uint64_t ts_i;
+
+	if (!ringb)
+		return;
+	if (ringb->tm_ind >= TEE_BENCH_MAX_STAMPS)
+		ringb->tm_ind = 0;
+
+	ts_i = ringb->tm_ind++;
+	ringb->stamps[ts_i].cnt = read_ccounter();
+	ringb->stamps[ts_i].addr = read_pc();
```
jenswi-linaro: Making sure that this function is never inlined instead and returning __builtin_return_address(0) instead is more portable and gives more or less the same result.

public/tee_bench.h
```diff
@@ -0,0 +1,97 @@
...
+	ringb->stamps[ts_i].src = source;
+}
+#else /* CFG_TEE_BENCHMARK */
+static inline __attribute__((always_inline)) void tee_add_timestamp
```
jenswi-linaro: Why the `__attribute__((always_inline))`?

public/tee_bench.h
igoropaniuk: right, not needed

public/tee_bench.h
igoropaniuk: In case it wasn't inlined by the compiler we would receive in ringb->stamps[ts_i].addr the address within this function (`tee_add_timestamp`), not the caller function (where `tee_add_timestamp` is used). And it's a bit useless for our purposes

To avoid inlining here we can use for example __builtin_return_address(**1**) for (which you described in the previous comment, but address of the caller), but I'm not sure. What do you think?

jenswi-linaro: In this ifdef path this function doesn't do anything.

igoropaniuk: ah, sorry. now I got what you meant :)

libteec/src/tee_client_api.c
igoropaniuk: For now - right.
I've added it, because now I'm finishing implementation NW apps(I hope to commit it today) that will calculate variance for specific cases, and this ringbuffer will be allocated from NW app, not here (additional check for already allocated ringbuffer to if(operation) will be added)

jenswi-linaro: OK

libteec/src/tee_client_api.c
```diff
@@ -49,6 +50,55 @@
 /* How many device sequence numbers will be tried before giving up */
 #define TEEC_MAX_DEV_SEQ	10
 
+
+#if defined(CFG_TEE_BENCHMARK) && defined(CFG_TEE_PRINT_LATENCY_STAT)
+static const char *bench_str_src(uint64_t source)
```
jbech-linaro: Could we move things related to bench-marking into a separate c-file? The `tee_client_api.c` file is rather clean otherwise. It's maybe even possible to break out the bench marking code in the function `TEEC_InvokeCommand` below?

libteec/src/tee_client_api.c
```diff
@@ -49,6 +50,55 @@
...
+		return "???";
+	}
+}
+
+static void print_latency_info(void *ringbuffer)
+{
+	struct tee_ringbuf *ringb = (struct tee_ringbuf *)ringbuffer;
+	uint64_t start = 0;
+
+	printf("Latency information:\n");
```
jbech-linaro: Maybe `DMSG` etc clutter down to much in this case, but otherwise that is an alternative here too.

libteec/src/tee_client_api.c
```diff
@@ -539,6 +588,36 @@ TEEC_Result TEEC_InvokeCommand(TEEC_Session *session, uint32_t cmd_id,
...
+	uint32_t allocated = 0;
+
+	TEEC_SharedMemory ringbuf_shm = {
+		.flags = TEEC_MEM_INPUT | TEEC_MEM_OUTPUT,
+		.buffer = NULL,
+		.size = TEE_BENCH_RB_SIZE
+	};
+
+	if (operation) {
+
```
jbech-linaro: Unnecessary space (IMHO).

libteec/src/tee_client_api.c
```diff
@@ -539,6 +588,36 @@ TEEC_Result TEEC_InvokeCommand(TEEC_Session *session, uint32_t cmd_id,
...
+		TEEC_Result res_ring = TEEC_AllocateSharedMemory
+				(session->ctx, &ringbuf_shm);
+
+		if (res_ring != TEEC_SUCCESS)
+			goto out;
+
+		allocated = 1;
```
jbech-linaro: Do you actually need this parameter? Wouldn't it be enough to look at `ringbuf_shm.buffer`?

public/tee_bench.h
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) 2014, Linaro Limited
```
jbech-linaro: 2016

public/tee_bench.h
```diff
@@ -0,0 +1,101 @@
...
+ * GNU General Public License for more details.
+ *
+ */
+
+#ifndef TEE_BENCH_H
+#define TEE_BENCH_H
+
+#include <inttypes.h>
+
+#define UNUSED(x) (void)(x)
```
jbech-linaro: I would like to add the same comment here as I did in the optee_os patch, but ... here it works for now, since we currently don't have the same defines here.

By the way, is this file an identical copy of the one in optee_os?

libteec/src/tee_client_api.c
igoropaniuk: sure

libteec/src/tee_client_api.c
igoropaniuk: Relates to https://github.com/OP-TEE/optee_client/pull/63#discussion_r87548667
In case I'm going to allocate/deallocate a timestamp ringbuffer within NW (for performing benchmarks)

public/tee_bench.h
igoropaniuk: > By the way, is this file an identical copy of the one in optee_os?

yes, it's the same

libteec/Makefile
```diff
@@ -28,13 +32,20 @@ TEEC_CFLAGS	:= $(addprefix -I, $(TEEC_INCLUDES)) $(CFLAGS) -D_GNU_SOURCE \
 		   -DDEBUGLEVEL_$(CFG_TEE_CLIENT_LOG_LEVEL) \
 		   -DBINARY_PREFIX=\"TEEC\"
 
+ifeq ($(CFG_TEE_BENCHMARK),y)
+TEEC_CFLAGS	+= -DCFG_TEE_BENCHMARK
+endif
+
+ifeq ($(CFG_TEE_PRINT_LATENCY_STAT),y)
```
jforissier: Please add a comment to explain what this does (print latency measurements for each `TEEC_Invoke()`).

PR #66, opened 2016-11-25
Allow O to be any kind of path (absolute, relative, with ~)

jforissier:
Tested with:
  $ export CFG_WITH_SQLFS=y
  $ make
  $ make O=build
  $ make O=/tmp/build
  $ make O=~/build

Suggested-by: Zeng Tao <prime.zeng@hisilicon.com>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -38,10 +38,12 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 # such as 1008.5 that test loading of corrupt TAs.
 CFG_TA_TEST_PATH ?= 1
 
-# Default out dir.
-# Must be a relative path with respect to the op-tee-client root directory
+# Default output directory.
+# May be absolute, or relative to the optee_client source directory.
 O               ?= out
-export O
+
+# To be used instead of $(O) in sub-directories
+OO := $(if $(filter /%,$(O)),$(O),$(CURDIR)/../$(O))
 
 #########################################################################
 # Private Values                                                        #
--- a/libsqlfs/Makefile
+++ b/libsqlfs/Makefile
@@ -4,9 +4,9 @@
 # and/or linking it statically! See COPYING section 6.
 
 include ../flags.mk
+include ../config.mk
 
-O ?= out
-OUT_DIR	?= ${CURDIR}/../$(O)/libsqlfs
+OUT_DIR := $(OO)/libsqlfs
 
 .PHONY: all libsqlfs clean
 
@@ -39,10 +39,11 @@ SQLFS_CFLAGS := $(addprefix -I, $(SQLFS_INCLUDES)) \
 	-Wno-missing-format-attribute
 
 SQLFS_LFLAGS := -lpthread
-SQLFS_LIBS := $(CURDIR)/../$(O)/libsqlite3/libsqlite3.a
+SQLFS_LIBS := $(OUT_DIR)/../libsqlite3/libsqlite3.a
 SQLFS_LIBRARY := $(OUT_DIR)/$(LIB_MAJ_MIN)
 
 libsqlfs: $(SQLFS_LIBRARY)
+	$(VPREFIX)mkdir -p $(OUT_DIR)
 	$(VPREFIX)ln -sf $(SQLFS_LIBRARY) $(OUT_DIR)/$(LIB_MAJOR)
 	$(VPREFIX)ln -sf $(OUT_DIR)/$(LIB_MAJOR) $(OUT_DIR)/$(LIB_NAME)
 
--- a/libsqlite3/Makefile
+++ b/libsqlite3/Makefile
@@ -1,7 +1,7 @@
 include ../flags.mk
+include ../config.mk
 
-O ?= out
-OUT_DIR	?= ${CURDIR}/../$(O)/libsqlite3
+OUT_DIR	:= $(OO)/libsqlite3
 
 .PHONY: all libsqlite3 clean
 
--- a/libteec/Makefile
+++ b/libteec/Makefile
@@ -1,6 +1,7 @@
 include ../flags.mk
+include ../config.mk
 
-OUT_DIR		?= ${CURDIR}/../$(O)/libteec
+OUT_DIR := $(OO)/libteec
 
 .PHONY: all libteec clean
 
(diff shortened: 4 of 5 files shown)


config.mk
```diff
@@ -38,11 +38,14 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 # such as 1008.5 that test loading of corrupt TAs.
 CFG_TA_TEST_PATH ?= 1
 
-# Default out dir.
-# Must be a relative path with respect to the op-tee-client root directory
+# Default output directory.
+# May be absolute, or relative to the optee_client source directory.
 O               ?= out
 export O
```
jforissier: Yes, you're right. I'll remove this line.

config.mk
```diff
@@ -38,11 +38,14 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
...
 
+# To be used instead of $(O) in sub-directories
+OO := $(if $(filter /%,$(O)),$(O),$(CURDIR)/../$(O))
```
prime-zeng: For O=~, OO=$(CURDIR)../$(O), i think it does not work as expected.

prime-zeng:
The make clean also need to be fixed as what we have discussed in #62 , if the O=/XXX/XXX/optee-client, the clean will remove the source code.

config.mk
jforissier: You think, or you tested and you found this not to be working? ;)
As I wrote in the commit description, I have tested with `make O=~/build` and I found no issue. Which is not surprising since the tilde is expanded by the shell before make is invoked, so O is actually an absolute path. Please give more details if it's not working for you.

So, unless I missed something, I will reword the commit and not mention ~ because it's irrelevant to the Makefile.

jforissier:
@prime-zeng about the clean rule: I agree, I will make a separate PR.

config.mk
prime-zeng: make -C /home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/build/../optee_client CROSS_COMPILE=aarch64-histbv100-linux- O=~/build
make[1]: Entering directory `/home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/optee_client'
Building libteec.so
ln -sf /home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/optee_client/libteec/../~/build/libteec/libteec.so.1.0 /home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/optee_client/libteec/../~/build/libteec/libteec.so.1
ln -sf /home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/optee_client/libteec/../~/build/libteec/libteec.so.1 /home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/optee_client/libteec/../~/build/libteec/libteec.so
Building tee-supplicant
make --directory=tee-supplicant  --no-print-directory --no-builtin-variables
make[2]: Nothing to be done for `all'.
mkdir -p ~/build/export/lib ~/build/export/include ~/build/export/bin
cp ~/build/libteec/libteec.so* ~/build/export/lib
cp: cannot stat ‘/home/os_net/zengtao/build/libteec/libteec.so*’: No such file or directory
make[1]: *** [copy_export] Error 1
make[1]: Leaving directory `/home/os_net/zengtao/work/HiDessertV100R001/Gerrit/mainline/optee_client'
make: *** [optee-client-common] Error 2

prime-zeng: I tried the patch with the above error, did i miss something ?

prime-zeng: zengtao@key:build$ echo $SHELL
/bin/bash
zengtao@key:build$ echo ~
/home/os_net/zengtao
zengtao@key:build$

jforissier: Weird. Is the make command above run from a makefile/shell script maybe? Then please try to replace `make -C ...` with `echo $SHELL ; echo ~ ; false` and run the build again. This must be that the shell is not expanding the tilde.

lackan: It seems the link also has to change. Because it has the below output:
LINK    /xxx/optee_client/tee-supplicant/../~/build/tee-supplicant/tee-supplicant
and the so also have the same output.

jforissier: Then I don't understand why ~ is not expanded in your build log.
We could parse `$(O)` and expand the ~ in `config.mk`, but I don't really like this option because it's not the job of the Makefile to do that (and it's potentially complex -- should we also handle `~login` for instance? or more exotic forms like `~+` and `~-`?).

prime-zeng: what is your gnu make version,  maybe some version of make can't expand the ~. 
I am using GNU Make 3.81.

jforissier: I'm using GNU Make 4.1, but I have also tried 3.81 and it's OK. Can you check the value of O in `config.mk` with the following patch?

```diff
diff --git a/config.mk b/config.mk
index 2cffc55..689f6ac 100644
--- a/config.mk
+++ b/config.mk
@@ -41,6 +41,7 @@ CFG_TA_TEST_PATH ?= 1
 # Default output directory.
 # May be absolute, or relative to the optee_client source directory.
 O               ?= out
+$(info O=$(O))
 
 # To be used instead of $(O) in sub-directories
 OO := $(if $(filter /%,$(O)),$(O),$(CURDIR)/../$(O))
```

Here is what I get:
```
$ make O=~/build build-libteec
O=/home/jerome/build
Building libteec.so
O=/home/jerome/build
  CC      src/tee_client_api.c
  CC      src/teec_trace.c
  LINK    /home/jerome/build/libteec/libteec.so.1.0
```

jforissier:
@prime-zeng https://github.com/OP-TEE/optee_client/pull/67

config.mk
prime-zeng: @jforissier ^_^，i just want to make the solution complete.So i think we have two choices:
1. Support ~, and we need to get the realpath from O(for example, using realpath).
2. Don't support ~, and refuse the ~ if the shell don't expand it.

prime-zeng: @jforissier  ok, i got it,  ( ˇˍˇ )，i have got no good idea now if we want to address a  perfect solution.

jforissier:
Ping?
FYI, I will reword the commit comment to read "Allow O to be an absolute or relative path" and mention that tilde expansion is assumed to be done by the calling shell (since we discussed a lot about this point).

prime-zeng:
@jforissier  agree.

PR #67, opened 2016-11-28
Fix clean

jforissier:
Please ignore the first two commits, they are from another pending PR, and consider only the last one: "Update clean rules".


etienne-lms:
3rd commit "Update clean rules": nice, reviewed ok.

jforissier:
Rebased on master (the two other commits were not actually needed) and applied Etienne's R-b.

PR #69, opened 2016-12-06
README: Update legacy toolchain download url

vchong:

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,18 @@ charge.
 
 ## Get and build the software
 ### Get the compiler
-We will strive to use the latest available compiler from Linaro. Start by
-downloading and unpacking the compiler. Then export the PATH to the bin folder.
+We will strive to use the latest available compiler from Linaro. This
+[file](https://github.com/OP-TEE/build/blob/master/toolchain.mk) refers to a
+list of toolchains we have used and know are working in our setups. Start by
+downloading and unpacking a compiler from it. Then export the PATH to the bin
+folder.
 
 	$ cd $HOME
 	$ mkdir toolchains
 	$ cd toolchains
-	$ wget http://releases.linaro.org/14.05/components/toolchain/binaries/gcc-linaro-arm-linux-gnueabihf-4.9-2014.05_linux.tar.xz
-	$ tar xvf gcc-linaro-arm-linux-gnueabihf-4.9-2014.05_linux.tar.xz
-	$ export PATH=$HOME/toolchains/gcc-linaro-arm-linux-gnueabihf-4.9-2014.05_linux/bin:$PATH
+	$ wget [url/to/gcc_tarball]
+	$ tar xvf [gcc_tarball]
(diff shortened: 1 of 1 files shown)


vchong:
I checked all the other repos and think this and https://github.com/OP-TEE/optee_os/pull/1207 are the last ones for this issue. Hopefully I didn't miss anything and release team doesn't change file locations anymore.

jenswi-linaro:
I think we're giving the url to the toolchain in too many places. Do we really need to have it anywhere except https://github.com/OP-TEE/build/blob/master/toolchain.mk ?

vchong:
Yes, scripts are broken all over the place in other groups too. It is a pain. So replace the text with something general like `Download the appropriate toolchain from http://releases.linaro.org/components/toolchain/binaries/` and `/path/to/toolchain` in the instructions? Any other suggestions?

jforissier:
@vchong sounds good.

jbech-linaro:
@vchong instead of referring to `release.l.o`, what about instead referring to the the toolchains in `toolchain.mk` as suggested by Jens. By doing that, we know that we're referring to toolchains we have used and that we know are working in our setups?

vchong:
Yes ok, good idea. Will make changes to refer to `toolchain.mk`.

vchong:
Updated. Please see if new changes are ok. I'll also change commit title to `README: Replace toolchain URL with reference to toolchain.mk` when adding tags.

jbech-linaro:
Yeps, that is great, thanks!

vchong:
You're welcome!
Commits squashed and tag added.

jforissier:
Why did you split the subject line?

vchong:
The vim editor follows the 50/72 rule and 'warns' me (by changing text color) after the 50 chars. Should I just ignore it and put everything on 1 line?

jforissier:
50 characters? That's not much for a commit subject. `checkpatch.pl` warns for >75 if I'm not mistaken. Actually, I have run some stats on the Linux kernel. If you consider all commit subjects in the last 6 months (except merges), 42.04% are 50 characters or less, 54.33% are between 51 and 75 characters, and 3.63% are more than 76 characters. So, yes, please merge the first two lines ;)

jbech-linaro:
Here is a nice image showing the distribution (peeks around 50 characters).
https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c#.nm9p5nby4

vchong:
Lines merged. Thanks!

jforissier:
Yeah thanks for the link, nice graphical distribution. Personally, I don't approve the 50 rule, and it's good that checkpatch actually warns only for >75 (here: https://github.com/torvalds/linux/blob/master/scripts/checkpatch.pl#L2508). I'm OK with the 72 rule in the commit body, though (although by the same checkpatch standards we could do 75... not a big difference).

Anyway, thanks @vchong, patch merged.

PR #72, opened 2016-12-13
tee_client_api: register user memory

lorc:
This patch is the userspace part of dynamic shared memory in OP-TEE (refer to https://github.com/OP-TEE/optee_os/pull/1232). This patch is authred by @jenswi-linaro, I just tidied it a bit and rebased on current master.

This patch utilizes new ioctl that allows client application to share any buffer with OP-TEE.

### DIFF
--- a/libteec/include/linux/tee.h
+++ b/libteec/include/linux/tee.h
@@ -49,6 +49,7 @@
 #define TEE_MAX_ARG_SIZE	1024
 
 #define TEE_GEN_CAP_GP		(1 << 0)/* GlobalPlatform compliant TEE */
+#define TEE_GEN_CAP_REG_MEM	(1 << 2)/* Supports registering shared memory */
 
 /*
  * TEE Implementation ID
@@ -143,6 +144,36 @@ struct tee_ioctl_shm_register_fd_data {
  */
 #define TEE_IOC_SHM_REGISTER_FD	_IOWR(TEE_IOC_MAGIC, TEE_IOC_BASE + 8, \
 				     struct tee_ioctl_shm_register_fd_data)
+
+/**
+ * struct tee_ioctl_shm_register_data - Shared memory register argument
+ * @addr:      [in] Start address of shared memory to register
+ * @length:    [in/out] Length of shared memory to register
+ * @flags:     [in/out] Flags to/from registration.
+ * @id:                [out] Identifier of the shared memory
+ *
+ * The flags field should currently be zero as input. Updated by the call
+ * with actual flags as defined by TEE_IOCTL_SHM_* above.
+ * This structure is used as argument for TEE_IOC_SHM_REGISTER below.
+ */
+struct tee_ioctl_shm_register_data {
+	__u64 addr;
+	__u64 length;
+	__u32 flags;
+	__s32 id;
+};
+
+/**
+ * TEE_IOC_SHM_REGISTER - Register shared memory
+ *
+ * Registers shared memory between the user space process and secure OS.
+ *
+ * Returns a file descriptor on success or < 0 on failure
+ *
+ * The shared memory is unregisterred when the descriptor is closed.
+ */
+#define TEE_IOC_SHM_REGISTER   _IOWR(TEE_IOC_MAGIC, TEE_IOC_BASE + 9, \
+				     struct tee_ioctl_shm_register_data)
 
 /**
  * struct tee_ioctl_buf_data - Variable sized buffer
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -75,11 +75,14 @@ struct tee_shm {
 	int id;
 	void *p;
 	size_t size;
+	bool registered;
+	int fd;
 	struct tee_shm *next;
 };
 
 struct thread_arg {
 	int fd;
+	uint32_t gen_caps;
 	bool abort;
 	size_t num_waiters;
 	pthread_mutex_t mutex;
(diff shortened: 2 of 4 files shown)


etienne-lms:
why registering the shared memory to the optee secure world ?

lorc:
@etienne-lms, Idea is not to use preshared pool. There are many reasons:

- Current carveout have size of 1M on some platforms. This can be not enough in some use cases.
- TEEC_RegisterSharedMemory() uses shadow buffer and copies data on every command invocation back and forth.
- Preshared pool is not compatible with virtualization, because there are no easy way to split pool between VMs.

If client can use part of own memory for shared buffers, than all issues above can be solved easily.
I'm proposing this changes mainly because I want to use OP-TEE in virtualized environment.

etienne-lms:
ok, i see your idea. And it is true that many buffer parameters can be provided to TAs without copies to the current single physically contiguous 'non secure shared memory area'.

I wonder why nonsecure should register shared buffers to the secure side.

If non-secure wishes to check if a buffer is ok to be used as shared mem, then it's fine, but this is not mandatory and waste cpu cycles. Non-secure could simply invoke the secure side with the memref parameters.

On secure side, we don't care if non-secure presents a shared memory area as "non-secure". Secure side will simply verify the shared memory area conformance against TA memref permissions. And maps it accordingly.

As far a nonsecure handle shared memory allocation, it is nonsecure world responsibility to track these memref, not the secure side.

jenswi-linaro:
For a one shot invocation it's more efficient to do as you say @etienne-lms, but then for a single invocation performance isn't as critical as there's so much else that eats CPU time, checking the RSA signature of the TA for instance.

If we make multiple invocations it quickly pays off to have the shared memory registered in advance as it's represented by a single handle instead of a long list of physical pages.

Another advantage with always using registered memory is that we then can get rid of yet another way of describing a chunk of memory.

igoropaniuk:
@etienne-lms 
As I understand (based on @lorc comments and code) it's just some way of extending of existing preshared pool, which used for bypassing GP params, but with ability to work with this shared memory in more generic way (comparing wth GP params). Purposes are almost the same + we can avoid some restrictions of GP specification(amount of params, param size restrictions), and, as @lorc mentioned, unnesessary copying back & forth.

And, speaking about my benchmarking task it's kind of a life saver. I can easily, with minimum amount of code changes, move to this SHM API instead of using GP params :)

igoropaniuk:
`Reviewed-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

etienne-lms:
late feedback... nice indeed.

jbech-linaro:
Same here, late reply, all good. But I'd be nice to have our https://github.com/OP-TEE/optee_os/blob/master/documentation/optee_design.md#7-shared-memory updated to reflect the changes (and the current implementation).

jforissier:
Hi @lorc, would you please rebase this patch on top of the current master branch and add the R-b tags so that I can pick it up?

lorc:
@jforissier I planed to wait till corresponding patches to OP-TEE kernel and OP-TEE driver will be ready for merge. But, actually you are right. This patch can be merged at any time. It is rebased on current `master` now. Thanks.

jforissier:
@lorc thanks for rebasing. Actually, I'm not sure I want to merge this immediately since it depends on PRs that are still pending in optee_os and the linux driver. I'd rather wait until all 3 PRs are reviewed. Sorry I didn't realize this when I wrote the above ;)

lorc:
Hello. I have rebased it again. Also I have added new patch that makes supplicant to register own memory instead of allocating it thru SHM pool.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -339,13 +394,18 @@ static uint32_t process_free(size_t num_params, struct tee_ioctl_param *params)
...
-		     shm->p, shm->size, strerror(errno));
-		free(shm);
-		return TEEC_ERROR_BAD_PARAMETERS;
+	if (shm->registered) {
+		free(shm->p);
+	} else  {
+		if (munmap(shm->p, shm->size) != 0) {
+			EMSG("munmap(%p, %zu) failed - Error = %s",
+			     shm->p, shm->size, strerror(errno));
+			free(shm);
```
etienne-lms: this should also `close(shm->fd);`

tee-supplicant/src/tee_supplicant.c
```diff
@@ -99,6 +101,8 @@ static struct tee_shm *shm_head;
 
 static const char *ta_dir;
 
+static uint32_t gen_caps = 0;
```
etienne-lms: minor: useless init

tee-supplicant/src/tee_supplicant.c
```diff
@@ -284,41 +288,92 @@ static uint32_t load_ta(size_t num_params, struct tee_ioctl_param *params)
...
+static uint32_t process_alloc(int fd, size_t num_params,
+			      struct tee_ioctl_param *params)
+{
+	struct tee_ioctl_param_value *val;
+	struct tee_shm *shm;
+
+	if (num_params != 1 || get_value(num_params, params, 0, &val))
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	if (gen_caps & TEE_GEN_CAP_REG_MEM)
```
jenswi-linaro: Wouldn't it be possible to rely on `ctx->reg_mem` instead?

lorc: @jenswi-linaro I'm not sure that I understood you. `ctx->reg_mem` is a kernel thing, isn't it?

jenswi-linaro: Sorry is was confusing this with how we use `TEEC_Context` in libteec.

Using global variables should in general be avoided, this `gen_caps` should be stored together with the file descriptor as `gen_caps` is holds information specific for this file descriptor.

lorc:
@jenswi-linaro, I want ask you about interface between optee client and kernel driver.
Currently we have `TEE_IOC_SHM_ALLOC`, then I have added `TEE_IOC_SHM_REGISTER`. Now I want to share `tmem` buffers. And there are two ways. I can add ` TEE_IOC_SHM_ADD_TMEM_BUF` or:
I can rename `TEE_IOC_SHM_REGISTER` to `TEE_IOC_SHM_ADD` and use flags to select desired operation: either register shared buffer in OP-TEE core, or register it in `tee` framework as a temporary buffer. Which way do you prefer?

jenswi-linaro:
So the difference between `TEE_IOC_SHM_REGISTER` and `TEE_IOC_SHM_ADD_TMEM_BUF` is that the latter is not registered in secure world and has to be represented as tmem when invoking for instance.

Why would we need `TEE_IOC_SHM_ADD_TMEM_BUF`? To me it seems like something user space shouldn't need to worry about.

lorc:
> So the difference between `TEE_IOC_SHM_REGISTER` and `TEE_IOC_SHM_ADD_TMEM_BUF` is that the latter is not registered in secure world and has to be represented as tmem when invoking for instance?

Exactly.

> Why would we need TEE_IOC_SHM_ADD_TMEM_BUF? To me it seems like something user space shouldn't need to worry about.

Because current `userspace<->kernelspace` interface for TA parameter passing uses `shm_id` for memory references. My idea is to call `TEE_IOC_SHM_ADD_TMEM_BUF` to obtain `shm_id`. Another approach is to add new type for `struct tee_ioctl_param.attr`. Something like `TEE_IOCTL_PARAM_ATTR_TYPE_TMPMEMREF_*`

jenswi-linaro:
Why do we need a way for userspace to control if buffers passed to secure world are registered in secure world or not?

lorc:
> A Shared Memory block can be registered or allocated once and then used in multiple Commands, and even in multiple Sessions, provided they exist within the scope of the TEE Context in which the Shared Memory was created. This pre-registration is typically more efficient than registering a block of memory using temporary registration if that memory buffer is used in more than one Command invocation.

I thought that `TEEC_RegisterSharedMemory()` and `TEEC_AllocateSharedMemory()` will register buffers in secure world (by using `OPTEE_MSG_CMD_REGISTER_SHM` SMC), while temporary memory buffers will not be registered and will be passed as list of pages every time TA function is invoked.

jenswi-linaro:
OK, if I understand it correctly it's something that doesn't need to be done now.

It sounds like some optimization that could be useful but we need to know that it's actually is needed before we do something about it as it complicates things further a tiny bit. But we should right now make sure that if we choose to implement this later that the current ABI can be easily extended to deal with it. The flags field in `struct tee_ioctl_shm_register_data` seems suitable for this.

lorc:
Ah, I see. Yes. You can think about this as an optimization. It needs extensions in `tee` core and in client library, but it is not necessary right now. So yes, we actually can postpone this for now. 

I have tested ABI between NW and SW and it supports this use-case. I can share my patches for client library, kernel driver and `xtest`  if you also want to test it.

jenswi-linaro:
Right now I'd like to focus on getting the current stuff in place.
Once that's in place I'd be very interested in taking a look and test it.

lorc:
Hello.

I rebased this PR onto current master. There was conflicts in `#include` part of the files. So I have moved `#include <stdbool.h>` into public header during conflict resolving. You might want to review this. I can't make this change as `[review]` patch, because it is product of conflict resolution during rebase.

Also I have addressed @jenswi-linaro's comment regarding global `gen_caps` variable in supplicant code. It is in the separate patch.

lorc:
Pushed fix for `!!()`, also added Jens's R-b tag.

libteec/include/linux/tee.h
```diff
@@ -49,6 +49,7 @@
 #define TEE_MAX_ARG_SIZE	1024
 
 #define TEE_GEN_CAP_GP		(1 << 0)/* GlobalPlatform compliant TEE */
+#define TEE_GEN_CAP_REG_MEM	(1 << 2)/* Supports registering shared memory */
```
etienne-lms: as you are so patient, would you mind adding a tabulation before these comments.

libteec/include/linux/tee.h
```diff
@@ -145,6 +146,36 @@ struct tee_ioctl_shm_register_fd_data {
...
+ */
+struct tee_ioctl_shm_register_data {
+	__u64 addr;
+	__u64 length;
+	__u32 flags;
+	__s32 id;
+};
+
+/**
+ * TEE_IOC_SHM_REGISTER - Register shared memory argument
```
etienne-lms: minor: remove `argument`

tee-supplicant/src/tee_supplicant.c
```diff
@@ -283,41 +286,92 @@ static uint32_t load_ta(size_t num_params, struct tee_ioctl_param *params)
...
+	if (num_params != 1 || get_value(num_params, params, 0, &val))
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	if (arg->gen_caps & TEE_GEN_CAP_REG_MEM)
+		shm = register_local_shm(arg->fd, val->b);
+	else
+		shm = alloc_shm(arg->fd, val->b);
+
+	if (!shm) {
+		return TEEC_ERROR_OUT_OF_MEMORY;
```
etienne-lms: minor: remove `{`/`}` ?

libteec/include/linux/tee.h
lorc: I'll do this in a separate patch, okay?

etienne-lms: as you like. (sorry to ask you to :)

lorc:
Addressed @etienne-lms's comments. Added separate patch that fixes comments in `tee.h`

BTW, I'm sorry that my English is far from perfect. Feel free to correct me, especially if my mistakes are terrible :)

etienne-lms:
Same with me, so do not hesitate to contradict/correct me.

etienne-lms:
r-b on the whole.

lorc:
Thanks, @etienne-lms. I have squashed all fixes and added your r-b tag.
@jenswi-linaro, there is a new patch. It changes only comments/formatting. Probably, you want take a look. Or I can put it in a separate PR, to finish this faster :)

jenswi-linaro:
Later is fine with me a least.

lorc:
Okay, so I dropped this patch for now. Will push it as a separate PR after this one will be merged.

PR #74, opened 2017-01-03
Gp socket support

jenswi-linaro:
Depends on https://github.com/linaro-swg/linux/pull/30

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -38,6 +38,10 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 # such as 1008.5 that test loading of corrupt TAs.
 CFG_TA_TEST_PATH ?= y
 
+# CFG_GP_SOCKETS
+# Enable Global Platform Sockets support
+CFG_GP_SOCKETS ?= y
+
 # Default output directory.
 # May be absolute, or relative to the optee_client source directory.
 O               ?= out
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -25,19 +25,21 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <rpmb.h>
-#include <tee_client_api.h>
-#include <teec_trace.h>
-#include <string.h>
-#include <netinet/in.h>
-#include <sys/ioctl.h>
+#include <fcntl.h>
 #include <linux/types.h>
 #include <linux/mmc/ioctl.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
+#include <netinet/in.h>
+#include <pthread.h>
+#include <rpmb.h>
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <tee_client_api.h>
+#include <teec_trace.h>
+#include <tee_supplicant.h>
 #include <unistd.h>
 
 #ifdef RPMB_EMU
@@ -103,6 +105,9 @@ struct rpmb_data_frame {
 #define RPMB_MSG_TYPE_RESP_AUTH_DATA_READ		0x0400
 };
 
+
+static pthread_mutex_t rpmb_mutex = PTHREAD_MUTEX_INITIALIZER;
+
 /*
  * ioctl() interface
  * Comes from: uapi/linux/major.h, linux/mmc/core.h
@@ -138,6 +143,7 @@ struct rpmb_data_frame {
 			EMSG("ioctl ret=%d errno=%d", ret, errno); \
 		ret;						   \
 	})
+
 
 /* Open and/or return file descriptor to RPMB partition of device dev_id */
 static int mmc_rpmb_fd(uint16_t dev_id)
@@ -728,12 +734,13 @@ err:
 	return res;
 }
 
+
 /*
  * req is one struct rpmb_req followed by one or more struct rpmb_data_frame
  * rsp is either one struct rpmb_dev_info or one or more struct rpmb_data_frame
  */
-uint32_t rpmb_process_request(void *req, size_t req_size, void *rsp,
-			      size_t rsp_size)
+static uint32_t rpmb_process_request_unlocked(void *req, size_t req_size,
+					      void *rsp, size_t rsp_size)
 {
 	struct rpmb_req *sreq = req;
 	size_t req_nfrm;
@@ -770,6 +777,19 @@ uint32_t rpmb_process_request(void *req, size_t req_size, void *rsp,
 		res = TEEC_ERROR_BAD_PARAMETERS;
 		break;
 	}
+
+	return res;
+}
+
+
+uint32_t rpmb_process_request(void *req, size_t req_size, void *rsp,
+			      size_t rsp_size)
+{
+	uint32_t res;
+
+	tee_supp_mutex_lock(&rpmb_mutex);
+	res = rpmb_process_request_unlocked(req, req_size, rsp, rsp_size);
+	tee_supp_mutex_unlock(&rpmb_mutex);
 
 	return res;
 }
--- a/tee-supplicant/src/tee_socket.c
+++ b/tee-supplicant/src/tee_socket.c
@@ -1,0 +1,736 @@
+/*
+ * Copyright (c) 2016-2017, Linaro Limited
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <inttypes.h>
+#include <limits.h>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <optee_msg_supplicant.h>
+#include <poll.h>
+#include <pthread.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/queue.h>
+#include <sys/socket.h>
+#include <tee_client_api.h>
+#include <teec_trace.h>
+#include <tee_socket.h>
+#include <tee_supplicant.h>
+#include <unistd.h>
+
+#include "handle.h"
+#include "__tee_isocket_defines.h"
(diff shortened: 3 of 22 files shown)


jenswi-linaro:
Rebased

tee-supplicant/src/tee_supplicant.c
```diff
@@ -356,44 +387,203 @@ static int usage(void)
...
+		ret = process_rpmb(num_params, params);
+		break;
+	case RPC_CMD_SHM_ALLOC:
+		ret = process_alloc(arg->fd, num_params, params);
+		break;
+	case RPC_CMD_SHM_FREE:
+		ret = process_free(num_params, params);
+		break;
+	case OPTEE_MSG_RPC_CMD_SOCKET:
+		ret = tee_socket_process(num_params, params);
```
jforissier: Missing #include <tee_socket.h>

tee-supplicant/src/tee_socket.c
```diff
@@ -0,0 +1,742 @@
...
+	case TEE_TCP_SET_SENDBUF:
+		if (setsockopt(fd, SOL_SOCKET, SO_SNDBUF, buf, *blen))
+			return TEEC_ERROR_BAD_PARAMETERS;
+		return TEEC_SUCCESS;
+	default:
+		return TEEC_ERROR_NOT_SUPPORTED;
+	}
+}
+
+static TEEC_Result sa_set_port(struct sockaddr *sa, socklen_t slen,
```
jforissier: Unused

tee-supplicant/src/tee_socket.c
```diff
@@ -0,0 +1,742 @@
...
+#include <tee_client_api.h>
+#include <teec_trace.h>
+#include <tee_socket.h>
+#include <tee_supplicant.h>
+#include <unistd.h>
+
+#include "handle.h"
+#include "tee_api_defines.h"
+#include "tee_isocket_defines.h"
+#include "__tee_ipsocket.h"
```
jforissier: Why the leading underscores?

jenswi-linaro: It's a header file containing definitions both for tee_tcpsocket and tee_udpsocket.h.

jforissier:
I think the socket code should be guarded by `#ifdef CFG_SOCKET` or something, so that it can easily be disabled.

tee-supplicant/src/tee_socket.c
jforissier: OK fine, but what's wrong with `tee_ipsocket.h`?

jenswi-linaro: GP may suddenly decide to use that name for something.

jforissier: Ah, got it. Wouldn't it be the same for `tee_isocket_defines.h`?

jenswi-linaro:
Fixed compile errors.
It turns out that `CFG_SQL_FS=y` disables some warnings when compiling tee-supplicant, that's why I didn't see those errors before.
I'm currently investigating why I don't get the current error that travis gets.

tee-supplicant/src/tee_socket.c
jenswi-linaro: Here I'm just following our old pattern with the `_defines.h` files.
I guess we have the same problem here. I'll fix for these new files.

tee-supplicant/src/tee_socket.c
jenswi-linaro: It seems part of the code is missing, fixed now (and test cases updated).

jforissier:
@jenswi-linaro about `CFG_SQL_FS=y` disabling some warnings in `tee-supplicant`: that's a Bad Thing(TM), I'll look into it ;)

jenswi-linaro:
Renamed some internal header files

jenswi-linaro:
Added `CFG_SOCKET`, default `= y`

jenswi-linaro:
Note there's still the dependency to https://github.com/linaro-swg/linux/pull/30

tee-supplicant/src/tee_socket.c
```diff
@@ -0,0 +1,745 @@
...
+#include <sys/queue.h>
+#include <sys/socket.h>
+#include <tee_client_api.h>
+#include <teec_trace.h>
+#include <tee_socket.h>
+#include <tee_supplicant.h>
+#include <unistd.h>
+
+#include "handle.h"
+#include "tee_api_defines.h"
```
jforissier: Is this really needed?
For generic error codes, I think we should stick with `TEEC_*` rather than `TEE_*` because that's how it's done all over the place. As for socket-specific codes, you have them in `__tee_isocket_defines.h` etc. and that's fine by me.
Then hopefully we can avoid importing `tee_api_defines.h`.
Or did I miss something?...

tee-supplicant/src/tee_socket.c
```diff
@@ -0,0 +1,745 @@
...
+	if (val == -1)
+		return -1;
+
+	val |= flags;
+
+	return fcntl(fd, F_SETFL, val);
+}
+
+static TEEC_Result sock_connect(uint32_t ip_vers, unsigned int protocol,
+				const char *server, uint32_t port, int *ret_fd)
```
jforissier: A TCP/UDP port number is usually `uint16_t`. Change everywhere?

tee-supplicant/src/tee_api_defines.h
```diff
@@ -0,0 +1,474 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* Based on GP TEE Internal Core API Specification Version 1.1 */
+
```
etienne-lms: should we really store all these for the socket part to use TEE error codes ?
could define in the isocket part some TEE_ISOCKET_ERROR_GENERIC/_COMMUNICATION/...).

tee-supplicant/src/tee_socket.c
```diff
@@ -0,0 +1,597 @@
...
+	fd = sock_handle_to_fd(instance_id, handle);
+	if (fd < 0)
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	buf = tee_supp_param_to_va(params + 1);
+	bytes = params[1].u.memref.size;
+	res = write_with_timeout(fd, buf, &bytes, params[2].u.value.a);
+	if (res == TEEC_SUCCESS)
+		params[2].u.value.b = bytes;
+	return res;;
```
etienne-lms: s/;;/;/
also in `tee_socket_recv()` below.

tee-supplicant/src/tee_socket.c
```diff
@@ -0,0 +1,597 @@
...
+	fd = sock_handle_to_fd(instance_id, handle);
+	if (fd < 0)
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	buf = tee_supp_param_to_va(params + 1);
+
+	switch (command) {
+	case TEE_TCP_SET_RECVBUF:
+		return ioctl_set_tcp_recvbuf(fd, buf, params[1].u.memref.size);
+		break;
```
etienne-lms: remove  `break`

tee-supplicant/src/tee_socket.c
jenswi-linaro: OK, I'll fix

tee-supplicant/src/tee_api_defines.h
jenswi-linaro: Removing this file from tee-supplicant.

tee-supplicant/src/tee_socket.c
jenswi-linaro: OK

jenswi-linaro:
Comments addressed.

It's getting harder and harder to keep "Add support for UDP iSocket interface" and "Add support for TCP iSocket interface" apart so I'll squash them in the end now if no one minds.

jenswi-linaro:
Fix problem in test case 2002

jenswi-linaro:
Updated with mutex for SQLFS and RPMB

tee-supplicant/src/rpmb.c
```diff
@@ -25,19 +25,20 @@
...
-#include <linux/types.h>
-#include <linux/mmc/ioctl.h>
-#include <sys/types.h>
 #include <sys/stat.h>
-#include <fcntl.h>
-#include <stdbool.h>
-#include <stdlib.h>
+#include <sys/types.h>
+#include <tee_client_api.h>
+#include <teec_trace.h>
```
etienne-lms: `#include <tee_supplicant.h>`

jenswi-linaro:
`tee_supplicant.h` is now included.

jenswi-linaro:
Rebased and squashed.

jenswi-linaro:
Update

config.mk
```diff
@@ -38,6 +38,10 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 # such as 1008.5 that test loading of corrupt TAs.
 CFG_TA_TEST_PATH ?= 1
 
+# CFG_SOCKET
+# Enable Global Platform Socket support
+CFG_SOCKET ?= y
```
jforissier: `CFG_GP_SOCKETS` would be a better name IMO.

config.mk
```diff
@@ -38,6 +38,10 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 # such as 1008.5 that test loading of corrupt TAs.
 CFG_TA_TEST_PATH ?= 1
 
+# CFG_SOCKET
+# Enable Global Platform Socket support
```
jforissier: Sockets (plural preferred)

tee-supplicant/src/tee_supplicant.c
```diff
@@ -396,22 +435,137 @@ static bool find_params(union tee_rpc_invoke *request, uint32_t *func,
...
+	if (!read_request(arg->fd, &request))
+		return false;
+
+	if (!find_params(&request, &func, &num_params, &params, &num_meta))
+		return false;
+
+	if (num_meta && !num_waiters_dec(arg) && !spawn_thread(arg))
+		return false;
+
+
```
etienne-lms: minor: empty line

tee-supplicant/src/tee_supplicant.c
```diff
@@ -396,22 +435,137 @@ static bool find_params(union tee_rpc_invoke *request, uint32_t *func,
...
+		break;
+	case RPC_CMD_SHM_FREE:
+		ret = process_free(num_params, params);
+		break;
+	case RPC_CMD_GPROF:
+		ret = gprof_process(num_params, params);
+		break;
+	default:
+		EMSG("Cmd [0x%" PRIx32 "] not supported", func);
+		/* Not supported. */
```
etienne-lms: minor: could remove this comment.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -396,22 +435,137 @@ static bool find_params(union tee_rpc_invoke *request, uint32_t *func,
...
+		num_waiters_dec(arg);
+		return false;
+	}
+
+	e = pthread_detach(tid);
+	if (e)
+		EMSG("pthread_detach: %s", strerror(e));
+
+	return true;
+
```
etienne-lms: minor: empty line

jenswi-linaro:
Changed to `CFG_GP_SOCKETS` as requested, tags applied.

jenswi-linaro:
Threads are spawned as needed to make sure that at least one thread available to serve a new request in case the current requests are blocked for a long time.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -416,77 +618,14 @@ int main(int argc, char *argv[])
...
-	struct tee_ioctl_buf_data data;
-
-	data.buf_ptr = (uintptr_t)request;
-	data.buf_len = sizeof(*request);
-	if (ioctl(fd, TEE_IOC_SUPPL_RECV, &data)) {
-		EMSG("TEE_IOC_SUPPL_RECV: %s", strerror(errno));
-		return -1;
+	while (!arg.abort) {
+		if (!process_one_request(&arg))
+			arg.abort = true;
```
etienne-lms: As each call to `process_one_request()` spawns a thread for the next request, won't this spawn more threads than required?

I think we should spawn one thread then simply loop until a thread fails: 
```
	if (!process_one_request(&arg))
		arg.abort = true;

	while (!arg.abort)
		;
```

jenswi-linaro: With this code I'm trying to scale up the number of threads as needed without having any hard coded limits. More threads than needed could be spawned (depending on how the threads are scheduled), but not many more than needed and as long a at least one thread is left waiting in `read_request()` no more threads are spawned.

If we don't spawn in advance we'd need to check if we need to spawn a new thread each time we're doing an operation that may block for some time and that could be tricky as it's enough to miss one place to get hard to find occasional hangings.

jenswi-linaro: By the way, threads aren't expected to fail.

jforissier:

jenswi-linaro:
Rebased

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`
`Tested-by: Etienne Carriere <etienne.carriere@linaro.org> (b2260)`

jenswi-linaro:
Tags applied

PR #76, opened 2017-01-23
tee-supplicant: disable warnings only for sql_fs.c

jforissier:
sql_fs.c includes an API header (sqlfs.h in libsqlfs) which generates
warnings when built with our default flags. Silence them only for this
particular file, so that we won't miss potential issues in new code.


jenswi-linaro:

PR #77, opened 2017-01-31
ree fs rpc: open with best-effort allowing RO data fs

lws-team:
The REE handler for open always uses flags O_RDWR at the moment.

This breaks the possibility to use read-only operations on a
read-only mounted filesystem at /data.

Change it to try to open with O_RDWR, if not possible fall back
to O_RDONLY and only fail if that isn't workable either.

If O_RDONLY isn't enough, then the TA should use another means
to get the fs mounted rw for the duration of his write activity.
He will get failures on the write action if not, but that is
fair enough.

Signed-off-by: Andy Green <andy@warmcat.com>

### DIFF
--- a/tee-supplicant/src/tee_supp_fs.c
+++ b/tee-supplicant/src/tee_supp_fs.c
@@ -410,8 +410,15 @@ static TEEC_Result ree_fs_new_open(size_t num_params,
 		return TEEC_ERROR_BAD_PARAMETERS;
 
 	fd = open_wrapper(abs_filename, O_RDWR);
-	if (fd < 0)
-		return TEEC_ERROR_ITEM_NOT_FOUND;
+	if (fd < 0) {
+		/*
+		 * In case the problem is the filesystem is RO, retry with the
+		 * open flags restricted to RO.
+		 */
+		fd = open_wrapper(abs_filename, O_RDONLY);
+		if (fd < 0)
+			return TEEC_ERROR_ITEM_NOT_FOUND;
+	}
 
 	params[2].u.value.a = fd;
 	return TEEC_SUCCESS;


tee-supplicant/src/tee_supp_fs.c
```diff
@@ -410,8 +410,11 @@ static TEEC_Result ree_fs_new_open(size_t num_params,
 		return TEEC_ERROR_BAD_PARAMETERS;
 
 	fd = open_wrapper(abs_filename, O_RDWR);
-	if (fd < 0)
-		return TEEC_ERROR_ITEM_NOT_FOUND;
+	if (fd < 0) {
+		fd = open_wrapper(abs_filename, O_RDONLY);
```
jforissier: Please add a comment here, for instance "Try read-only to allow RO operations on a RO mounted filesystem"

lws-team:
I added the comment... no I haven't tried SQL storage, my customer just needs ree_fs.

For the larger issue of managing the flush + mount state though, everything has the same ultimate problem it's sitting on top of a REE caching / journalling filesystem that performs writes if mounted rw and offers no guarantees as it is about when it is actually committed.  Some kinds of storage have strict limits on the write / erase lifetime and can't afford to take the approach to default to rw mount.

I have a solution as I mentioned but it's basically an OPTEE_ api outside of the TEE_ spec to reference count if anyone wants write access to the mount... and by default mount it ro.

jenswi-linaro:
So `fsync()` is ignored?

lws-team:
In my case the /data fs is ext4 with a journal... I received certs from a remote server in the TA and used the Create API and closed the object.  After that it panic'd on the next step.

The next boot valid certs could never be found.

I can't recall if they were zero length stubs or just missing.

Implementing the "I need a writable mount" api solved it without changing anything else (since after the Create / close it then remounted the filesystem ro, definitively flushing it).

jenswi-linaro:
Yes, but if `fsync()` (or `O_DSYNC` flag to `open()`) works it's a lot more lightweight. And reliable too, what happens if crashing while the file system is temporarily mounted RW?

lws-team:
First there's nothing wrong with fsync() or other sticking-plasters.  However that did not work for me, and we require RO mount of data during normal operation for larger reasons.

Second... "reliable" ^^ Mounting critical certs writeable the whole time is not reliable.

For my customer, as I explained, he needs a writeable /data only in the factory, and for recovery.

If it crashed or lost power in the factory, well, my work on the TA must have been poor, or there are physical build problems on that device.  In the initramfs, if it cannot mount /data he reformats it and the device must re-provision itself, which it knows how to do in the factory and afterwards.

There are several different issues here getting mixed together:

 - OP-TEE should work if it finds /data is RO, and it only wants to perform RO operations, right?  It doesn't support that right now and this patch is necessary for that.

 - In the case the device operation is wholly with a RO /data unless the TA sees some situation like reprovisioning is necessary, the TA must be able to request /data becomes writeable.  This isn't part of the standard set of operations, and you can say it's out of scope if you like.  But many real TAs will require this whether it is in scope for OP-TEE or not.

 - I experienced commits on Hikey not making it to eMMC with the existing code under the conditions I described.  An explicit commit / flush by /data remount ro solved it.

jenswi-linaro:
If you ever need RW storage available you could perhaps have two FS spaces. One which is normally RO (except during production) with a special storage ID and another RW using the default storage ID.

jenswi-linaro:

lws-team:
Thanks... multiple kinds of storage mountpoints down /data would also be a heretical extension, although yes it would be useful is we had factory-time certs that should be RO, and also a need for other storage (we don't at the minute but others presumably will).

One thought... how about detect in tee-supplicant any operation with WRITE tee perms and do the reference counting from that and the object handle close?  Then the only footprint of it would be build-time config option on tee-supplicant.

jenswi-linaro:
Reference counting the R/W operations sound OK, not complicated but both REE and SQL FS has to be dealt with.

From a system point of view it sounds a bit dangerous to let tee-supplicant decide when it should remount RW though. If everything is automated it could in the end be a TA developer who doesn't have the complete view of the system who controls when /data is RW mounted.

jforissier:

lws-team:
Yeah.

There are per-TA dirs, each of which is potentially a discrete mountpoint (even if it is a bind mount).  But the naming binding to the TA is opaque (at least, it is to me).

However tee-supplicant knows the path being acted on with "write intent", actually that is enough to do a remount ro/rw if it is a mountpoint; it doesn't need the block device just the mountpoint for remount.

But how to tie all that together with a /etc/tee-supplicant/fstab or so is missing.  Filesystems / partitioning are usually a design decision at the system level.

lws-team:
Maybe every TA defaults to going in `/data/good-luck-bro/hash` similar to now.

But the critical guys (isn't there some capabilities thing for TAs?):

 - have non-opaque dirs like `/data/<TA_UUID>`

 - have filesystems already mounted ro there by the system initscript
 
 - do the reference counting and manage their own mount

jenswi-linaro:
Sound like a complicated workaround for the fact that normal world isn't using a crash safe filesystem.
This remount RW/RO thing isn't crash safe either, it only makes the window a bit smaller.

lws-team:
As I said, different things are getting mixed together.

Can you explain to me how mounting the storage read-only, so that it never takes a write after the factory, is less safe than mounting it rw every boot when it will never be written?  Or how "never" is making "the window a bit smaller"?

jenswi-linaro:
As long as it's always RO mounted it should be pretty safe, but then there's no need for the refcounting etc you're mentioning above.

To avoid confusing this PR further we should perhaps continue this discussion elsewhere.

lws-team:
I opened https://github.com/OP-TEE/optee_os/issues/1325 to discuss it a few days ago, this fix came out of that.

PR #81, opened 2017-03-01
Use the mode also in open_wrapper

jbech-linaro:
When source code is built with FORTIFY, it will give an error if you're
doing an open syscall without specifing the mode parameter.

Signed-off-by: Joakim Bech <joakim.bech@linaro.org>

### DIFF
--- a/tee-supplicant/src/tee_supp_fs.c
+++ b/tee-supplicant/src/tee_supp_fs.c
@@ -379,7 +379,7 @@ static int open_wrapper(const char *fname, int flags)
 	int fd;
 
 	while (true) {
-		fd = open(fname, flags);
+		fd = open(fname, flags, 0600);
 		if (fd >= 0 || errno != EINTR)
 			return fd;
 	}


tee-supplicant/src/tee_supp_fs.c
```diff
@@ -379,7 +379,7 @@ static int open_wrapper(const char *fname, int flags)
 	int fd;
 
 	while (true) {
-		fd = open(fname, flags);
+		fd = open(fname, flags, S_IRUSR | S_IWUSR);
```
jbech-linaro: Jerome, you prefix numbers instead? I think I saw that ... in some other PR a while ago?

jforissier: Indeed, they're not recommended in the kernel (https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/scripts/checkpatch.pl?id=refs/tags/v4.10#n6175) and I tend to agree. `0600` is faster to grasp than `S_IRUSR | S_IWUSR`, and if you don't know what it means then you should probably not be writing or reviewing code that deals with POSIX files ;)

jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

etienne-lms:
There are still calls to `open()` not providing a mode argument (libteec/src/tee_client_api.c tee-supplicant/src/rpmb.c tee-supplicant/src/tee_supplicant.c). Shouldn't they be fixed too ?

Also, shouldn't we rather use `(S_IRUSR | S_IWUSR)` instead of `0600` ?

jbech-linaro:
I think they should Etienne, since sooner or later we will have issues with those too. Let's try to not forget about it.

> Also, shouldn't we rather use (S_IRUSR | S_IWUSR) instead of 0600 ?

No, it seems like general consensus is that it's easier to understand 0600 than seeing a long line of define. I know .. it's just the other way round to everything else related to magic numbers. But I tend to agree in this case with 0XYZ is easier.

PR #82, opened 2017-03-09
tee-supplicant: correct block device path for android

zhizhouzh:
ueventd create block devices under /dev/block/.

Signed-off-by: Zhizhou Zhang <zhizhouzhang@asrmicro.com>

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -150,10 +150,14 @@ static int mmc_rpmb_fd(uint16_t dev_id)
 {
 	static int id;
 	static int fd = -1;
-	char path[21];
+	char path[PATH_MAX];
 
 	if (fd < 0) {
+#ifdef __ANDROID__
+		snprintf(path, sizeof(path), "/dev/block/mmcblk%urpmb", dev_id);
+#else
 		snprintf(path, sizeof(path), "/dev/mmcblk%urpmb", dev_id);
+#endif
 		fd = open(path, O_RDWR);
 		if (fd < 0) {
 			EMSG("Could not open %s (%s)", path, strerror(errno));
@@ -172,9 +176,13 @@ static int mmc_rpmb_fd(uint16_t dev_id)
 static int mmc_fd(uint16_t dev_id)
 {
 	int fd;
-	char path[17];
+	char path[PATH_MAX];
 
+#ifdef __ANDROID__
+	snprintf(path, sizeof(path), "/dev/block/mmcblk%u", dev_id);
+#else
 	snprintf(path, sizeof(path), "/dev/mmcblk%u", dev_id);
+#endif
 	fd = open(path, O_RDONLY);
 	if (fd < 0)
 		EMSG("Could not open %s (%s)", path, strerror(errno));


vchong:
Hi @zhizhouzh Thanks! Do you know which xtest numbers will fail without this patch?

zhizhouzh:
@vchong  That's xtest 6001. And you should turn off CFG_REE_FS first in optee-os side.

vchong:
@zhizhouzh Thanks! Do you mean you only get an error when using CFG_RPMB_FS = y?

zhizhouzh:
@vchong Yes.

tee-supplicant/src/rpmb.c
```diff
@@ -150,10 +150,14 @@ static int mmc_rpmb_fd(uint16_t dev_id)
 {
 	static int id;
 	static int fd = -1;
-	char path[21];
+	char path[32];
```
jforissier: You're only adding 6 characters so `char path [27];` should be enough. Or is there a specific reason why 32 would be better?

tee-supplicant/src/rpmb.c
```diff
@@ -172,9 +176,13 @@ static int mmc_rpmb_fd(uint16_t dev_id)
 static int mmc_fd(uint16_t dev_id)
 {
 	int fd;
-	char path[17];
+	char path[24];
```
jforissier: Same here.

tee-supplicant/src/rpmb.c
zhizhouzh: That's to make this array aligned to 8 bytes. Because the compiler will allocate the same stack size even if I set this array size to 27. So change it to 32 bytes may be more explicitly compared with assemble code.

The major difference is the check in snprintf, since `sizeof(path)` increased some bytes. But I didn't see any problem here.

zhizhouzh: That's to make this array aligned to 8 bytes. Because the compiler will allocate the same stack size even if I set this array size to 27. So change it to 32 bytes may be more explicitly compared with assemble code.

The major difference is the check in snprintf, since `sizeof(path)` increased some bytes. But I didn't see any problem here.

jenswi-linaro: I'd rather we use `PATH_MAX`, trying to use as short buffer as possible for file names is obviously causing problems now and then.

jforissier: I don't object to the use of `PATH_MAX` although it seems a bit silly here, since the max size if pretty obvious and easy to double check.

jforissier:
With the arrays updated to `PATH_MAX`:

zhizhouzh:
Updated, thanks.

jenswi-linaro:

jbech-linaro:
@zhizhouzh @jforissier is this ready to merge? Seems so to me. It will eventually need a rebase.

jforissier:
@jbech-linaro yep, I'll merge this right away.

jbech-linaro:
Thanks @jforissier and thanks @zhizhouzh for the contribution.

PR #83, opened 2017-03-22
tee-supplicant: Report OK only for actual loading

dryvenn:
I know that optee_os' strategy is to first pass all-zeroed parameters so that it gets back the size to allocate.
But let's say that for some reason the allocation is invalid, that it to say it was successful, but not good enough for TA copying. That can happen for instance if the allocated size is too small, or if the allocation was not done by the supplicant itself (this is what I did, I wrongfully tried to do it by calling `tee_shm_alloc` from kernel space and pass that as buffer).
In that case, during the second call, `try_load_secure_module` will return `TA_BINARY_FOUND`, letting the caller believe that the copy worked as expected.

I don't know if this is the right fix, and it probably has consequences on the linux driver and/or on optee_os, but this is a start!

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -279,6 +279,14 @@ static uint32_t load_ta(size_t num_params, struct tee_ioctl_param *params)
 	}
 
 	params[1].u.memref.size = size;
+
+	/*
+	 * If a buffer wasn't provided, just tell which size it should be.
+	 * If it was provided but isn't big enough, report an error.
+	 */
+	if (shm_ta.buffer && size > shm_ta.size)
+		return TEEC_ERROR_SHORT_BUFFER;
+
 	return TEEC_SUCCESS;
 }
 


tee-supplicant/src/teec_ta_load.c
```diff
@@ -148,7 +148,9 @@ static int try_load_secure_module(const char* prefix,
 		 * let the caller increase the size of the buffer and try
 		 * again.
 		 */
-		goto out;
+		*ta_size = s;
+		fclose(file);
+		return TA_INSUFFICIENT_SIZE;
```
etienne-lms: if would rather fix straight the tee-supplicant:
```
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -281,6 +281,8 @@ static uint32_t load_ta(size_t num_params, struct tee_ioctl_param *params)
        }
 
        params[1].u.memref.size = size;
+       if (size > shm_ta.size)
+               return TEEC_ERROR_SHORT_BUFFER;
        return TEEC_SUCCESS;
 }
```

jenswi-linaro: I prefer your approach @etienne-lms, but it is `params[1].u.memref.size` that `size` need to be checked against (before the assignment).

etienne-lms: @jenswi-linaro, no, after `TEECI_LoadSecureModule()` is called, `size` is updated with the TA binary size, while `shm_ta.size` remains to what the caller provided.

dryvenn: Ok, checking first the buffer isn't NULL will discard the need to change other projects I guess.

jenswi-linaro: You're right.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -281,6 +281,13 @@ static uint32_t load_ta(size_t num_params, struct tee_ioctl_param *params)
 	}
 
 	params[1].u.memref.size = size;
+
+	/* If a buffer wasn't provided, just tell which size it should be.
```
etienne-lms: minor style issue: multi-line comments are expected as:
`/*`<new-line>
` * ...`
` */`

jforissier:
Looks good to me.

etienne-lms:

jenswi-linaro:
Doesn't this need a change in `rpc_load()` in `core/arch/arm/kernel/user_ta.c` also?

jbech-linaro:
> Doesn't this need a change in rpc_load() in core/arch/arm/kernel/user_ta.c also?

I'm not sure, having looked at the code I think we're only returning `TEE_SUCCESS` when things are as expected and the `TEEC_ERROR_SHORT_BUFFER` is only returned when something in indeed is wrong. So it looks like the code can remain same in the function you refer to.

jenswi-linaro:
Ah, yes you're right.

jforissier:
Thanks @dryvenn!

PR #84, opened 2017-03-23
ree_fs_new_opendir(): ignore empty directory

jforissier:
The behavior of ree_fs_new_opendir() should be the same when a TA
directory does not exist, or when it is empty: return
TEEC_ERROR_ITEM_NOT_FOUND in both cases.

Currently, ree_fs_new_opendir() returns TEE_SUCCESS when the TA
directory exists.  This causes an issue with REE FS, when the data
directory is mounted over NFS. "xtest 6009 6010" fails with
TEE_StartPersistentObjectEnumerator() returning TEE_SUCCESS instead of
TEE_ERROR_ITEM_NOT_FOUND.

This is because test 6009 does not properly clean up. The test creates
persistent objects, closes them, opens them again with the meta accesss
flag and deletes them by calling TEE_CloseAndDeletePersistentObject1().
It turns out that tee-supplicant *does* attempt to remove the directory
but receives error ENOTEMPTY when the filesystem is NFS-mounted. The
reason is that when it removes the file, as a consequence of
syscall_storage_obj_del() calling o->pobj->fops->remove(o->pobj),
tee-supplicant still has an open file descriptor to the object. This
causes the NFS server to keep a hidden file (.nfsXXX) in the directory,
causing the subsequent rmdir() to fail. Then, when
syscall_storage_obj_del() calls tee_obj_close(), the fd is closed and
the .nfsXXX file disappears. We're left with an empty TA directory, and
when TEE_StartPersistentObjectEnumerator() invokes ree_fs_new_opendir()
it ends up with TEE_SUCCESS instead of TEE_ERROR_ITEM_NOT_FOUND.

One option to fix the issue is to fix the directory cleanup sequence,
i.e., make sure that the file descriptor is closed before the file is
unlinked and the directory removed. But we choose to adjust
ree_fs_new_opendir() instead, because it is less intrusive in OP-TEE OS
and also because the secure storage code is being reworked anyway for
other reasons.

### DIFF
--- a/tee-supplicant/src/tee_supp_fs.c
+++ b/tee-supplicant/src/tee_supp_fs.c
@@ -726,6 +726,8 @@ static TEEC_Result ree_fs_new_opendir(size_t num_params,
 	char *fname;
 	DIR *dir;
 	int handle;
+	struct dirent *dent;
+	bool empty = true;
 
 	if (num_params != 3 ||
 	    (params[0].attr & TEE_IOCTL_PARAM_ATTR_TYPE_MASK) !=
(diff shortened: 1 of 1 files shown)


jforissier:
FYI, the other option I mentioned in the commit description is implemented here: https://github.com/jforissier/optee_os/commit/nfs-fix but as I said it may not be the best fix.

Comments welcome.

jforissier:
(Updated to add a comment and not ignore `.nfs???` files, it's useless)

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -748,6 +750,31 @@ static TEEC_Result ree_fs_new_opendir(size_t num_params,
...
+	 * ree_fs_new_remove(). In this case the NFS server may create a
+	 * temporary reference called .nfs????, and the rmdir() call fails
+	 * so that the TA directory is left over. Checking this special case
+	 * here avoids that TEE_StartPersistentObjectEnumerator() returns
+	 * TEE_SUCCESS when it should return TEEC_ERROR_ITEM_NOT_FOUND.
+	 * Test case: "xtest 6009 6010".
+	 */
+	while ((dent = readdir(dir))) {
+		if (!strcmp(dent->d_name, ".") ||
+		    !strcmp(dent->d_name, ".."))
```
jenswi-linaro: We should perhaps ignore all `.*` files?

jforissier: Hmmmm... not really necessary... but it would be more consistent with the code in `tee_fs_readdir()` and save some typing ;)
I'll update.

jenswi-linaro:

PR #85, opened 2017-03-23
Android.mk: fix CFG_TEE_CLIENT_LOG_FILE checking

rpiasetskyi:
CFG_TEE_CLIENT_LOG_FILE variable contains the client library log file
path. The correct way to check if the variable is not empty is using
ifneq operator.

Signed-off-by: Ruslan Piasetskyi <ruslan.piasetskyi@gmail.com>


jenswi-linaro:

jforissier:
@rpiasetskyi thanks for the patch

PR #87, opened 2017-05-16
Fix incorrect share memory size

kong1191:
Accroding to TEE Client API Spec section 4.3.8 TEEC_RegisteredMemoryReference,
when handling a TEEC_RegisteredMemoryReference parameter of TEEC_MEMREF_WHOLE
type, the memory region size should read from its parent TEEC_SharedMemory
structure instead of the size described in TEEC_RegisteredMemoryReference
structure.

Signed-off-by: James Kung <kong1191@gmail.com>

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -340,7 +340,7 @@ static void teec_post_process_whole(TEEC_RegisteredMemoryReference *memref,
 		 * the shadow buffer into the real buffer now that we've
 		 * returned from secure world.
 		 */
-		if (shm->shadow_buffer && param->u.memref.size <= memref->size)
+		if (shm->shadow_buffer && param->u.memref.size <= shm->size)
 			memcpy(shm->buffer, shm->shadow_buffer,
 			       param->u.memref.size);
 


libteec/src/tee_client_api.c
```diff
@@ -340,7 +340,7 @@ static void teec_post_process_whole(TEEC_RegisteredMemoryReference *memref,
 		 * the shadow buffer into the real buffer now that we've
 		 * returned from secure world.
 		 */
-		if (shm->shadow_buffer && param->u.memref.size <= memref->size)
+		if (shm->shadow_buffer && param->u.memref.size <= shm->size)
```
etienne-lms: TEE side may have updated the `size` field if less byte that provided by th ebuffer were output by the secure sequence. So current `memref->size` seems better. 

Note that if this is to be changed, the same should be done in `teec_post_process_partial()`. But i think it's not required. Yet, maybe this deserves another review. @jforissier ?

jbech-linaro: James! Hi there old fellow! I hope you are doing good.

Regarding the change. I understand you both, but still, if we're not following the spec, shouldn't we update?

kong1191: Hi @etienne-lms , as far as I know, memref->size is assigned by CA, but memref->size is available if the memref type is **NOT** TEEC_MEMREF_WHOLE.  In other words, CA do not need to assign size in memref->size if memref type is TEEC_MEMREF_WHOLE. The actual output size updated by TA is kept in param->u.memref.size. If CA does not assign any value in memref->size, it will be the default value 0 and will avoid the shadow output buffer copy to the real share memory buffer.

This issue was found by our internal test suite, I can provide a simple test case to reproduce the issue if needed.

Hi @jbech-linaro , long time not to see you, I would like to say thank you to all OP-TEE team members and contributors. OP-TEE is really a great reference design TEE solution for us, thank you.

etienne-lms: Thanks @kong1191 for those details. I understand you are right:
* your change is fine (and fixes an issue)
* `teec_post_process_partial()` is out of scope since not a `MEMREF_WHOLE` (contrary to my earlier comment).

Indeed, the op-tee should include such a test where CA does not populate the `size`field for a `MEMREF_WHOLE`.

jforissier:
@kong1191 thanks for the contribution :)

Issue #88, opened 2017-05-16
tee_client_api.h provides incorrect max shared memory size


LKostyra:
In tee_client_api.h there is an [incorrect constant for max shared memory size](https://github.com/OP-TEE/optee_client/blob/master/public/tee_client_api.h#L46):
`#define TEEC_CONFIG_SHAREDMEM_MAX_SIZE 0x8000`

According to GP API specification (and the comment above the constant), the limit should be equal to 512 kB (0x8000**0**), now it is 32kB, which in result does not allow you to allocate higher memory sizes for bigger data.

jforissier:
Hi @LKostyra ,

You're right, that's obviously a typo and I'm surprise we did not found it sooner. We should probably add a test to our test suite.

Would you please create a pull request to fix this bug?

jforissier:
Closing this issue since pull request has been merged.

PR #89, opened 2017-05-16
tee_client_api: Fix typo in max shared memory size

LKostyra:
Resolves issue #88

### DIFF
--- a/public/tee_client_api.h
+++ b/public/tee_client_api.h
@@ -40,10 +40,11 @@
 
 /**
  * Defines the maximum size of a single shared memory block, in bytes, of both
- * API allocated and API registered memory. The size is currently set to
- * 512 * kB (512 * 1024).
+ * API allocated and API registered memory. There is no good value to put here
+ * (limits depend on specific config used), so this define does not provide any
+ * restriction in this implementation.
  */
-#define TEEC_CONFIG_SHAREDMEM_MAX_SIZE 0x8000
+#define TEEC_CONFIG_SHAREDMEM_MAX_SIZE ULONG_MAX
 
 /**
  * Flag constants indicating the type of parameters encoded inside the


jforissier:

Can you please also add your sign-off? Signed-off-by: tag, see https://github.com/OP-TEE/optee_os/blob/master/Notice.md#contributions and please provide a real name/email address as explained in the linux documentation: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?h=v4.12-rc1#n424

etienne-lms:
Actually this define is used nowhere. Isn't it a deprecated macro ?

LKostyra:
All done. Sorry about that, missed the info about contribution.

etienne-lms:
> Actually this define is used nowhere. Isn't it a deprecated macro ?

My mistake. This is specified by the GP TEE specs and there is a xtest over it.

LKostyra:
@etienne-lms I've decided to check how Allocate/RegisterSharedMemory functions define this in the specification (4.5.4 and 4.5.5 - description of APIs TEEC_RegisterSharedMemory and TEEC_AllocateSharedMemory in TEE Client API specification):

>An Implementation MAY put a hard limit on the size of a single Shared Memory block, defined by the constant TEEC_CONFIG_SHAREDMEM_MAX_SIZE.

Basically, there's nothing wrong with the implementation on Client's side, as this check is not obligatory. That might explain why some xtests pass while surpassing this limit.

As for the value, it is stated in macro description in the same document (4.4.1):

> In systems where there is no limit imposed by the Implementation then this definition should be defined to be the size of the address space

If 31 MB mentioned by you is the limit for TAs in core implementation, then I guess this is the correct value.

jforissier:
@etienne-lms you're probably right that 512 KB is not a good value, because it does not sound right that an allocation larger than `TEEC_CONFIG_SHAREDMEM_MAX_SIZE` can succeed.
Let's use 31 MB instead (where does this value come from? I think it would be nice to explain in a comment next to the #define)

etienne-lms:
In current release, non-LPAE (resp. LPAE) offers 31MB (resp. 4GB) of va space to each TA.

The TA va space size can be retrieved at runtime through `core_mmu_get_user_va_range()`. I let you dig in the code to find how it's computed, it is not that complex but not that straightforward :)

jforissier:
Thanks @etienne-lms. So, there's no unique good value, since it depends on the config, right?
So, should we do `#define TEEC_CONFIG_SHAREDMEM_MAX_SIZE ULONG_MAX`?

jenswi-linaro:
Sounds good, especially with regards to the shared memory work @lorc is doing.

LKostyra:
Corrected and modified the comment above the define.

jforissier:

jenswi-linaro:

jforissier:
Merged manually. Thanks!

PR #90, opened 2017-05-19
Storage: Added CFG_TEE_FS_PATH config option

rpiasetskyi:
tee-supplicant needs to create folder which will contain secure files.
In case folder can not be created, tee-supplicat exits.

Sometimes developers have to change this location to another.
For example, Android OS can not use /data folder, because tee-supplicant
is needed for decrypt userdata.img (which is mounted in /data folder).

* Add CFG_TEE_FS_PATH config option
* Cleanup secure storage init logic

Signed-off-by: Ruslan Piasetskyi <ruslan.piasetskyi@gmail.com>


tee-supplicant/Makefile
```diff
@@ -41,7 +41,9 @@ TEES_INCLUDES 	:= ${CURDIR}/../libteec/include \
 
 TEES_CFLAGS	:= $(addprefix -I, $(TEES_INCLUDES)) $(CFLAGS) \
 		   -DDEBUGLEVEL_$(CFG_TEE_SUPP_LOG_LEVEL) \
-		   -DBINARY_PREFIX=\"TEES\"
+		   -DBINARY_PREFIX=\"TEES\" \
+		   -DTEE_FS_PATH=\"$(CFG_TEE_FS_PATH)\" \
```
igoropaniuk: `\`in the end of the last line is not needed

tee-supplicant/tee_supplicant_android.mk
```diff
@@ -9,6 +9,7 @@ LOCAL_CFLAGS += $(optee_CFLAGS)
 LOCAL_CFLAGS += -DDEBUGLEVEL_$(CFG_TEE_SUPP_LOG_LEVEL) \
                 -DBINARY_PREFIX=\"TEES\" \
                 -DTEEC_LOAD_PATH=\"$(CFG_TEE_CLIENT_LOAD_PATH)\" \
+                -DTEE_FS_PATH=\"$(CFG_TEE_FS_PATH)\" \
```
igoropaniuk: same here

rpiasetskyi: I just add my define in the same code style as the previous line: using backslash in the end, so diff should be just one line, not two (remove previous line, add previous line with backslash and add new line).

tee-supplicant/Makefile
rpiasetskyi: I just add my define in this code style: using backslash in the end of line, so diff should be just one line, not two (remove previous line, add previous line with backslash and add new line).

jforissier: Then insert your new line before the last one. It is not a simple cosmetic issue: if the line next to this definition happens to be non-empty it will cause problems. I've been bitten in the past.

tee-supplicant/tee_supplicant_android.mk
rpiasetskyi: Done.

tee-supplicant/Makefile
rpiasetskyi: Done.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,47 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+	char *pp;
+	int res;
+	size_t n;
+
+	/*
+	 * Remove trailing spaces and unnecessary slashes.
+	 * Check that path exists.
+	 * After this subpath variable will contain absolute path like /a/b/c.
+	 */
+	pp = realpath(TEE_FS_PATH, subpath);
```
rpiasetskyi: Is it acceptable behaviour: tee-supplicat returns error if TEE_FS_PATH doesn't exist? Otherwise, tee-supplicant continue with creation of TEE_FS_PATH/tee/ folder.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,47 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+		return -1;
 
-	mkdir(TEE_FS_SUBPATH, 0700);
-	mkdir(TEE_FS_PATH, 0700);
-	if (stat(TEE_FS_PATH, &st) != 0)
+	/*
+	 * Check output size
+	 */
+	n = res;
+	if (n >= sizeof(tee_fs_root) || n != strlen(subpath) + strlen("/tee/"))
```
etienne-lms: why those tests ? could simply check `(res < 0) || (res >= sizeof(tee_fs_root))`.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,47 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
 	struct stat st;
+	char subpath[PATH_MAX];
+	char *pp;
+	int res;
+	size_t n;
+
+	/*
+	 * Remove trailing spaces and unnecessary slashes.
+	 * Check that path exists.
+	 * After this subpath variable will contain absolute path like /a/b/c.
```
etienne-lms: Could remove these comments and the one below. Code is explicit enough.

config.mk
```diff
@@ -32,6 +32,11 @@ CFG_TEE_SUPP_LOG_FILE ?= \"/data/teesupp.log\"
 # The location of the client library file.
 CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib
 
+# CFG_TEE_FS_PATH
+# Path to folder that will contain TEE filesystem.
+# This folder should be created before tee-supplicant start.
+CFG_TEE_FS_PATH ?= /data
```
etienne-lms: Rather call it `CFG_TEE_FS_PARENT_PATH`

tee-supplicant/src/tee_supp_fs.c
etienne-lms: Maybe keep the `mkdir(TEE_FS_SUBPATH, 0700);` for legacy support.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,47 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+	 * Check output size
+	 */
+	n = res;
+	if (n >= sizeof(tee_fs_root) || n != strlen(subpath) + strlen("/tee/"))
 		return -1;
+
+	/*
+	 * Create TEE filesystem root folder
+	 */
+	res = mkdir(tee_fs_root, S_IRWXU);
```
etienne-lms: Seems that `0700` is preferred to `S_IRWXU`. You should rather stick to it for coherency in the tee-supplicant sources.

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Done.

config.mk
rpiasetskyi: Done.

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Done. Change to variable mode.

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Done.
Check both cases in one comparison.

vchong:
Hi Ruslan,

So, just curious, rather than /data, what do you use? There's not really an equivalent of linux's /tmp on Android.

Also, is the encryption of userdata.img an implementation specific to your work only?

Thanks!

rpiasetskyi:
Hello Victor,

You can use any partition which is mounted for reading and writing.
Very bad equivalent can be /mnt folder which has tmpfs filesystem. Also, it can be /cache folder.

No, it isn't specific to my work only.
https://source.android.com/security/encryption/
https://source.android.com/security/keystore/

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,37 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
 int tee_supp_fs_init(void)
 {
 	struct stat st;
+	char subpath[PATH_MAX];
+	char *pp;
+	int res;
+	size_t n;
+	unsigned mode = 0700;
+
+	mkdir(TEE_FS_PARENT_PATH, mode);
```
rpiasetskyi: If it is necessary, I can add functionality to create all parent folders.

vchong:
Hi Ruslan, thanks for the quick reply! Aren't access to `/mnt` and `/cache` also restricted by selinux, or have you written rules for them (or whichever dir you're using) to allow access by tee-supplicant?

rpiasetskyi:
Yes, the developer has to write the SELinux rules for tee-supplicant for accessing to `/mnt`, `/cache` or `/data` folders.

vchong:
Ah, ok, thanks! I thought you were moving it from `/data` to another location where access rules were not required.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,37 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
 int tee_supp_fs_init(void)
 {
 	struct stat st;
+	char subpath[PATH_MAX];
+	char *pp;
+	int res;
+	size_t n;
+	unsigned mode = 0700;
```
jenswi-linaro: Type should be `mode_t`

tee-supplicant/src/tee_supp_fs.c
jenswi-linaro: Yes, please add code to create all parent folders

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,37 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+
+	mkdir(TEE_FS_PARENT_PATH, mode);
+	pp = realpath(TEE_FS_PARENT_PATH, subpath);
```
jenswi-linaro: Why is this needed?

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,37 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+	/*
+	 * Check permissions
+	 */
+	if (stat(tee_fs_root, &st) != 0)
+		return -1;
 
-	mkdir(TEE_FS_SUBPATH, 0700);
-	mkdir(TEE_FS_PATH, 0700);
-	if (stat(TEE_FS_PATH, &st) != 0)
+	if ((st.st_mode & mode) != mode)
```
jenswi-linaro: What's the point with this check?

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: That's why I added comments to this line first time :)

```
/*
 * Remove trailing spaces and unnecessary slashes.
 * Check that path exists.
 * After this subpath variable will contain absolute path like /a/b/c.
 */
```

`realpath()  expands  all  symbolic  links and resolves references to /./, /../ and extra '/' characters in the null-terminated string named by path to produce a canonicalized absolute pathname.`

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Checking that mode contains all necessary rights (`rwx`) for the user (like we asked).

For example, the previous `umask` call can remove right for execute.

tee-supplicant/src/tee_supp_fs.c
jenswi-linaro: Yes, I've read the man page too. Why is it useful to do this?

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,37 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+	unsigned mode = 0700;
+
+	mkdir(TEE_FS_PARENT_PATH, mode);
+	pp = realpath(TEE_FS_PARENT_PATH, subpath);
+	if (!pp)
+		return -1;
+
+	n = snprintf(tee_fs_root, sizeof(tee_fs_root), "%s/tee/", subpath);
+	if (n >= sizeof(tee_fs_root))
+		return -1;
```
etienne-lms: maybe check the appended path string is ok before eventually creating the parent path.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -302,11 +302,37 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+
+	/*
+	 * Create TEE filesystem root folder
+	 */
```
etienne-lms: style: useless comment

tee-supplicant/src/tee_supp_fs.c
jforissier: IMO it's (almost) useless because it does not ensure the rights will still be OK when the files are actually accessed (chmod can be done externally at any time for instance). In other words, checking permissions here do not save us from properly dealing with access errors later. I think it's the point @jenswi-linaro is trying to make.

tee-supplicant/src/tee_supp_fs.c
etienne-lms: Could it be that is can help the path to fit in the `subpath[]` string array ?

rpiasetskyi: Normalise path (remove unnecessary elements like `/` and `.`).
Also, check path existence.

If you do not like this function, it is no problem to remove it.

tee-supplicant/src/tee_supp_fs.c
jenswi-linaro: OK, but it doesn't deal with the problem. It would be better to unconditionally `chmod()` after each `mkdir()` because we want a usable directory regardless of umask.

etienne-lms: if **/data** exists with permission `0555,` `mkdir(/data, 0700)` will fail with `-EEXIST` code, yet the directory won't have the expected permissions. This code would trap it.
Note that if **/data** was `0500,` the directory won't neither have the expected attributes, yet the propose code will not catch this failure and it is likely that a later API or request will fail.

tee-supplicant/src/tee_supp_fs.c
jenswi-linaro: It seems rather pointless to me, I'd rather that `stat()` is used to check existence of the directory.

If tee-supplicant is configured with a path that contains redundancies, that's not much of a problem and regardless nothing the prevents supplicant from working. Who's configuring the system should know better than us what's the optimal path.

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Done.

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Removed this code.

tee-supplicant/src/tee_supp_fs.c
rpiasetskyi: Done.

rpiasetskyi: Removed.

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -299,14 +299,57 @@ static int tee_fs_access(struct tee_fs_rpc *fsrpc)
...
+		status = do_mkdir(path, mode);
+
+	free(subpath);
+	return status;
+}
+
+int tee_supp_fs_init(void)
+{
+	size_t n;
+	mode_t mode = S_IRWXU;
```
jenswi-linaro: The actual octal number is generally preferred.

rpiasetskyi: No problem, but `ree_fs_new_create` function, for example, contains `S_` defines.

rpiasetskyi: Done.

jforissier: https://github.com/OP-TEE/optee_client/pull/93

rpiasetskyi: Then my fault.

jforissier: @rpiasetskyi not your fault ;) I'm creating a PR because we need to fix our code to comply to our own recommendations!

jenswi-linaro:

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`. nice cleanup. thanks.

jforissier:
@etienne-lms review tag added, typo in commit message fixed, PR merged. Thanks @rpiasetskyi!

PR #92, opened 2017-05-22
supplicant: clean duplicated command IDs

etienne-lms:
Use `OPTEE_MSG_RPC_CMD_xxx` macros from **optee_msg_supplicant.h** rather
than redefining local macros. Add missing macros `OPTEE_MSG_RPC_CMD_SHM_ALLOC/FREE`.

### DIFF
--- a/tee-supplicant/src/optee_msg_supplicant.h
+++ b/tee-supplicant/src/optee_msg_supplicant.h
@@ -43,13 +43,6 @@
  */
 #define OPTEE_MSG_RPC_CMD_FS		2
 
-/* Was OPTEE_MSG_RPC_CMD_SQL_FS, which isn't supported any longer */
-#define OPTEE_MSG_RPC_CMD_SQL_FS_RESERVED	8
-
-/*
- * Values 3-7 are reserved in optee_msg.h for use by the kernel driver
- */
-
 /*
  * Define protocol for messages with .cmd == OPTEE_MSG_RPC_CMD_FS and first
  * parameter has the attribute OPTEE_MSG_ATTR_TYPE_VALUE_INPUT.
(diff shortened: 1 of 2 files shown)


tee-supplicant/src/optee_msg_supplicant.h
```diff
@@ -160,6 +156,21 @@
...
 
 /*
+ * Values 3, 4 and 5 are reserved in optee_msg.h for use by the kernel driver
+ */
+
+/*
+ * Shared memory allocation
+ */
+#define OPTEE_MSG_RPC_CMD_SHM_ALLOC	6
+#define OPTEE_MSG_RPC_CMD_SHM_FREE	7
```
jforissier: This should be moved up after `OPTEE_MSG_RPC_CMD_FS`. And `OPTEE_MSG_RPC_CMD_SQL_FS_RESERVED` should be moved here. So that the file is ordered on increasing (command, subcommand) values.

etienne-lms: line [44](https://github.com/OP-TEE/optee_client/pull/92/files#diff-2845320461671308ec53c0fa6d435eebR44) defines Ids 0x2 (and 0x8 since legacy rsv sqlfs), followed by the FS cmd Ids ABI.
Then at line [176](https://github.com/OP-TEE/optee_client/pull/92/files#diff-2845320461671308ec53c0fa6d435eebR176), socket cmd Ids is defined following by its ABI.
Hence i think its is the rigth location.

jenswi-linaro:

tee-supplicant/src/optee_msg_supplicant.h
jforissier:
```
$ grep 'define OPTEE_MSG_RPC_CMD_' optee_msg_supplicant.h 
#define OPTEE_MSG_RPC_CMD_LOAD_TA	0
#define OPTEE_MSG_RPC_CMD_RPMB		1
#define OPTEE_MSG_RPC_CMD_FS		2
#define OPTEE_MSG_RPC_CMD_SQL_FS_RESERVED	8
#define OPTEE_MSG_RPC_CMD_SHM_ALLOC	6
#define OPTEE_MSG_RPC_CMD_SHM_FREE	7
#define OPTEE_MSG_RPC_CMD_GPROF		9
#define OPTEE_MSG_RPC_CMD_SOCKET	10
```

I wouldn't mind having 8 after 7 and before 9, but it's not a big deal ;)

etienne-lms:
I will update with a 2 commit proposal as Jerome suggested (i hope i guessed it right).

jforissier:
Thanks @etienne-lms I wouldn't have done better :D

etienne-lms:
tags applied

PR #93, opened 2017-05-23
Replace file permission symbolic values by octal numbers

jforissier:
Octal numbers are prefered over symbolic values for file access modes.

### DIFF
--- a/tee-supplicant/src/tee_supp_fs.c
+++ b/tee-supplicant/src/tee_supp_fs.c
@@ -174,7 +174,7 @@ static TEEC_Result ree_fs_new_create(size_t num_params,
 	strncpy(abs_dir, abs_filename, sizeof(abs_dir));
 	abs_dir[sizeof(abs_dir) - 1] = '\0';
 	d = dirname(abs_dir);
-	if (!mkdir(d, S_IRUSR | S_IWUSR | S_IXUSR)) {
+	if (!mkdir(d, 0700)) {
 		fd = open_wrapper(abs_filename, flags);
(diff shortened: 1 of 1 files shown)


rpiasetskyi:
What about `tee_fs_conv_mkdflags` function?

jbech-linaro:
Agree with @rpiasetskyi , if you are making the change, then everything in `tee_fs.h` should also be changed. Other than that:

jforissier:
Well, for `tee_fs_conv_mkdflags()` I'm not sure what's best. It's clearly mapping flags with similar names so it's quite easy to read and see it's correct. I don't have a strong opinion.

jbech-linaro:
> Well, for tee_fs_conv_mkdflags() I'm not sure what's best. It's clearly mapping flags with similar names so it's quite easy to read and see it's correct. I don't have a strong opinion.

True, but still, we are trying to replace the symbolic values and then I think it is a step in the right direction.

jbech-linaro:

jenswi-linaro:
`tee_fs_conv_mkdflags()` and the rest which is only used by `tee_supp_fs_process_primitive()` can be removed now since secure world doesn't need them any longer.

jforissier:
@jenswi-linaro true, let's leave that for another PR.

jforissier:
Rebased. Since legacy code has been removed, we're left with just 3 changes. I have applied @jbech-linaro's R-b since that change was already in the previous version.

jenswi-linaro:

PR #95, opened 2017-05-31
Android.mk: include teec_benchmark.c to fix compiling error

liuyq:
otherwise there will be following compiling error when build for android:
external/optee_client/libteec/src/tee_client_api.c:556: error: undefined reference to 'bm_timestamp'
external/optee_client/libteec/src/tee_client_api.c:592: error: undefined reference to 'bm_timestamp'
clang.real: error: linker command failed with exit code 1 (use -v to see invocation)

Signed-off-by: Yongqin Liu <yongqin.liu@linaro.org>

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -21,11 +21,16 @@ ifneq ($(CFG_TEE_CLIENT_LOG_FILE),)
 LOCAL_CFLAGS += -DTEEC_LOG_FILE=$(CFG_TEE_CLIENT_LOG_FILE)
 endif
 
+ifeq ($(CFG_TEE_BENCHMARK),y)
+LOCAL_CFLAGS += -DCFG_TEE_BENCHMARK
+endif
+
 LOCAL_CFLAGS += -DDEBUGLEVEL_$(CFG_TEE_CLIENT_LOG_LEVEL)
 LOCAL_CFLAGS += -DBINARY_PREFIX=\"TEEC\"
 
 LOCAL_SRC_FILES := libteec/src/tee_client_api.c\
-                  libteec/src/teec_trace.c
+                   libteec/src/teec_benchmark.c \
+                   libteec/src/teec_trace.c
 
 LOCAL_C_INCLUDES := $(LOCAL_PATH)/public \
                     $(LOCAL_PATH)/libteec/include \


liuyq:
@vchong @igoropaniuk please help to check this change.

Android.mk
```diff
@@ -25,7 +25,8 @@ LOCAL_CFLAGS += -DDEBUGLEVEL_$(CFG_TEE_CLIENT_LOG_LEVEL)
 LOCAL_CFLAGS += -DBINARY_PREFIX=\"TEEC\"
```
vchong: Add below?

```
ifeq ($(CFG_TEE_BENCHMARK),y)
LOCAL_CFLAGS += -DCFG_TEE_BENCHMARK
endif
```

vchong:
Other than comment above,

`Acked-by: Victor Chong <victor.chong@linaro.org>`

Android.mk
liuyq: right, we should have this too. and should we have this defined to y in the config.mk by default?

liuyq: updated to have change on LOCAL_CFLAGS, but not have change on config.mk, since heard that hikey is not supported yet. so do not have it enabled by default to avoid temporary errors for android compiling.

vishalbhoj: If this is a conditional include then the call to this function should also be conditional here:
https://github.com/OP-TEE/optee_client/blob/master/libteec/src/tee_client_api.c#L556

jforissier:
No, the way `bm_timestamp()` is done does not look good. It is a function for private use by libteec, right? I mean, no client application is supposed to call it? If so:
1. The file `libteec/linux/teec_benchmark.h` should be `libteec/src/teec_benchmark.h` (having it under linux makes it look like it's related to the driver)
2. The above file should declare the function prototypes when `CFG_TEE_BENCHMARK=y`, otherwise define static inline functions that do nothing.
3. As a consequence of (2), there is no need for `#ifdef CFG_TEE_BENCHMARK` in `libteec/src/teec_benchmark.c`. Instead this file should be compiled conditionally (by Makefile/Android.mk).
4. `benchmark_check_mode()` should be static (hence not declared in `teec_benchmark.h`).

So, someone please apply the above fixes, or explain to me what's wrong with this approach.

Android.mk
vchong: @vishalbhoj Don't know why it's done this way but the condition is here: https://github.com/OP-TEE/optee_client/blob/master/libteec/src/teec_benchmark.c#L215

liuyq:
@jforissier regarding to you comments, since the related files are added by a commit from @igoropaniuk, so I will leave to him to explain on that.

liuyq:
BTW, if anyone has any new update for this problem, please go ahead to submit new request.

vchong:
@jforissier This PR is just to fix the android build error. `bm_timestamp()` issue is sort of out of scope for this patch and for @liuyq / lmg as well. Is it ok to merge this first to fix the build and have @igoropaniuk look at `bm_timestamp()` rework in a different patch?

jforissier:
https://github.com/OP-TEE/optee_client/pull/96

vchong:
Thanks! Closing.

PR #96, opened 2017-06-01
libteec: benchmark cleanup

jforissier:
Some cleanup which should also fix a build error on Android:

external/optee_client/libteec/src/tee_client_api.c:556: error: undefined reference to 'bm_timestamp'
external/optee_client/libteec/src/tee_client_api.c:592: error: undefined reference to 'bm_timestamp'
clang.real: error: linker command failed with exit code 1 (use -v to see invocation)

- Move include/linux/teec_benchmark.h up one level, because this file
is not related to the Linux TEE driver
- Make bm_timestamp() a no-op static inline if CFG_TEE_BENCHMARK != y
- Remove conditional compilation in teec_benchmark.c, instead, add the
file to the build only when CFG_TEE_BENCHMARK = y
- Declare benchmark_check_mode() static in teec_benchmark.c, since it
is not used anywhere else

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -29,24 +29,24 @@
 #include <fcntl.h>
 #include <limits.h>
 #include <pthread.h>
-#include <string.h>
-#include <stdio.h>
 #include <stdbool.h>
-#include <sys/types.h>
+#include <stdio.h>
+#include <string.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
-#include <tee_client_api.h>
+#include <sys/types.h>
 #include <tee_client_api_extensions.h>
+#include <tee_client_api.h>
+#include <teec_trace.h>
 #include <unistd.h>
 
 #ifndef __aligned
 #define __aligned(x) __attribute__((__aligned__(x)))
 #endif
 #include <linux/tee.h>
-#include <linux/teec_benchmark.h>
 
-#include <teec_trace.h>
+#include "teec_benchmark.h"
 
 /* How many device sequence numbers will be tried before giving up */
 #define TEEC_MAX_DEV_SEQ	10
--- a/libteec/src/teec_benchmark.c
+++ b/libteec/src/teec_benchmark.c
@@ -26,20 +26,21 @@
  */
 #include <err.h>
 #include <fcntl.h>
-#include <linux/teec_benchmark.h>
 #include <math.h>
 #include <pthread.h>
 #include <sched.h>
 #include <stdbool.h>
-#include <stdlib.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <tee_bench.h>
 #include <tee_client_api.h>
 #include <unistd.h>
 
-#ifdef CFG_TEE_BENCHMARK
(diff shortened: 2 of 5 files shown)


libteec/src/tee_client_api.c
```diff
@@ -29,24 +29,24 @@
...
 #include <stdbool.h>
-#include <sys/types.h>
+#include <stdio.h>
+#include <string.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
-#include <tee_client_api.h>
+#include <sys/types.h>
+#include "teec_benchmark.h"
```
etienne-lms: style would move `#include ""` fatter the `#include <>`

libteec/src/teec_benchmark.c
```diff
@@ -26,20 +26,20 @@
...
 #include <sched.h>
 #include <stdbool.h>
-#include <stdlib.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <tee_bench.h>
+#include "teec_benchmark.h"
```
etienne-lms: ditto

libteec/src/tee_client_api.c
jforissier: That is the order obtained from `sort` (vim: select block then `:!sort`). We discussed this before (https://github.com/OP-TEE/optee_os/pull/1421#discussion_r106377045).

One other option is to have two blocks, one with `<...>` and the other one with `"..."`, each one sorted by `sort`. In fact that was my first intent. @jenswi-linaro what do you prefer? We need a tiebreaker.

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with comments on includes addressed or not. Actually i don't really care splitting includes. It's more a personal habit I can change of if preferred.

Regarding the comment on CA needing to be able to store some timestamps, let's see that ilater.

libteec/src/tee_client_api.c
jbech-linaro: > One other option is to have two blocks

+1 ... if you need input from more people :)

jforissier:
> Drawback is that a CA cannot use bm_timestamp() as with previous implementation.

How could the CA use the function? `teec_benchmark.h` was not exported. Hence why I considered it a private file, and bm_timestamp() an internal function. Now it's quite possible it was not the intent, and the empty implementation in `teec_benchmark.c` indeed hints in that direction... But commit 63106fb4e086 ("OP-TEE Benchmark") does not mention the CA, it only refers to round-trip time between libteec and the TEE OS core.

@igoropaniuk let me know what you think. If `bm_timestamp()` is useful to client applications, then I'll drop this PR, to be replaced a patch to fix the Android build error as suggested in https://github.com/OP-TEE/optee_client/issues/95, and another one to export `teec_benchmark.h`.

igoropaniuk:
@jforissier theoretically, we can use `bm_timestamp()` in CAs and it will work, but, frankly speaking, I'm not sure about real cases where such timestamps would be needed.

jforissier:
@igoropaniuk OK, so, if you don't mind I will merge this PR with your Acked-by:. Thanks.

igoropaniuk:
@jforissier And thanks for this cleanup!
`Acked-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

PR #98, opened 2017-06-12
Makefile: preserve links to libraries during make copy_export

marckleinebudde:
During "make copy_export" and "make install" the shared library is
copied into the LIBDIR. However "cp" is used, thus the links to the
shared library are not preserved. This patch fixes the problem by using
"cp -a" instead.

Signed-off-by: Marc Kleine-Budde <mkl@pengutronix.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,6 @@ distclean: clean
 
 copy_export: build
 	mkdir -p $(DESTDIR)$(BINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
-	cp ${O}/libteec/libteec.so* $(DESTDIR)$(LIBDIR)
+	cp -a ${O}/libteec/libteec.so* $(DESTDIR)$(LIBDIR)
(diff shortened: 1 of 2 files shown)


jforissier:

jforissier:
On a related note: I think it would be better if the target of the symlinks would be relative rather then absolute paths. Absolute paths are ugly, and in there is always a risk that they may be packaged incorrectly. The `*.so*` files are always expected to reside in the same directory anyway.

To put it shortly, I think https://github.com/OP-TEE/optee_client/blob/master/libteec/Makefile#L42-L44 should be:

```make
libteec: $(TEEC_LIBRARY)
        $(VPREFIX)ln -sf $(LIB_MAJ_MIN) $(OUT_DIR)/$(LIB_MAJOR)
        $(VPREFIX)ln -sf $(LIB_MAJOR) $(OUT_DIR)/$(LIB_NAME)
```

Would you mind adding a patch to this PR to do just that? Thanks!

marckleinebudde:
Will do.

jforissier:

jforissier:
Acks applied and PR merged manually. @marckleinebudde thanks for contributing!

PR #102, opened 2017-08-28
tee-supplicant: add -DTEEC_LOAD_PATH in Makefile

airbak:
TEEC_LOAD_PATH default value is "/lib", it could be changed by
CFG_TEE_CLIENT_LOAD_PATH in tee_supplicant_android.mk.

As "/lib" could not meet all the scenes, it should be able to be
changed in Makefile for Linux.

Signed-off-by: Jianhui Li <airbak.li@hisilicon.com>


tee-supplicant/Makefile
```diff
@@ -42,7 +42,8 @@ TEES_INCLUDES 	:= ${CURDIR}/../libteec/include \
 TEES_CFLAGS	:= $(addprefix -I, $(TEES_INCLUDES)) $(CFLAGS) \
 		   -DDEBUGLEVEL_$(CFG_TEE_SUPP_LOG_LEVEL) \
 		   -DBINARY_PREFIX=\"TEES\" \
-		   -DTEE_FS_PARENT_PATH=\"$(CFG_TEE_FS_PARENT_PATH)\"
+		   -DTEE_FS_PARENT_PATH=\"$(CFG_TEE_FS_PARENT_PATH)\" \
+		   -DTEEC_LOAD_PATH=\"$(CFG_TEE_CLIENT_LOAD_PATH)\"
```
jforissier: This will change the default Linux build, because we have `CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib` in `config.mk`. So Linux would now use this instead of `/lib` which is the current default (see [tee-supplicant/src/teec_ta_load.c:39-41](https://github.com/OP-TEE/optee_client/blob/master/tee-supplicant/src/teec_ta_load.c#L39-L41)).

My suggestion is, in addition to the above change:
- Remove the `#ifndef TEEC_LOAD_PATH` block in `teec_ta_load.c`
- Set `CFG_TEE_CLIENT_LOAD_PATH ?= /lib` in `config.mk` (instead of `/system/lib`)
- In Android.mk, set `CFG_TEE_CLIENT_LOAD_PATH ?= /system/lib` before line 6 (where `config.mk` is included).

airbak:
Thanks for your suggestion,  I will pull another request.

PR #104, opened 2017-09-16
benchmark: use arm generic timers as time/tick source

igoropaniuk:
Switching to ARM Generic timers solves these problems:
1. Avoid any collisions with linux kernel perf subsystem when using PMU
as a tick source. Although perf subystem has a kernel
API(perf_event_create_kernel_counter etc.), unfortunately it doesn't
have any capabilites to enable EL0 access to cycle counter on
armv7/armv8 architectures (only for x86 platforms, where it's possible
to set bit which permits executing rdtsc from usermode)
2. Avoid any conditions, when frequency scaling occurs (moslty different
power states), or core is in wait-for-interrupt/wait-for-events states
(PMU doesn't count CPU cycles in these states).
3. Common tick source for all cores with fixed frequency

PL1 physical timer runs at frequency typically in the range 1-50MHz.

`Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

### DIFF
--- a/libteec/src/teec_benchmark.c
+++ b/libteec/src/teec_benchmark.c
@@ -54,11 +54,11 @@ static inline uint64_t read_ccounter(void)
 {
 	uint64_t ccounter = 0;
 #ifdef __aarch64__
-	asm volatile("mrs %0, PMCCNTR_EL0" : "=r"(ccounter));
+	asm volatile("mrs %0, cntvct_el0" : "=r" (ccounter));
 #else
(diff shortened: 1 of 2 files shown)


jbech-linaro:
`Acked-by: Joakim Bech <joakim.bech@linaro.org>`

igoropaniuk:
applied tag and rebased, thanks

Also fixed typo in the commit message

igoropaniuk:
closing PR as it's obsolete and should be reworked

Issue #64, opened 2016-11-23
Makefile hardcoded /lib breaks when distro uses /lib64


koenkooi:
As the subject says, if the distro uses /lib64 libtee*.so will get installed in the wrong place and won't be found be e.g. optee-test.

fboudra:
The use case is multilib enabled distributions. See also https://review.linaro.org/#/c/15766 for our workaround until this issue is fixed.

jbech-linaro:
We used to have different paths, but when we stepped up to GCC5.x, it seemed like it always default to `/lib/`, so we made a change roughly a year ago. Does anyone of you know if one can query gcc to figure out where it would like to see the libs etc to be placed? I.e, something similar to: `$ gcc --print-file-name=include` and friends?

jforissier:
@jbech-linaro the path where shared libraries should be installed ultimately depends on the distro, like @koenkooi said. The compiler has little to do with this (well, there is probably a default path embedded by the linker into the ELF files, I'm not sure but IMO it doesn't really matter).

The only viable option I see is to add a configuration variable to play a similar role as the autoconf "--libdir" option (`./configure --libdir=/lib64 ...`), let's call it `LIBDIR`. And while we're at it, let's do the same for `bin` and `include` and whack in `$(DESTDIR)` to help choose the staging location.

@koenkooi would that work for you?

```diff
diff --git a/Makefile b/Makefile
index 93bfa0e..3805dc6 100644
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,10 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 
+LIBDIR ?= $(DESTDIR)$(EXPORT_DIR)/lib
+INCLUDEDIR ?= $(DESTDIR)$(EXPORT_DIR)/lib
+BINDIR ?= $(DESTDIR)$(EXPORT_DIR)/lib
+
 CFG_SQL_FS ?= n
 CFG_TA_GPROF_SUPPORT ?= n
 
@@ -123,11 +127,11 @@ checkpatch-all-files: checkpatch-pre-req
 distclean: clean
 
 copy_export: build
-       mkdir -p ${EXPORT_DIR}/lib ${EXPORT_DIR}/include ${EXPORT_DIR}/bin
-       cp ${O}/libteec/libteec.so* ${EXPORT_DIR}/lib
[...]

PR #110, opened 2017-11-21
[RFC] use autotools to handle optee_client build

etienne-lms:
optee_client generates standard Linux userland binaries.
Going to a generic package build support should eases TEE client library and supplicant to be integrated in OS, mainly dealing with package dependencies and installation pathes.

This RFC proposes to use the GNU autotools to handle configuration, build and installation of optee_client.
New configure.ac script to define package content, configuration and some autotools settings.
Makefile are moved to Makefile.am format.

The pkgconfig part is not finalized/tested.

Dependency on the OP-TEE build process https://github.com/OP-TEE/build/pull/212.

Comments are welcome.


jenswi-linaro:
I suppose this works and sometimes it's even not too inconvenient, but it's also a bit like forcing Emacs on a Vi user.

I fail to see what generic with this though, to me it looks like just another complicated way of doing the same thing.

etienne-lms:
True, this is another way of defining the build.
It can be integrated in build env  (as our OP-TEE/build) where we manually set dependencies/installations/..., but it can also be integrated in build env that use generic deps/install/... management, and this is the point. Preventing each integrator from maintaining a build glue layer between its build env and the packages it integrates in.

jenswi-linaro:
In our build system it complicates stuff, so at least that is on the negative side. Which ones will benefit and how much by this? RPM, Deb and Yocto?

Given that we already maintain makefiles for Android. Will we reach far enough with autotools for all the other users here? Is it specifically autotools that's needed or is it just acting on some standard `make` targets and variables?

You mention glue layer, will that be removed completely for all users with this?

Sorry for being such a pain in the neck, but I want to make sure that we're doing the right thing here.

jbech-linaro:
I've tested this and it works as expected. I'm personally not a fan of autotools. I have no problems with **running** the stuff, i.e., `autoreconf`, `bootstrap`, `configure` etc..., my main concern is that it is rather complex syntax configuring it. Having that said, just like everything else, that is probably just a learning curve. 

If I set my own concerns and dislike for autotools a side for a moment, I understand that it is convenient to use this for other people.
- Yocto/OpenEmbedded people, can get our stuff built and included in their rootfs by more or less just point our git.
- Buildroot: Same there, I've played with [buildroot](https://github.com/jbech-linaro/buildroot/commits/qemu_optee) and without something like autotools/cmake it becomes hacky.
- Linux users: I guess everyone that has been using Linux knows about the `configure`, `make`, `make install` steps seen in many programs. I.e., they are familiar with this.
- I'm not sure, but I would not be surprised if this isn't also is something making the life easier for preparing packages for package mangers, `pacman`, `apt-get`, `dnf`.
- Even for us when preparing our rootfs for OP-TEE, it would be quite convenient to have all binaries, libs etc going to a common place (the `make install` step), that would make it trivial to create our rootfs and we could avoid the complicated stuff we have in our [common.mk](https://github.com/OP-TEE/build/blob/master/common.mk#L411-L472) file.

So, is autotools the right way? If it makes life easier for the general mass, then I'd be happy to put my own dislike for it a side and go for it. Having that said, I think we definitely should give `cmake` a thought also, since that seems to be very popular (and supported by the stuff mentioned above) and in my opinion, that is much simpler to work with compared to autotools. Here I'm talking about writing the `CMakeLists.txt` in cmake vs writing the `configure.ac`, `Makefile.am`, `optee_client.pc.in` in autotools. Compare for example my two year old [cmake/optee_client](https://github.com/jbech-linaro/optee_client/commit/2ab049e1dcaf64e3d6de278f865ca8b02e482106) test that does almost the same as this particular PR (but not all flags are supported in that commit).

This is my 2 cents so far ...

koenkooi:
Things like WARNINGCFLAGS should go into makefile.am instead of hiding in configure.ac

koenkooi:
Also, s/yocto/OpenEmbedded/g in all comments above.

jforissier:
I don't mind switching to autotools (I've used them successfully in past projects more complex than `optee_client`), although I think it is a bit overkill. Is there some feature that `autotools` provides, that we could not easily implement in our current Makefiles? `./configure --prefix=...`? Easy. `make install DESTDIR=...`? Easy, too. Generate a `pkgconfig` file? Should not be hard either. All of this, without the overhead associated with `autotools` (waiting for configure script to run, checking for things that are not even needed...).

So I agree with @jbech-linaro in that we should listen to our users and if there is a way it can make their life easier, then we go for it. Otherwise I don't really see the point.

petegriffin:
If you do move to autotools can we ensure that the Open Embedded recipes and also the OP-TEE Android integration is updated at the same time?

d3zd3z:
@jforissier The main advantage of using autotools/cmake over just making the existing makefile support the necessary features is that it makes things like openembedded and other packaging systems a lot easier. These generally have rules defined for autotools or cmake projects and supporting that package just requires declaring that. Supporting a custom makefile requires writing configure, build, install, etc definitions, even if they are blank, because the package system doesn't know what to use for defaults.

Speaking of cmake, Zephyr has started using cmake, and so far my experience with it is pretty good.

fboudra:
fwiw, it's a step in the right direction. Moving away from hand crafted Makefiles to automated build system will make our life easier for the distro integration and has been requested by users (as well as members) for a long time.

Wrt which one, it's a matter of who's doing the work :) and personal opinion. In general, autotools are well supported but not the best one these days. My personal opinion for that matter is that we're better served by using CMake or Meson, than autotools. Bonus point for CMake which is supported in the Android world.

@jforissier issues have been raised in the past because you hardcode compiler flags as well and don't respect the distro settings. Handling Werror was one of these issues or lib path: https://github.com/OP-TEE/optee_client/issues/64

jforissier:
@fboudra thanks for your feedback.

So, what next? Would someone create another RFC PR with CMake, so that we can compare both approaches?

jbech-linaro:
> So, what next? Would someone create another RFC PR with CMake, so that we can compare both approaches?

I can try to dig into it, I haven't go that much experience, but I have used cmake in a couple of small projects. Eventually I can just tweak the stuff I linked to above.

jforissier:
@jbech-linaro sounds good

etienne-lms:
Hi, coming back to this. Agree on need to test also Cmake.

I think I should setup a client application that relies on autotools, and check it nicely finds the client library resources, to get the whole autotools picture.

Aside of the pure autotools story, I had other feedback on this rfc, regarding the optee_client delivery. One told me that it was a shame to need to rebuild the library or the supplicant to get some debug support. If all the hard coded debug config should be run-time modifiable (we're in the non secure side:); log level. Also this rfc use DEBUG to force gcc directives (-O0 -g -DEBUG=1). I have understood that is it not the clean way and could confuse the toolchain directive.

Makefile.am
```diff
@@ -0,0 +1,17 @@
+AM_CFLAGS	= -Wall -Wbad-function-cast -Wcast-align \
```
etienne-lms: This seems not the right way to inject this build directive.
Actually, this part is useless.
Is there a standard way to define such from here for all sub builds (lib and supplicant), or should we copy these in each sub **Makefile.am** as currently proposed in this rfc ?

Makefile.am
```diff
@@ -0,0 +1,17 @@
...
+		  -Wfloat-equal -Wformat-nonliteral -Wformat-security \
+		  -Wformat=2 -Winit-self -Wmissing-declarations \
+		  -Wmissing-format-attribute -Wmissing-include-dirs \
+		  -Wmissing-noreturn -Wmissing-prototypes -Wnested-externs \
+		  -Wpointer-arith -Wshadow -Wstrict-prototypes \
+		  -Wswitch-default -Wunsafe-loop-optimizations \
+		  -Wwrite-strings -Werror
+
+if DEBUG
+AM_CFLAGS	+= -DDEBUG -O0 -g
```
etienne-lms: This is ugly. Should go through the debug directives of the relate autotools.

libteec/Makefile.am
```diff
@@ -0,0 +1,34 @@
...
+libteec_la_SOURCES	+= src/teec_benchmark.c
+endif
+
+libteec_la_CFLAGS	+= $(AM_CFLAGS)
+libteec_la_CFLAGS	+= $(PTHREAD_CFLAGS)
+
+# install libteec.so.x.y which is not installed by the autotools
+install-exec-hook:
+	test -e $(DESTDIR)/$(libdir)/libteec.so.$(LIBTEEC_MAJOR).$(LIBTEEC_MINOR).$(LIBTEEC_MICRO)
+	$(LN_S) -f libteec.so.$(LIBTEEC_MAJOR).$(LIBTEEC_MINOR).$(LIBTEEC_MICRO) \
```
etienne-lms: Actually this is useless. Contrary to want I (though I) experienced, no client expect **libteec.so.0.1** installed.

jforissier: If it is the case, then the SONAME is not set properly. From the current build env (not autotools):
```
$ objdump -p optee_test/out/xtest/xtest | grep NEEDED
  NEEDED               libteec.so.1.0
  NEEDED               libpthread.so.0
  NEEDED               libm.so.6
  NEEDED               libc.so.6
$ objdump -p optee_client/out/libteec/libteec.so | grep SONAME
  SONAME               libteec.so.1.0
```

jbech-linaro:
> > So, what next? Would someone create another RFC PR with CMake, so that we can compare both approaches?

> I can try to dig into it, I haven't got that much experience, but I have used cmake in a couple of small projects. Eventually I can just tweak the stuff I linked to above.

Just created a proposal using CMake ... doing the same, you find that here: https://github.com/OP-TEE/optee_client/pull/111

etienne-lms:
closing this P-R in favour to the Cmake way #111 .

PR #111, opened 2017-11-29
[RFC] Use CMake to build optee_client

jbech-linaro:
This is the "CMake" way of doing what we are already doing with plain make and what Etienne has proposed in #110 (and https://github.com/OP-TEE/build/pull/212).

(@fboudra you volunteered to review CMake, so please have a look and let me know everything I got wrong :smile: ... since this time I used more (advanced) features than I've used in the past when using CMake.)

There are two ways you can play with this, the first just locally to get a feel for CMake and the second is a full QEMU v7 setup as we usually do.
1. Just checkout this particular commit (or `wget https://github.com/OP-TEE/optee_client/pull/111.patch` and `git am <patch>`)
```bash
$ export CC="<your_project_path>toolchains/aarch32/bin/arm-linux-gnueabihf-gcc"
$ mkdir build
$ cd build
$ cmake ..
$ make
```
I didn't know about this, but the tool `ccmake` (and for GUI lovers `cmake-gui`), you can easily see and change all flags (and the meaning of the flags). This is pretty cool and quite handy. Try for for example "`ccmake .`" (after running the "`cmake ..`" line. Also keep track of "c" vs "cc" and the dots).

2. Alternatively run it as a full "QEMU v7" build ...
 2.1 Check out a normal QEMU v7 forest (repo, since I use the standalone build.git).
 2.2 In `optee_client.git` checkout this particular commit.
 2.3 In `build.git` checkout my branch `rfc_cmake`: https://github.com/jbech-linaro/build/tree/rfc_cmake (do you want me to submit a PR from this also at this stage?)
 2.4 Run `make run` etc as usual from `build`.

I've been using the `GNUInstallDirs` module in CMake meaning that when exporting the libs, headers and binaries it by default use the folder structure proposed by GNU Coding Standards. The way to not end up trying to install files in the root of your disk when running `make install` is to add a flag to CMake, which I've done [here](https://github.com/jbech-linaro/build/commit/04423e1b56c3da9a24faa1158a190d22c4020b1e#diff-54428ee6e5a953f611b2d75d1baaa0f6R325) (this is basically the equivalent to autotools `--prefix` when running configure). The nice thing with using the `GNUInstallDirs` is that it would be trivial to include or put thing in the correct location when building rootfs using buildroot and similar tools.

There are some tweaks here and there left to be done, but in general I think this serves as a good example. Please let me know what you think and also, compare this to keep using make as today or using autotools as proposed in the PR(s) by Etienne.

**CMake way of building optee_client ...**
```bash
$ find -name "CMake*" | xargs wc -l
  33 ./CMakeLists.txt
   5 ./CMakeToolchain.txt
  55 ./libteec/CMakeLists.txt
  61 ./tee-supplicant/CMakeLists.txt
 154 total
```

**Pure "make" way of building optee_client ...**
```bash
$ find -name "Makefile" -o -name "*.mk" | grep -v -i "android" | xargs wc -l
  130 ./Makefile
   29 ./flags.mk
   61 ./libteec/Makefile
   70 ./config.mk
   85 ./tee-supplicant/Makefile
  375 total
```


.gitignore
```diff
@@ -7,3 +7,6 @@ Debug
 .project
 /libs
 /obj
+GPATH
+GRTAGS
+GTAGS
```
igoropaniuk:
```
cmake_install.cmake
CMakeCache.txt
Makefile
CMakeFiles
```

tee-supplicant/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+add_definitions (
+	-DCFG_TEE_SUPP_LOG_LEVEL=${CFG_TEE_SUPP_LOG_LEVEL}
+	-DCFG_TEE_SUPP_LOG_FILE="${CFG_TEE_SUPP_LOG_FILE}"
+	-DTEEC_LOAD_PATH="${CFG_TEE_CLIENT_LOAD_PATH}"
+	-DTEE_FS_PARENT_PATH="${CFG_TEE_FS_PARENT_PATH}"
+	-DBINARY_PREFIX="TSUP"
+)
+
+# Files always in use in all builds
+set (SRC
```
igoropaniuk:
```
file(GLOB SRC *.c)
```

CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+	-Werror-implicit-function-declaration -Wextra \
+	-Wfloat-equal -Wformat-nonliteral -Wformat-security \
+	-Wformat=2 -Winit-self -Wmissing-declarations \
+	-Wmissing-format-attribute -Wmissing-include-dirs \
+	-Wmissing-noreturn -Wmissing-prototypes -Wnested-externs \
+	-Wpointer-arith -Wshadow -Wstrict-prototypes \
+	-Wswitch-default -Wunsafe-loop-optimizations \
+	-Wwrite-strings -Werror -fPIC"
+)
+
```
igoropaniuk: Subset of debug flags can be introduced:
```
set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS} -Wall -ggdb -O0")
```

they are chosen automatically, when `CMAKE_BUILD_TYPE` is set to "Debug"

libteec/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+endif()
+
+add_library (teec SHARED ${SRC})
+
+set_target_properties (teec PROPERTIES
+	VERSION ${APP_VER}
+	SOVERSION ${APP_VER_MAJOR}
+)
+
+target_link_libraries (teec pthread)
```
igoropaniuk:
```
find_package(Threads REQUIRED)
if(NOT THREADS_FOUND)
	message(FATAL_ERROR "Threads not found")
endif()
```
and then instead of `pthread` put `${CMAKE_THREAD_LIBS_INIT}`

tee-supplicant/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+	src/handle.c
+	src/hmac_sha2.c
+	src/rpmb.c
+	src/sha2.c
+	src/tee_supp_fs.c
+	src/tee_supplicant.c
+	src/teec_ta_load.c
+)
+
+if (CFG_GP_SOCKETS)
```
igoropaniuk: I'd rather check explicitly for "ON" value here
```
if(${CFG_GP_SOCKETS} MATCHES ON)
```

CMakeToolchain.txt
```diff
@@ -0,0 +1,5 @@
+include (CMakeForceCompiler)
```
igoropaniuk: `CMakeForceCompiler` is deprecated

tee-supplicant/CMakeLists.txt
igoropaniuk: sorry, it won't work.
Just found this condition here
```
if (CFG_TA_GPROF_SUPPORT)
	add_definitions (-DCFG_TA_GPROF_SUPPORT)
	set (SRC ${SRC} src/gprof.c)
endif()
```

libteec/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+project(libteec)
+
+# Global needed variables
+set (APP_NAME ${PROJECT_NAME})
```
igoropaniuk: why introduce additional APP_NAME (we already have $PROJECT_NAME)

libteec/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+set_target_properties (teec PROPERTIES
+	VERSION ${APP_VER}
+	SOVERSION ${APP_VER_MAJOR}
+)
+
+target_link_libraries (teec pthread)
+
+# FIXME: This should in someway harmonize with CFG_TEE_CLIENT_LOAD_PATH
+# FIXME: Should we change this to /usr/local/lib?
+install (TARGETS teec DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
igoropaniuk: Minor: Explicitly specify type of the target (in this case `LIBRARY`)

tee-supplicant/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+if (CFG_TA_GPROF_SUPPORT)
+	add_definitions (-DCFG_TA_GPROF_SUPPORT)
+	set (SRC ${SRC} src/gprof.c)
+endif()
+
+add_executable (tee-supplicant ${SRC})
+
+target_link_libraries (tee-supplicant teec)
+
+install (TARGETS tee-supplicant DESTINATION ${CMAKE_INSTALL_BINDIR})
```
igoropaniuk: Maybe I'm wrong, but I guess all `install` commands are a bit useless without CPack (so we can create any type of package, deb/rpm/tar.gz by running `make install`), until we sre cross compiling optee_client (it is not supposed to be installed on the same host, where it was compiled)

jforissier: `make install DESTDIR=some/path` is quite useful to install into a staging directory, such as when one wants to create a rootfs. Does CMake support this usage too?

CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# https://cmake.org/Wiki/CMake_Useful_Variables
+set (CMAKE_TOOLCHAIN_FILE CMakeToolchain.txt)
+project (optee_client C)
+
+include(GNUInstallDirs)
+
+# Required cmake version
+# FIMXE: We can probably bump this up to > 3.4
+cmake_minimum_required (VERSION 2.8.12)
```
igoropaniuk: From https://cmake.org/cmake/help/v3.8/command/cmake_minimum_required.html

> Call the cmake_minimum_required() command at the beginning of the top-level CMakeLists.txt file even before calling the project() command. It is important to establish version and policy settings before invoking other commands whose behavior they may affect.

tee-supplicant/CMakeLists.txt
igoropaniuk: @jforissier sure, it's supported. The other option: defining `CMAKE_INSTALL_PREFIX` with needed  path in CMakeFiles.txt

tee-supplicant/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+if (CFG_TA_TEST_PATH)
+	add_definitions (-DCFG_TA_TEST_PATH=${CFG_TA_TEST_PATH})
+endif()
+
+if (CFG_TA_GPROF_SUPPORT)
+	add_definitions (-DCFG_TA_GPROF_SUPPORT)
+	set (SRC ${SRC} src/gprof.c)
+endif()
+
+add_executable (tee-supplicant ${SRC})
```
igoropaniuk: minor:  `${PROJECT_NAME}` instead of `tee-supplicant`

tee-supplicant/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+option (CFG_TA_GPROF_SUPPORT "Enabled tee-supplicant profiling support" OFF)
+
+set (CFG_TEE_SUPP_LOG_LEVEL "1" CACHE STRING "tee-supplicant log level")
+set (CFG_TEE_SUPP_LOG_FILE "/data/teesupp.log" CACHE STRING "Location of tee-supplicant log")
+set (CFG_TEE_CLIENT_LOAD_PATH "/lib" CACHE STRING "Location of libteec.so")
+set (CFG_TEE_FS_PARENT_PATH "/data" CACHE STRING "Location of TEE filesystem (secure storage)")
+# FIXME: Why do we have if defined(CFG_GP_SOCKETS) && CFG_GP_SOCKETS == 1 in the c-file?
+set (CFG_GP_SOCKETS "1" CACHE STRING "Enable GlobalPlatform Socket API support")
+
+include_directories (
```
jbech-linaro: After checking this talk: https://youtu.be/eC9-iRN2b04?t=30m34s ... this is probably not a good way of dealing with dependencies. If I understand that guy correctly, you should define all that stuff in the library itself (using PUBLIC/PRIVATE) and when done right you should only need to tell cmake that you need that library. I.e., you should only need to say to that "tee-supplicant needs libteec" and cmake will figure out include and flags automagically! Sweet!

CMakeLists.txt
jbech-linaro: After doing some more background checking I believe this should be done as an option to `add_compile_target(...)` which I've just introduced (you will see that on my review comments that I'll push later). I'll save this one and see if and how that should be done.

tee-supplicant/CMakeLists.txt
jbech-linaro: Since I'm a bit of a noob here, is that what you prefer personally or is that how people tend to do it normally in CMake? I'd aim for the greater mass and if that is `MATCHES ON` I'd be more than happy to change.

libteec/CMakeLists.txt
jbech-linaro: Probably because I'm a noob on CMake :smile:

libteec/CMakeLists.txt
jbech-linaro: What is the syntax your are looking for? I tried:
```cmake
install (LIBRARY teec DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
But that made CMake a bit upset.

tee-supplicant/CMakeLists.txt
jbech-linaro: Right, that is what I mean in the header text of this PR, i.e., basically what I'm doing [here](https://github.com/jbech-linaro/build/commit/04423e1b56c3da9a24faa1158a190d22c4020b1e#diff-54428ee6e5a953f611b2d75d1baaa0f6R325).

jbech-linaro:
I've pushed a few more review commits, which address most of Igor's comments (not yet the debug flag and not yet the MATCHES ON). I also re-wrote some parts which decouples dependencies etc. The I added some comment separators. I'm not sure if that is good or bad, but personally I think it makes it easier to find "sections" in the various CMakeLists.txt files.

tee-supplicant/CMakeLists.txt
igoropaniuk: @jbech-linaro Frankly, I do that way just for the sake of readability and emphasizing that this variable is supposed to have ON/OFF values.

FYI, in CMAKE too much stuff can mean True. The same for False :)
> True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number. False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND

jforissier:

`README.md` needs updating ("Build" section).

igoropaniuk:
`Reviewed-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

jenswi-linaro:
Can we have these additional changes too?

Take tee-supplicant install directory as a parameter
```
diff --git a/tee-supplicant/CMakeLists.txt b/tee-supplicant/CMakeLists.txt
index ef1118be4020..39fba65febf4 100644
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -81,4 +81,4 @@ target_link_libraries (${PROJECT_NAME}
 # Install targets
-install (TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR})
+install (TARGETS ${PROJECT_NAME} DESTINATION ${TEE_SUPPLICANT_INSTALL_BINDIR})
```

Avoid some build error with buildroot:
```
diff --git a/libteec/CMakeLists.txt b/libteec/CMakeLists.txt
index 7aec1156f0e0..4698494a1076 100644
--- a/libteec/CMakeLists.txt
+++ b/libteec/CMakeLists.txt
@@ -1,4 +1,4 @@
-project(libteec)
+project(libteec C)
 
 set(PROJECT_VERSION "1.0.0")
 
diff --git a/public/CMakeLists.txt b/public/CMakeLists.txt
index 7d719b421381..8e6071cb7762 100644
--- a/public/CMakeLists.txt
+++ b/public/CMakeLists.txt
@@ -1,4 +1,4 @@
-project (optee-client-headers)
+project (optee-client-headers C)
 
 # Built library
diff --git a/tee-supplicant/CMakeLists.txt b/tee-supplicant/CMakeLists.txt
index 43e59f81f16a..ef1118be4020 100644
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -1,4 +1,4 @@
-project (tee-supplicant)
+project (tee-supplicant C)
 
 # Configuration flags always included

```

etienne-lms:
With this change we definitely move from Makefile to Cmake for optee_client.
Do we expect some transition with both operating during a transition time frame? (maybe simply rename current `Makefile`'s into some `Makefile.gnu` or `Makefile.raw`).

jforissier:
@etienne-lms yes.
Why rename the Makefiles during the transition period? They pose no problem: either they are ignored (if you run `cmake` from a build directory) or they are overwritten (if you run `cmake` from the source directory).

etienne-lms:
As ignored, **Makefile** script files are no more tracked by git. No possible evolution from current state.

jforissier:
@etienne-lms ah, you're talking about `.gitignore`, right? Indeed I think it should be like that until the legacy Makefiles are removed:
```
Makefile
!/Makefile
!/tee-supplicant/Makefile
!/libteec/Makefile
```

etienne-lms:
if one build from the source tree (for some reason), running `cmake` will corrupt the **Makefile**s. This can be dealt manually, but it is a bit of a shame, no?

jforissier:
@etienne-lms not a big deal IMO, and it will not last forever.

jbech-linaro:
@jforissier @etienne-lms , I don't think it is a problem having them live side by side during a transition period. The recommended way to build using CMake is to build from a separate folder (what Jerome also said) and that is what I'm doing [here](https://github.com/jbech-linaro/build/blob/rfc_cmake/common.mk#L323-L327). I.e, all built files will be kept there and it won't interfere with existing files.

jbech-linaro:
> @jenswi-linaro wrote:
> Can we have these additional changes too?
> 
> Take tee-supplicant install directory as a parameter
```make
-install (TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR})
+install (TARGETS ${PROJECT_NAME} DESTINATION ${TEE_SUPPLICANT_INSTALL_BINDIR})
```

If I make that change right away, then `TEE_SUPPLICANT_INSTALL_BINDIR` would not be defined by default? That would also mean that we move away from using the default CMake target directories, i.e., not using the `GNUInstallDirs`. If the reason is that you want to have it in `/sbin/` instead of `/bin/`, then we should use `CMAKE_INSTALL_SBINDIR` instead. Or do you need the parameter for another reason?

> @jenswi-linaro wrote:
> Avoid some build error with buildroot:

Sure!

> @jforissier wrote:
> README.md needs updating ("Build" section).

Sure, I'll fix.

jenswi-linaro:
`CMAKE_INSTALL_SBINDIR` sounds good.

jbech-linaro:
I've pushed a commit addressing the README.md and Jens comments. If everyone's happy with it, I'll squash and rebase.

jenswi-linaro:

jbech-linaro:
Squashed, rebased and tag(s) applied (should be) ready for merge! Thanks!

PR #121, opened 2018-05-22
Pkcs#11 services through OP-TEE SKS TA

etienne-lms:
This changes introduces an implementation of the Pkcs#11 Cryptoki library that interface a specific  trusted application in the secure side.

This P-R implements the basics for the client library. The effective implementation of the services will come later. 

This P-R does not introduce the TA API itself that is being reviewed through https://github.com/OP-TEE/optee_os/pull/2342. The client lib will integrate the TA API once its is approved and merged.


libsks/include/pkcs11.h
```diff
@@ -0,0 +1,265 @@
+/* Copyright (c) OASIS Open 2016. All Rights Reserved./
```
jenswi-linaro: SPDX header please

etienne-lms: Looks like the OASIS IPR Policy is not registered to SPDX.
Tricky...

etienne-lms: I think i will go another way.
The [opencryptoki](https://github.com/opencryptoki/) project includes a package (opencryptoki) that proposes a pkcs11 cryptoki header under the creative commons license, [here](https://github.com/opencryptoki/opencryptoki/blob/master/usr/include/pkcs11/pkcs11types.h#L4).

The header file reflects a slightly older pkcs11 and must be updated with more recent IDs and few crypto parameter structures. From license view, it is more suitable for a project like the OP-TEE.  @jenswi-linaro, @jforissier, what do you think?

jforissier: [It is the "Common Public License, version 1.0 (CPL-1.0)", SPDX identifier "CPL-1.0". Nothing to do with the Creative Commons licenses AFAICT.]

I think we may have a problem with sections 2.a and 3.a.iv of https://spdx.org/licenses/CPL-1.0.html (source code).

etienne-lms: yes, CPL-1.0, thanks for fixing. From wikipedia, CPL-1.0 is not a good choice.
> CPL lacks compatibility with both versions of the GPL because it has a "choice of law" section in section 7, which restricts legal disputes to a certain court. Another source of incompatibility is the differing copyleft requirements. [[2]](https://en.wikipedia.org/wiki/Common_Public_License#cite_note-FSFLicenseList-2)

Rewrite from scratch based on the specs?

libsks/include/pkcs11.h
```diff
@@ -0,0 +1,277 @@
...
+#ifndef NULL_PTR
+#define NULL_PTR 0
+#endif
+
+/* All the various Cryptoki types and #define'd values are in the
+ * file pkcs11t.h.
+ */
+#include "pkcs11t.h"
+
+#define __PASTE(x,y)      x##y
```
jbech-linaro: Rename to `__JOIN` (or `__CONCATENATE`) might better describe what this actually does?

etienne-lms:
> If we cannot include it as it is, then I think that rewrite is the only option we have. I just wonder, wouldn't you end up with almost the same kind of defines etc in the end.

Yes, same IDs, labels, structures and prototypes...
I have started, it's a bit painful... will restrict to what is currently supported/planned for SKS.

etienne-lms:
Pushed proposal for a new **pkcs11.h** header file implementation.

etienne-lms:
Comments addressed and squashed.

etienne-lms:
> Not sure several commits are justified instead of one single "Add stubbed liboptee_cryptoki"?

True. I have split initial work into smaller change too easy review. Maybe too much...
I can squash in one if you feel it is better and safer.

Makefile
```diff
@@ -129,3 +137,6 @@ copy_export: build
 	cp -a ${O}/libteec/libteec.a $(DESTDIR)$(LIBDIR)
 	cp ${O}/tee-supplicant/tee-supplicant $(DESTDIR)$(BINDIR)
 	cp public/*.h $(DESTDIR)$(INCLUDEDIR)
+	cp libsks/include/*.h $(DESTDIR)$(INCLUDEDIR)
+	cp -a ${O}/libsks/liboptee_cryptoki.so* $(DESTDIR)$(LIBDIR)
```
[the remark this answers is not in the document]
etienne-lms: true. better be cleaned now. yet it find name `libsks` a bit short for a quite specific library (optee/pkcs11).

libsks/src/local_utils.h
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#ifndef __LOCAL_UTILS_H
+#define __LOCAL_UTILS_H
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#define ASSERT(cond) \
```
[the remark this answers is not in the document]
etienne-lms: true. thanks.

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1000 @@
+/*
+ * Copyright (c) 2017, Linaro Limited
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
```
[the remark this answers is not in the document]
etienne-lms: not mandated but ok.

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1000 @@
...
+	do { \
+		if (!inited) \
+			return CKR_CRYPTOKI_NOT_INITIALIZED; \
+	} while (0)
+
+#define SANITY_NONNULL_PTR(ptr) \
+	do { \
+		if (!ptr) \
+			return CKR_ARGUMENTS_BAD; \
+	} while (0)
```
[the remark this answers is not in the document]
etienne-lms: ok, i will go another way. thanks to point out.

etienne-lms:
all comments not addressed yet. in progress...

etienne-lms:
Comments addressed, but the one about squashing all the commits into a single one.

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1127 @@
...
+	DO_NOT_REGISTER_CK_FUNCTION(C_SeedRandom),
+	DO_NOT_REGISTER_CK_FUNCTION(C_GenerateRandom),
+	DO_NOT_REGISTER_CK_FUNCTION(C_GetFunctionStatus),
+	DO_NOT_REGISTER_CK_FUNCTION(C_CancelFunction),
+	DO_NOT_REGISTER_CK_FUNCTION(C_WaitForSlotEvent),
+};
+
+static int inited;
+
+static int sanity_lib_init(void)
```
jforissier: Remove. This function sounds like it initializes the library, but actually it does not (`C_Initialize()` does).

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1127 @@
...
+{
+	return inited == 0;
+}
+
+static int sanity_nonnull_ptr(void *ptr)
```
jforissier: Remove.

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1127 @@
...
+
+	inited = 1;
+	return CKR_OK;
+}
+
+CK_RV C_Finalize(CK_VOID_PTR res)
+{
+	(void)res;
+
+	if (sanity_lib_init())
```
jforissier: `if (!inited)`

What about thread safety? Is the API supposed to be thread-safe?

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1127 @@
...
+		return CKR_CRYPTOKI_NOT_INITIALIZED;
+
+	return CKR_FUNCTION_NOT_SUPPORTED;
+}
+
+CK_RV C_GetFunctionList(CK_FUNCTION_LIST_PTR_PTR ppFunctionList)
+{
+	/* Note: no need to call sanity_lib_init() here */
+	if (sanity_nonnull_ptr(ppFunctionList))
```
jforissier: `if (!ppFunctionList)`

libsks/src/pkcs11_api.c
etienne-lms: Yes but I would rather postpone this issue to later...

libsks/src/pkcs11_api.c
etienne-lms: ok

libsks/src/pkcs11_api.c
etienne-lms: ok

libsks/src/pkcs11_api.c
etienne-lms: ok

libsks/src/pkcs11_api.c
jforissier: Sure.

libsks/src/pkcs11_api.c
```diff
@@ -0,0 +1,1117 @@
...
+
+CK_RV C_Initialize(CK_VOID_PTR init_args)
+{
+	(void)init_args;
+	CK_C_INITIALIZE_ARGS_PTR args = (CK_C_INITIALIZE_ARGS_PTR)init_args;
+
+	/* Argument currently unused */
+	(void)args;
+
+	if (lib_inited)
```
jbech-linaro: inited? What about ...initiated? Some comment for the other places further down.

jbech-linaro:
Besides a minor suggestion, I think this looks good. Then we have the golden question, whether to merge something in the middle of being developed or not. It would be nice to live up to "submit early and often" and get pieces of the complete implementation merged, one by one ... @OP-TEE/linaro thoughts?

jforissier:
@jbech-linaro generally speaking, my opinion is we should "submit early and often" but into a separate branch, until the code actually provides a real service or at least can be somewhat tested. If I understand correctly, it is the case for this PKCS#11 stuff already?
Anyway, I don't feel strongly about it, and I will not object to merging code into master as long as it has been sufficiently reviewed.

etienne-lms:
@jforissier, yes this P-R comes with related P-R in **[optee_os](https://github.com/OP-TEE/optee_os/pull/2342)** and **[optee_test](https://github.com/OP-TEE/optee_test/pull/281)** and do provide some test materials. All these P-R rely on configuration directive `CFG_SECURE_KEY_SERVICES` to build related code in each component. The currently proposed build sequence (OP-TEE/build#261) is independent of this P-R.
Anyway, i strongly agree with "I will not object to merging code into master as long as it has been sufficiently reviewed.".

jforissier:
@etienne-lms all good then, thanks. I will take all this as soon as 3.2.0 is out, which may be in a couple of days or so (I'd like to better understand OP-TEE/optee_os#2414 and OP-TEE/optee_os#2437).

etienne-lms:
@jforissier, fine with me.

etienne-lms:
Superseded by https://github.com/OP-TEE/optee_client/pull/138.

PR #124, opened 2018-05-29
libteec: Fix incorrect SOVERSION number

jbech-linaro:
Seems to be working as expected after applying @scarabeusiv fix.
```bash
...
$ cmake -DCMAKE_INSTALL_PREFIX=/tmp/client ..
...
$ make install                                                                 
[ 25%] Built target teec                                                       
[100%] Built target tee-supplicant                                             
Install the project...                                                         
-- Install configuration: ""                                                   
-- Installing: /tmp/client/lib/libteec.so.1.0.0                                
-- Installing: /tmp/client/lib/libteec.so.1                                    
-- Installing: /tmp/client/lib/libteec.so
....
```


jbech-linaro:
Commit message: s/Reviewed--by/Reviewed-by/

jforissier:
The author's `Signed-off-by:` is missing.

jbech-linaro:
> The author's Signed-off-by: is missing.

Isn't my Signed-off by sufficient? I.e., I'm the one taking the blame in case there would be a legal issue? I'm not really sure how that works, but I remember that Greg K-H asked Vee to add his SoB tag for patches he sent on behalf of someone else.

jbech-linaro:
Alright, then let me change the patch to be me as author and I add Tomáš as "Suggested-by" and "Reported-by" instead.

jforissier:
> Isn't my Signed-off by sufficient?

I don't think so. At least not if we keep the original author's name as the Git Author. It's probably OK if you claim to be the author however. In this case the DCO applies, you certify that the origin of the patch is OK for the project and we're done (you're the one to blame as you say although for this PR there is no big risk for sure :=)).
Let me try to clarify based on the Linux practice and what we have been doing and see if it makes sense.

In the Linux kernel, there is one Sob for each person carrying the patch (carrying means forwarding it in one way or another, typically by committing to a maintainer tree), starting from the author(s) and then the subsystem maintainers up the trees that finally reach Linus' tree. Reference: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?h=v4.17-rc7#n516

Therefore in any Linux Git tree there is normally a Sob for both the Author and the Committer. See  http://lkml.iu.edu/hypermail/linux/kernel/1709.0/00964.html.

Ever since we've been using the DCO in OP-TEE we have always followed this rule that the author must sign off. Admittedly, for the committer we haven't. It is for practical reasons, namely it is cumbersome to add the committer Sob instead of using the GitHub interface "rebase and merge". Since we don't have a hierarchy of Git repos, unlike Linux, I think it is acceptable to assume that the committer's Sob is implicit. There can be no one else in the path of the patch. And after all the committer adds nothing to the patch. When it happened that we had to do some small fixups we did add a Sob, as well as a description of the changes, as recommended in the Linux documentation. For instance https://github.com/OP-TEE/optee_os/commit/1043920434c6 (change by me) or https://github.com/OP-TEE/optee_os/commit/c6672fdcd95b (change by Jens).

jforissier:
> Alright, then let me change the patch to be me as author and I add Tomáš as "Suggested-by" and "Reported-by" instead.

Yep that would be fine I think.

jbech-linaro:
I've pushed it once again with updated git message.

jforissier:
Thanks

jbech-linaro:
Now we didn't get any reviewed-by tag, but it's not the end of the world for tiny patch like this.

PR #127, opened 2018-06-27
benchmark: fix compilation issues

igoropaniuk:
This fixes "cast from pointer to integer of different" warning, when
CFG_TEE_BENCHMARK is enabled.

```bash
src/teec_benchmark.c: In function 'mmap_paddr':
src/teec_benchmark.c:128:11: error: cast from pointer to integer of different
size [-Werror=pointer-to-int-cast]
  offset = (off_t)hw_addr % getpagesize();
           ^
src/teec_benchmark.c:129:14: error: cast from pointer to integer of different
size [-Werror=pointer-to-int-cast]
  page_addr = (off_t)(hw_addr - offset);
```

`Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>`


jforissier:
LGTM

igoropaniuk:
Tag applied, thanks!

jforissier:
Oops, I don't think it was the proper fix after all :(
@igoropaniuk please see https://github.com/OP-TEE/optee_client/pull/128 "benchmark: fix computation of page address".

PR #128, opened 2018-07-05
Fixes for GCC 8.1

jforissier:


libteec/src/teec_trace.c
```diff
@@ -69,57 +68,33 @@ static const char * const trace_level_strings[] = {
...
-			memset(func_buf + flen, 0x20,
-			       (MAX_FUNC_PRINT_SIZE - flen));
-		}
-		func_buf[MAX_FUNC_PRINT_SIZE - 1] = '\0';
-		func = func_buf;
-#else
-		(void)flen;
-		func = function;
-#endif
+		int thread_id = syscall(SYS_gettid);
```
jenswi-linaro: This is new, we didn't use to have this before. Do we really need it?

jforissier: `thread_id` was there. I didn't add it.
I don't think it is useful, I can remove it.

jenswi-linaro: Aha I see that now, sorry. Since it's already there let's keep it to not change behavior in the patch.

jforissier:
Regarding commit "Plain make and Android: do not set -Werror by default": anyone knows how to do that with CMake/buildroot?

jforissier:
Ping?

jbech-linaro:

jforissier:
Updated. I have added @jbech-linaro's R-b where applicable, squashed a couple of commits, and inserted "benchmark: fix computation of page address" in first position.

jforissier:
@igoropaniuk about "benchmark: fix computation of page address": does it look OK to you?

libteec/src/teec_benchmark.c
```diff
@@ -125,8 +125,8 @@ static void *mmap_paddr(intptr_t paddr, uint64_t size)
 	if (!devmem)
 		return NULL;
 
-	offset = (off_t)(uintptr_t)hw_addr % getpagesize();
-	page_addr = (off_t)(uintptr_t)(hw_addr - offset);
+	offset = (off_t)paddr % getpagesize();
```
jenswi-linaro: Now that `hw_addr` isn't used here, we can remove the initialization of `hw_addr` above. While you're at it please remove the initialization of `offset` too.

PR #131, opened 2018-08-17
Makefile and config.mk: remove export in config.mk

liuyq:
to workaorund export restriction when building for android platform

Signed-off-by: Yongqin Liu <yongqin.liu@linaro.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,12 +30,12 @@ all: build install
 build-libteec:
 	@echo "Building libteec.so"
 	@$(MAKE) --directory=libteec --no-print-directory --no-builtin-variables \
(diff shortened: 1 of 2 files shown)


jforissier:
Why is `export` a problem? You're not saying in the commit description.

liuyq:
@jforissier, 
sorry, this change is workaround for the restriction in AOSP master described here:
https://android.googlesource.com/platform/build/+/master/Changes.md#export_keyword

with export used in make files, the aosp build will report error like here:
https://ci.linaro.org/job/android-lcr-reference-hikey-master/95/console 

will update the change later after tested with android platform.

jforissier:
Thanks @liuyq.

`Tested-by: Jerome Forissier <jerome.forissier@linaro.org> (QEMU)`

PR #132, opened 2018-08-17
tee-supplicant: add daemon mode (-d)

jforissier:
Add the missing code to make tee-supplicant a proper daemon when invoked
with -d.

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -455,10 +455,12 @@ static int get_dev_fd(uint32_t *gen_caps)
 	return -1;
 }
 
-static int usage(void)
+static int usage(int status)
 {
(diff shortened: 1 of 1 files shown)


tee-supplicant/src/tee_supplicant.c
```diff
@@ -670,6 +686,19 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	if (daemonize) {
+		pid = fork();
```
jenswi-linaro: Consider using `daemon(3)`  or you should do something about std{in,out,err} too.

jforissier: Updated to use `daemon(3)`, thanks

jenswi-linaro:

PR #133, opened 2018-09-07
tee-supplicant: cmake: enable RPMB emulation by default

jenswi-linaro:
Enables RPMB emulation by default also when compiling with cmake.


jenswi-linaro:
Tag applied

PR #141, opened 2019-01-22
flags: add -D_FILE_OFFSET_BITS=64 to support large files

### LINKED ISSUE #140: Client is build without LFS support
I ran into a Large File Support issue today, where the OP-TEE client could not detect the creation of the /data directory correctly, since the stat64() call was used.
The fix inside my BSP was to append `-D_FILE_OFFSET_BITS` to the optee_client CFLAGS.
The [GLIBC Feature page](https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_13.html) describes all the macros, you might prefer another fix for the optee_client Makefile.
Please let me now and I will create a Pull request with your prefered fix.

Emantor:
The tee client library does not need to provide compatibility for the old file
system interface.

Fixes #140

Signed-off-by: Rouven Czerwinski <r.czerwinski@pengutronix.de>

### DIFF
--- a/flags.mk
+++ b/flags.mk
@@ -14,7 +14,7 @@ CFLAGS          := -Wall -Wbad-function-cast -Wcast-align \
 		   -Wmissing-noreturn -Wmissing-prototypes -Wnested-externs \
 		   -Wpointer-arith -Wshadow -Wstrict-prototypes \
 		   -Wswitch-default -Wunsafe-loop-optimizations \
-		   -Wwrite-strings
+		   -Wwrite-strings -D_FILE_OFFSET_BITS=64
 ifeq ($(CFG_WERROR),y)
 CFLAGS		+= -Werror
 endif


flags.mk
```diff
@@ -14,7 +14,7 @@ CFLAGS          := -Wall -Wbad-function-cast -Wcast-align \
 		   -Wmissing-noreturn -Wmissing-prototypes -Wnested-externs \
 		   -Wpointer-arith -Wshadow -Wstrict-prototypes \
 		   -Wswitch-default -Wunsafe-loop-optimizations \
-		   -Wwrite-strings
+		   -Wwrite-strings -D_FILE_OFFSET_BITS=64
```
jforissier: Should this be added to `android_flags.mk` too?

vchong: Adding the flag doesn't seem to break the current hikey aosp build, but that's my concern too (about breaking older stuffs) after some googling. I'd prefer leaving it off for now as well. @liuyq what do you think?

vchong: btw thanks @jforissier for noticing `android_flags.mk` too : )

jforissier: OK, sounds reasonable. @Emantor would you please add a comment about Android to the commit message?

jforissier:

flags.mk
Emantor: sure.

Emantor:
Applied Tag and added explanation for android.

jforissier:
Thanks @Emantor. I have merged this manually (edited the Fixes: tag to be a GitHub URL rather than just the issue number -- I prefer it this way).

PR #142, opened 2019-02-05
tee-supplicant: fix cmake to install in bin/ not sbin/

etienne-lms:
GNU makefile and Android makefile both install tee-supplicant
in the bin/ directory. Prio this change CMakeLists.txt installed
tee-supplicant in the sbin/ directory. This change fixes the
CMake script to intall tee-supplicant in direcory bin/.

### DIFF
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -85,4 +85,4 @@ target_link_libraries (${PROJECT_NAME}
 ################################################################################
 # Install targets
 ################################################################################
-install (TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_SBINDIR})
+install (TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR})


etienne-lms:
This change requires build to be updated, see https://github.com/OP-TEE/build/pull/343.

jenswi-linaro:
Isn't sbin more appropriate?

jbech-linaro:
Not sure about the truth of the source, but it is a nice wrap-up:
http://blog.taylormcgann.com/2014/04/11/difference-bin-sbin/

jenswi-linaro:
Except that it doesn't mention `/usr/sbin`. ;-) I guess rule of thumb is: if normal users can use it, put it in bin else put it in sbin. I don't see any point in having tee-supplicant in the `$PATH` of normal users. On the other hand, I don't care that much and if it's important for someone else who will continue to make this work fine with me.

etienne-lms:
As discussed in https://github.com/OP-TEE/build/pull/343#issuecomment-460687813, this is more a consistency issue between the several build/install means for the tee-supplicant.

The alternate way is to change [Makefile copy_export](https://github.com/OP-TEE/optee_client/blob/3.4.0/Makefile#L130). As for the Android way, I could not find how to specify it has to be installed in a **.../bin/** or **.../sbin** directory. @vchong, maybe you can tell if Android is touchy on this.

The installed path may not impact users but the init script uses hard coded paths and it has to match the right one.

jforissier:
+1 for `sbin` because `bin` is for *commands* while `sbin` is for `system binaries` (it doesn't matter if `/` or `/usr`). See https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard.

This patch looks backwards. IMO it is not CMake that needs changing, it is the GNU Makefile which historically picked the wrong path (and Android too I think although I don't know the Android specifics).

vchong:
We don't explicitly set `LOCAL_MODULE_PATH` so the supplicant is installed by default in `/vendor/bin` as of Treble. No reason to change the default in this case.

etienne-lms:
I guess that's fine for the Android which is a specific distrib. Just note that if Android moves optee_client build to use CMake, the supplicant install path will change.

PR #143, opened 2019-02-06
tee-supplicant: fix install path in GNU makefile

etienne-lms:
Change GNU makefile to default install **tee-supplicant** in target directory **/bin** instead of **/sbin**.

This change aligns the GNU makefile install sequence with the CMake install sequence.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,9 +12,9 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 DESTDIR ?= $(EXPORT_DIR)
-BINDIR ?= /bin
 LIBDIR ?= /lib
 INCLUDEDIR ?= /include
+SBINDIR ?= /usr/sbin
 
 CFG_TA_GPROF_SUPPORT ?= n
 
@@ -124,8 +124,8 @@ checkpatch-all-files: checkpatch-pre-req
 distclean: clean
 
 copy_export: build
-	mkdir -p $(DESTDIR)$(BINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
+	mkdir -p $(DESTDIR)$(SBINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
 	cp -a ${O}/libteec/libteec.so* $(DESTDIR)$(LIBDIR)
 	cp -a ${O}/libteec/libteec.a $(DESTDIR)$(LIBDIR)
-	cp ${O}/tee-supplicant/tee-supplicant $(DESTDIR)$(BINDIR)
+	cp ${O}/tee-supplicant/tee-supplicant $(DESTDIR)$(SBINDIR)
 	cp public/*.h $(DESTDIR)$(INCLUDEDIR)
--- a/Makefile
+++ b/Makefile
@@ -12,9 +12,9 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 DESTDIR ?= $(EXPORT_DIR)
-LIBDIR ?= /lib
-INCLUDEDIR ?= /include
 SBINDIR ?= /usr/sbin
+LIBDIR ?= /usr/lib
+INCLUDEDIR ?= /usr/include
 
 CFG_TA_GPROF_SUPPORT ?= n
 


etienne-lms:
sorry, will fix...

etienne-lms:
Change replaced with an updated commit description.

Makefile
```diff
@@ -12,7 +12,7 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 DESTDIR ?= $(EXPORT_DIR)
-BINDIR ?= /bin
+BINDIR ?= /sbin
```
jforissier: If I'm not mistaken, CMake installs in `/usr/sbin`.

etienne-lms: I though it was the distrib that should set `DESTDIR` to either target path */** or **/usr**, depending on the expected install. My main concern was about the **bin/** or **sbin/** subdirectory.
Yet, ok to move all 3 `BINDIR`, `LIBDIR` and `INCLUDEDIR` to **/usr**.

jforissier: No, here DESTDIR has the same semantics as defined in the GNU documetnation https://www.gnu.org/prep/standards/html_node/DESTDIR.html, meaning it is a "staging" directory where to store files before packaging for instance. Indeed, it has no influence on the build (embedded paths in scripts etc.) and is used only by "make install".

What you are thinking about sounds more like "prefix" as in `configure --prefix=/usr` see https://www.gnu.org/prep/standards/html_node/Directory-Variables.html.

jenswi-linaro:

jforissier:

jforissier:
But, who uses this Makefile anyway? Our build.git uses buildroot for all platforms, doesn't it? So perhaps we'd better remove the Makefile altogether? (my only concern is about maintaining something we cannot test on a regular basis).

jbech-linaro:
> But, who uses this Makefile anyway? Our build.git uses buildroot for all platforms, doesn't it? So perhaps we'd better remove the Makefile altogether? (my only concern is about maintaining something we cannot test on a regular basis).

I still use it and I've had this up for discussion a couple of times (I even wrote a long Google doc about it, which I believe you all have seen). I think if we make it possible to do ``make optee-client`` from build.git so that it rebuilds only optee-client (using CMake), then wouldn't mind to get rid of the it. But as long as we cannot build targets individually, I think it's better to keep them.

etienne-lms:
The linaro OpenEmbedded recipes do use these makefile rather than the CMake: https://git.linaro.org/openembedded/meta-linaro.git/tree/meta-optee/recipes-security/optee/optee-client.bb?h=master

jbech-linaro:
> The linaro OpenEmbedded recipes do use these makefile rather than the CMake:

That's a bit odd, since I believe both Buildroot and OE suggest to use CMake or Autotools first and plain Makefiles is the last resort thing. Having that said, I could have misunderstood that.

Makefile
etienne-lms: You're right. I see that even using CMake I am badly using `DESTDIR` and `CMAKE_INSTALL_PREFIX`.

etienne-lms:
> > The linaro OpenEmbedded recipes do use these makefile rather than the CMake:
> 
> That's a bit odd, since I believe both Buildroot and OE suggest to use CMake or Autotools first and plain Makefiles is the last resort thing. Having that said, I could have misunderstood that.

True. But we still need thoses GNU makefile for the secure parts. Well, not for this optee_client, but at least for optee_os, optee_test and optee_examples.

jbech-linaro:
> But we still need thoses GNU makefile for the secure parts

Ah, right!

jforissier:
@jbech-linaro 
> I still use it

That's fine :)

> I think if we make it possible to do `make optee-client` from build.git so that it rebuilds only optee-client (using CMake), then wouldn't mind to get rid of the it. But as long as we cannot build targets individually, I think it's better to keep them.

Fair enough.

And, I realize that .travis.yml does build optee_client using Makefile so my concern about our ability to test it doesn't stand. Sorry for the noise.

jforissier:
@etienne-lms 

> The linaro OpenEmbedded recipes do use these makefile rather than the CMake

Then whoever is maintaining them will need to update the paths I suppose?

etienne-lms:
Definitely that an issue. I wonder how we could smooth the transition. Add a symlink from **tee-supplicant** to the **../sbin/tee-supplicant**? I would need to build and test the distrib to see what's to be fixed.

jenswi-linaro:
Overriding the values of  `BINDIR`, `LIBDIR` and `INCLUDEDIR` I guess.

vchong:
>> The linaro OpenEmbedded recipes do use these makefile rather than the CMake:
> That's a bit odd, since I believe both Buildroot and OE suggest to use CMake or Autotools first and plain Makefiles is the last resort thing. Having that said, I could have misunderstood that.

Iirc the recipes use the makefiles because we didn't have CMake or Autotools available then.

Makefile
```diff
@@ -12,9 +12,9 @@ export VPREFIX
 
 EXPORT_DIR ?= $(O)/export
 DESTDIR ?= $(EXPORT_DIR)
-BINDIR ?= /bin
-LIBDIR ?= /lib
-INCLUDEDIR ?= /include
+BINDIR ?= /usr/sbin
```
etienne-lms: > Overriding the values of `BINDIR`, `LIBDIR` and `INCLUDEDIR` I guess.

Sounds the smater way :) thanks.

That questions me. If installing in `sbin/`, we should set `SBINDIR` here, not `BINDIR`. I hardly found how far `SBINDIR` is as widely spread as `BINDIR` for such package install purpose.

jforissier: True.

etienne-lms:
Updated commit content with s/`BINDIR`/`SBINDIR`/. 
@jenswi-linaro & @jforissier, as the change is quite small and obvious I kept your tag applied.

jforissier:
`s@/sbin@/usr/sbin@` in the commit message...

jforissier:
Still some more nitpicking on the commit subject: "fix" seems a bit strong, nothing was actually broken, the path was just unusual. And you could say what the patch does so that `git log --oneline` is more clear. How about "Makefile: install tee-supplicant in /usr/sbin by default"?

etienne-lms:
Thanks for being nitpicking :) Agree with the suggestion.

etienne-lms:
split in 2 commits, change commit message, removed tags, waiting for your approval to put back tags.

jforissier:
Uh, it seems I focused a bit too much on tee-supplicant, I did not intend to ask you to split the patch ;-)
But anyway it's fine by me as it is now, thanks!

jenswi-linaro:

etienne-lms:
ok. I split the commit to simplify the commit messages (I was lacking of inspiration to get a one-line synthetic description).
Tags applied. Tanks.

PR #44, opened 2016-01-14
RPMB emulation HMAC support

jforissier:
~~This PR is on top of https://github.com/OP-TEE/optee_client/pull/43, so please consider only the last commit, c74fbb2 ("RPMB: add HMAC to emulation mode").~~

PR #43, opened 2016-01-12
Replay Protected Memory Block (RPMB) support

jforissier:
This commit adds support for the TEE_RPC_RPMB command, used by OP-TEE
to access the RPMB partition of an eMMC device (/dev/mmcblkXrpmb).
This is independent from the regular filesystem access (TEE_RPC_FS),
so the TEE can use either RPMB or the regular FS (or both) to implement
persistent storage.

RPMB operations are:
 - Get device information (partition size, reliable write count)
 - Read write counter
 - Read and write data
 - Program authentication key

The code relies on the ioctl() interface to implement the required
functions. In addition, an emulation layer is provided so that the RPMB
feature may be tested even if an actual eMMC chip is not available, or
if the kernel does not support it. Emulated mode is currently the
default. It can be disabled in the tee-supplicant Makefile
(-DRPMB_EMU=1).

The eMMC emulation layer does not yet perform any authentication, i.e.,
it will not check the MAC contained in the requests, and it will not
sign the responses with a MAC either. This feature will be added later.

Tested on HiKey using the extended xtest suite (including tests 6001 to
6011) and eMMC emulation only.

Signed-off-by: Jerome Forissier <jerome.forissier@linaro.org>

PR #42, opened 2015-12-17
Fix trace

pascal-brand-st-dev:
Signed-off-by: Pascal Brand <pascal.brand@st.com>

PR #145, opened 2019-02-27
Flag NULL pointer using invalid shm id

cneveux:
Refering to GPTS test suite v2.0.0.1-2016-11-09 id de-14-33

Refer to the TEEC_RegisterSharedMemory function for error conditions
which can be triggered during temporary registration of a memory region.

The CA needs to be able to send a null pointer to the TA. Previously
this was done by allocating a block of shared memory but setting the
size to 0.  This did not allow sending a non-NULL pointer with size 0,
this failing GPTS tests.
This change flags a NULL pointer with an invalid shared memory id rather
than size 0.

Signed-off-by: Michael Whitfield <michael.whitfield@nxp.com>
Signed-off-by: Cedric Neveux <cedric.neveux@nxp.com>

### DIFF
--- a/libteec/include/linux/tee.h
+++ b/libteec/include/linux/tee.h
@@ -48,8 +48,11 @@
 
 #define TEE_MAX_ARG_SIZE	1024
 
-#define TEE_GEN_CAP_GP		(1 << 0)/* GlobalPlatform compliant TEE */
-#define TEE_GEN_CAP_REG_MEM	(1 << 2)/* Supports registering shared memory */
+#define TEE_GEN_CAP_GP		(1 << 0) /* GlobalPlatform compliant TEE */
+#define TEE_GEN_CAP_REG_MEM	(1 << 2) /* Support registering shared memory */
+#define TEE_GEN_CAP_MEMREF_NULL	(1 << 3) /* Support NULL MemRef */
+
+#define TEE_MEMREF_NULL		((__u64)-1) /* NULL MemRef Buffer */
 
 /*
  * TEE Implementation ID
@@ -242,6 +245,14 @@ struct tee_ioctl_buf_data {
  * a part of a shared memory by specifying an offset (@shm_offs) and @size
  * of the object. To supply the entire shared memory object set @shm_offs
  * to 0 and @size to the previously returned size of the object.
+ *
+ * As per GlobalPlatform Client API v1.0 specification, a client can present
+ * a NULL shared memory reference when querying an output (or in/out) shared
+ * memory size.
+ *
+ * If a null reference is passed to a TA in the TEE, the field @c of
+ * structure tee_ioctl_param_value must be set to TEE_MEMREF_NULL indicating
+ * a NULL memory reference.
  */
 struct tee_ioctl_param_memref {
 	__u64 shm_offs;
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -151,6 +151,7 @@ TEEC_Result TEEC_InitializeContext(const char *name, TEEC_Context *ctx)
 		if (fd >= 0) {
 			ctx->fd = fd;
 			ctx->reg_mem = gen_caps & TEE_GEN_CAP_REG_MEM;
+			ctx->memref_null = gen_caps & TEE_GEN_CAP_MEMREF_NULL;
 			return TEEC_SUCCESS;
 		}
 	}
@@ -190,13 +191,23 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 	}
 	shm->size = tmpref->size;
 
-	res = TEEC_AllocateSharedMemory(ctx, shm);
-	if (res != TEEC_SUCCESS)
-		return res;
+	if (!tmpref->buffer && ctx->memref_null) {
+		if (tmpref->size)
+			return TEEC_ERROR_BAD_PARAMETERS;
 
-	memcpy(shm->buffer, tmpref->buffer, tmpref->size);
+		/* Null pointer, indicate no shared memory attached */
+		param->u.memref.shm_id = TEE_MEMREF_NULL;
+		shm->id = -1;
+	} else {
+		res = TEEC_AllocateSharedMemory(ctx, shm);
+		if (res != TEEC_SUCCESS)
+			return res;
+
+		memcpy(shm->buffer, tmpref->buffer, tmpref->size);
+		param->u.memref.shm_id = shm->id;
+	}
+
 	param->u.memref.size = tmpref->size;
-	param->u.memref.shm_id = shm->id;
 	return TEEC_SUCCESS;
 }
 
--- a/public/tee_client_api.h
+++ b/public/tee_client_api.h
@@ -256,6 +256,7 @@ typedef struct {
 	/* Implementation defined */
 	int fd;
 	bool reg_mem;
+	bool memref_null;
 } TEEC_Context;
 
 /**


jforissier:
Why is IBART failing?

libteec/src/tee_client_api.c
```diff
@@ -186,13 +186,18 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
...
+		res = TEEC_AllocateSharedMemory(ctx, shm);
+		if (res != TEEC_SUCCESS)
+			return res;
 
-	memcpy(shm->buffer, tmpref->buffer, tmpref->size);
+		memcpy(shm->buffer, tmpref->buffer, tmpref->size);
+		param->u.memref.shm_id = shm->id;
+	} else {
+		/* Null pointer, indicate no shared memory attached */
+		param->u.memref.shm_id = -1;
```
jenswi-linaro: Please use `TEE_MEMREF_NULL` from `<linux/tee.h>`

cneveux: ok

libteec/src/tee_client_api.c
```diff
@@ -186,13 +187,19 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 	}
 	shm->size = tmpref->size;
 
-	res = TEEC_AllocateSharedMemory(ctx, shm);
-	if (res != TEEC_SUCCESS)
-		return res;
+	if ((!tmpref->buffer) && (ctx->memref_null)) {
+		/* Null pointer, indicate no shared memory attached */
+		param->u.memref.shm_id = TEE_MEMREF_NULL;
```
jenswi-linaro: `teec_free_temp_refs()` could be left unchanged by setting `shm->id = -1` here.

libteec/src/tee_client_api.c
```diff
@@ -186,13 +187,19 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 	}
 	shm->size = tmpref->size;
 
-	res = TEEC_AllocateSharedMemory(ctx, shm);
-	if (res != TEEC_SUCCESS)
-		return res;
+	if ((!tmpref->buffer) && (ctx->memref_null)) {
```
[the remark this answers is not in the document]
cneveux: ok, done

libteec/src/tee_client_api.c
cneveux: correct, fixed

jenswi-linaro:
Please squash the two commits. With that:

libteec/include/linux/tee.h
```diff
@@ -50,6 +50,9 @@
 
 #define TEE_GEN_CAP_GP		(1 << 0)/* GlobalPlatform compliant TEE */
 #define TEE_GEN_CAP_REG_MEM	(1 << 2)/* Supports registering shared memory */
+#define TEE_GEN_CAP_MEMREF_NULL	(1 << 3)/* Support NULL MemRef */
+
+#define TEE_MEMREF_NULL (__u64)(-1) /* NULL MemRef Buffer */
```
etienne-lms: minor: missing tab before value.

libteec/src/tee_client_api.c
```diff
@@ -186,13 +187,20 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 	}
 	shm->size = tmpref->size;
 
-	res = TEEC_AllocateSharedMemory(ctx, shm);
-	if (res != TEEC_SUCCESS)
-		return res;
+	if (!tmpref->buffer && ctx->memref_null) {
+		/* Null pointer, indicate no shared memory attached */
```
etienne-lms: this should reject non null `tmpref->size`.

libteec/include/linux/tee.h
```diff
@@ -242,6 +245,16 @@ struct tee_ioctl_buf_data {
  * a part of a shared memory by specifying an offset (@shm_offs) and @size
  * of the object. To supply the entire shared memory object set @shm_offs
  * to 0 and @size to the previously returned size of the object.
+ *
+ * A client may need to present a NULL pointer in the argument
+ * passed to a trusted application in the TEE.
+ * This is also a requirement in GlobalPlatform Client API v1.0c
+ * (section 3.2.5 memory references), which can be found at
+ * http://www.globalplatform.org/specificationsdevice.asp
```
etienne-lms: Remove the url from the comment. The centralized mean for this reference is currently in optee_docs.
I think you could rephrase this paragraph as:
```
As per GlobalPlatform Client API v1.0 specification, a client can present
a NULL shared memory reference when querying an output (or in/out) shared
memory size.
```

libteec/src/tee_client_api.c
cneveux: Sorry I don't understand your point here?

libteec/include/linux/tee.h
cneveux: ok

libteec/include/linux/tee.h
cneveux: sure

libteec/src/tee_client_api.c
etienne-lms: If `tmpref->buffer` is NULL then the buffer size **must** be null.
I though here was a good place to check that the buffer size is really null when the buffer reference is `NULL`, i.e:
 `if (tmpref->size) return TEEC_ERROR_BAD_PARAMETERS;`.

cneveux: ok. Fixed and added subtest to verify

libteec/include/linux/tee.h
```diff
@@ -50,6 +50,9 @@
 
 #define TEE_GEN_CAP_GP		(1 << 0)/* GlobalPlatform compliant TEE */
 #define TEE_GEN_CAP_REG_MEM	(1 << 2)/* Supports registering shared memory */
+#define TEE_GEN_CAP_MEMREF_NULL	(1 << 3)/* Support NULL MemRef */
+
+#define TEE_MEMREF_NULL		(__u64)(-1) /* NULL MemRef Buffer */
```
etienne-lms: nitpicking: prefer `((__u64)-1)`

libteec/include/linux/tee.h
```diff
@@ -242,6 +245,14 @@ struct tee_ioctl_buf_data {
  * a part of a shared memory by specifying an offset (@shm_offs) and @size
  * of the object. To supply the entire shared memory object set @shm_offs
  * to 0 and @size to the previously returned size of the object.
+ *
+ * As per GlobalPlatform Client API v1.0 specification, a client can present
+ * a NULL shared memory reference when querying an output (or in/out) shared
+ * memory size.
+ *
+ * If a NULL pointer is passed to a TA in the TEE, the (@c)
```
etienne-lms: replace `NULL pointer` with `null reference`. A null reference is a `NULL` pointer and a null byte size.

Also, I suggest to replace "`the (@c) IOCTL parameters value must be ...`" with "`field @c of struct tee_ioctl_param_value must be...`".

libteec/include/linux/tee.h
cneveux: ok

libteec/include/linux/tee.h
cneveux: sur

jbech-linaro:
@OP-TEE/maintainers , this issue popped up again in another thing we're working with. I guess we're all happy with it and it seems like @cneveux pushed fixes for the last review comments.

jforissier:
@jbech-linaro yep, perhaps @etienne-lms wants to provide a R-b/A-b tag?

vchong:
@cneveux Please update commit message per Etienne's comment, add R-b and A-b tags above to the commit message, rebase the patch onto master and force push to your branch.

cneveux:
> 
> 
> @cneveux Please update commit message per Etienne's comment, add R-b and A-b tags above to the commit message, rebase the patch onto master and force push to your branch.

Done

jforissier:
@cneveux you have not updated the subject (**libutee:** ...)

jforissier:
Still not good :cry:
And what about https://travis-ci.org/github/OP-TEE/optee_client/builds/661914859#L369?
```
  CC      src/tee_client_api.c
In file included from src/tee_client_api.c:47:0:
src/tee_client_api.c: In function ‘teec_open_dev’:
/home/travis/build/OP-TEE/optee_client/libteec/include/linux/tee.h:51:25: error: implicit declaration of function ‘BIT’ [-Werror=implicit-function-declaration]
 #define TEE_GEN_CAP_GP  BIT(0) /* GlobalPlatform compliant TEE */
                         ^
src/tee_client_api.c:84:24: note: in expansion of macro ‘TEE_GEN_CAP_GP’
  if (!(vers.gen_caps & TEE_GEN_CAP_GP))
                        ^
src/tee_client_api.c:84:2: warning: nested extern declaration of ‘BIT’ [-Wnested-externs]
  if (!(vers.gen_caps & TEE_GEN_CAP_GP))
  ^
```

cneveux:
BIT macro doesn't exist in optee client, hence I can not fix the 
CHECK: Prefer using the BIT macro

#42: FILE: libteec/include/linux/tee.h:51:

+#define TEE_GEN_CAP_GP		(1 << 0) /* GlobalPlatform compliant TEE */

CHECK: Prefer using the BIT macro

#43: FILE: libteec/include/linux/tee.h:52:

+#define TEE_GEN_CAP_REG_MEM	(1 << 2) /* Support registering shared memory */

CHECK: Prefer using the BIT macro

#44: FILE: libteec/include/linux/tee.h:53:

+#define TEE_GEN_CAP_MEMREF_NULL	(1 << 3) /* Support NULL MemRef */

jforissier:
The checkpatch warning is OK -- I was mentioning a compile error. In your previous push you were using the `BIT()` macro.

Looks like we're good now.

jbech-linaro:
> I was mentioning a compile error.

Weird, I was running with the rather old patch as late as yesterday and I saw no compiler issues. Anyhow, thanks @cneveux and the rest of you involved in this.

libteec/src/tee_client_api.c
jbech-linaro: Need to ask, we ended up with 
https://github.com/OP-TEE/optee_client/blob/aea20819b1ec55fc4c72b0d20e21109169726d7d/libteec/src/tee_client_api.c#L194-L196

Is that really correct? Imagine the use cases where you have an API where you need to call the same API twice. First time to query for the size, next you allocate the buffer based on the size from the first call and then call the secures side again, something like this half pseudo code.

```c
// Query
op.params[0].tmpref.buffer = NULL;
op.params[0].tmpref.size = 0;

TEEC_InvokeCommand(...);

buf_size = op.params[0].tmpref.size;
buf = malloc(buf_size);

// Do the actual call
op.params[0].tmpref.buffer = buf;
op.params[0].tmpref.size = buf_size;

TEEC_InvokeCommand(...);
```

With the check in `teec_pre_process_tmpref(..)`, then above you will not work and I think it should? @etienne-lms ? I.e., I'd argue that the ` if (tmpref->size) ` shouldn't be in `teec_pre_process_tmpref(..)`?

jbech-linaro: > ```c
> op.params[0].tmpref.size = 0;
> ```

I meant to write something other than 0 there. But I suppose you meant that for queries the caller should always provide a zero?

etienne-lms: I'd say that if the buffer ref is NULL, when invoking, then the size value should be meaningless.
When back from invocation, then buffer may be NULL and size value be meaningful, be that's not in the scope of `teec_pre_process_tmpref()`, rather `teec_pre_process_tmpref()` that already handle this case, 

So i think https://github.com/OP-TEE/optee_client/blob/aea20819b1ec55fc4c72b0d20e21109169726d7d/libteec/src/tee_client_api.c#L194-L196 is still valid.
Note that an alternate could be to discard `tmpref->size` value force `param->u.memref.size = tmpref->size` be 0 (line 210):
https://github.com/OP-TEE/optee_client/blob/aea20819b1ec55fc4c72b0d20e21109169726d7d/libteec/src/tee_client_api.c#L194-L210

Did you find out that these lines were the cause of the failing xtest @jforissier  mentionned in https://github.com/OP-TEE/optee_test/pull/333#issuecomment-609819289?

jbech-linaro: > I'd say that if the buffer ref is NULL, when invoking, then the size value should be meaningless.

I think that is what I'm also looking for, but the size value isn't "meaningless", it has to be 0 that is passed otherwise we'll return an error. It's not a big deal as such I think. But I think it'd be better to let the TA for example check the size value to decide if it is bogus or not. I have no good example in mind, but there could probably be cases where you want to use the size parameter with some valid value (not 0) and still want to have it populated. But as said, not a big deal, just wanted to hear what the rest of you are thinking.

jbech-linaro: > Did you find out that these lines were the cause of the failing xtest @jforissier mentionned in [OP-TEE/optee_test#333 (comment)](https://github.com/OP-TEE/optee_test/pull/333#issuecomment-609819289)?

No, but I got a regression in our Widevine code, because I didn't have size set to zero. So, it was an easy fix, but as mentioned already I think that requirement is a bit overkill. But regardless that, it works as it is supposed to.

PR #146, opened 2019-02-28
Fix build warnings and initialize functions local variables

etienne-lms:
A first commit to fix a build issue reported by Buildroot team.
A second commit to initialize (almost) all functions local variables. May prevent build errors in the future.

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -66,8 +66,10 @@ static void teec_mutex_unlock(pthread_mutex_t *mu)
 static int teec_open_dev(const char *devname, const char *capabilities,
 			 uint32_t *gen_caps)
 {
-	struct tee_ioctl_version_data vers = { 0 };
 	int fd = 0;
+	struct tee_ioctl_version_data vers;
+
+	memset(&vers, 0, sizeof(vers));
 
 	fd = open(devname, O_RDWR);
 	if (fd < 0)
@@ -104,9 +106,10 @@ err:
 static int teec_shm_alloc(int fd, size_t size, int *id)
 {
 	int shm_fd = 0;
-	struct tee_ioctl_shm_alloc_data data = { 0 };
+	struct tee_ioctl_shm_alloc_data data;
 
 	memset(&data, 0, sizeof(data));
+
 	data.size = size;
 	shm_fd = ioctl(fd, TEE_IOC_SHM_ALLOC, &data);
 	if (shm_fd < 0)
@@ -118,9 +121,10 @@ static int teec_shm_alloc(int fd, size_t size, int *id)
 static int teec_shm_register(int fd, void *buf, size_t size, int *id)
 {
 	int shm_fd = 0;
-	struct tee_ioctl_shm_register_data data = { 0 };
+	struct tee_ioctl_shm_register_data data;
 
 	memset(&data, 0, sizeof(data));
+
 	data.addr = (uintptr_t)buf;
 	data.length = size;
 	shm_fd = ioctl(fd, TEE_IOC_SHM_REGISTER, &data);
@@ -485,13 +489,16 @@ TEEC_Result TEEC_OpenSession(TEEC_Context *ctx, TEEC_Session *session,
 			TEEC_CONFIG_PAYLOAD_REF_COUNT *
 				sizeof(struct tee_ioctl_param)) /
 			sizeof(uint64_t)] = { 0 };
-	struct tee_ioctl_buf_data buf_data = { 0 };
 	struct tee_ioctl_open_session_arg *arg = NULL;
 	struct tee_ioctl_param *params = NULL;
 	TEEC_Result res = 0;
 	uint32_t eorig = 0;
-	TEEC_SharedMemory shm[TEEC_CONFIG_PAYLOAD_REF_COUNT] = { 0 };
 	int rc = 0;
+	TEEC_SharedMemory shm[TEEC_CONFIG_PAYLOAD_REF_COUNT];
+	struct tee_ioctl_buf_data buf_data;
+
+	memset(&shm, 0, sizeof(shm));
+	memset(&buf_data, 0, sizeof(buf_data));
 
 	(void)&connection_data;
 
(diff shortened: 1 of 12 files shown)


jbech-linaro:
Thanks @etienne-lms for doing this (a tedious task). I've test built both using GNU Make as well as with CMake, both works fine. I've also reviewed the code, so ...

libteec/src/tee_client_api.c
```diff
@@ -66,8 +66,8 @@ static void teec_mutex_unlock(pthread_mutex_t *mu)
 static int teec_open_dev(const char *devname, const char *capabilities,
 			 uint32_t *gen_caps)
 {
-	struct tee_ioctl_version_data vers;
-	int fd;
+	struct tee_ioctl_version_data vers = { 0 };
```
jenswi-linaro: For you information. I've recently got error reports (warnings) on some tee driver patches with this kind of initialization. It seems it's a bit compiler specific which way structs can be initialized without warnings. So in the kernel it's recommended to initialize those with `memset()` instead. I guess initialization with some designation (for instance as `{ .impl_id = 0 }`) should be safe against all warnings.

jbech-linaro: Do you know if someone has talked to the affected toolchain people? To me it sound like a bug in whatever toolchain that reports error when doing like this.

jenswi-linaro: This is all I know https://lkml.org/lkml/2019/2/19/170

jbech-linaro: Yikes, interesting email thread. So yet another thing where it seems like we cannot trust the compiler(s). We will probably have to support clang within a year (for AOSP builds), so in that case better to change to memset right away. Too bad, since the `= { 0 }` initialization is pretty compact.

jenswi-linaro: Yes, I used to like it too.

jforissier: @jbech-linaro yep and I liked GCC's extension `= { }` even better. But I see there are serious reasons to prefer `memset()` indeed.

etienne-lms: ok, I'll move to `memset()`. I think using ` = { <filed-label> = <compliant-value> };` is painful. `enum` variable are also harder to init. 

Notice the `TEEC_Result` variables. I initially initialized such with `TEEC_SUCCESS`. After what I though that starting a _secure_ sequence default assuming its success was bad. I changed to `TEEC_ERROR_GENERIC`, looked better, I think, even in both case, such init value are discarded at build time.

jbech-linaro: > Notice the `TEEC_Result` variables...

Agree, that is better.

etienne-lms: Vote time: +1 or -1 this way? `memset()` is called inside to variable definition block.
```C
static TEEC_Result tee_socket_ioctl_udp(int fd, uint32_t command,
					void *buf, size_t *blen)
{
	TEEC_Result res = TEEC_ERROR_GENERIC;
	struct sockaddr_storage sass;
	memset(&sass, 0, sizeof(sass));
	struct sockaddr *sa = (struct sockaddr *)&sass;
	socklen_t len = sizeof(sass);
	uint16_t port = 0;

	if (getpeername(fd, sa, &len))

	(...)
```

etienne-lms: Alternate: group structures inited from `memset()` below those inited from ` = ...`.
```C
static TEEC_Result tee_socket_ioctl_udp(int fd, uint32_t command,
					void *buf, size_t *blen)
{
	TEEC_Result res = TEEC_ERROR_GENERIC;
	struct sockaddr *sa = (struct sockaddr *)&sass;
	socklen_t len = sizeof(sass);
	uint16_t port = 0;
	struct sockaddr_storage sass;

	memset(&sass, 0, sizeof(sass));

	if (getpeername(fd, sa, &len))

	(...)
```

jforissier: Does checkpatch accept this? I don't think so (but have not tried).

libteec/src/tee_client_api.c
```diff
@@ -504,7 +511,7 @@ TEEC_Result TEEC_OpenSession(TEEC_Context *ctx, TEEC_Session *session,
 	buf_data.buf_ptr = (uintptr_t)buf;
 	buf_data.buf_len = sizeof(buf);
 
-	arg = (struct tee_ioctl_open_session_arg *)buf;
+	arg = (struct tee_ioctl_open_session_arg *)(void *)buf;
```
jenswi-linaro: Why the extra cast? I was expecting the alignment of buf to be large enough.

etienne-lms: There is this setup where GCC (`--version` returns `ls (GNU coreutils) 8.21`) complains, see http://lists.busybox.net/pipermail/buildroot/2019-February/243437.html.

```
/home/thomas/projets/buildroot/output/build/optee-client-3.4.0/libteec/src/tee_client_api.c: In function ‘TEEC_InvokeCommand’:
/home/thomas/projets/buildroot/output/build/optee-client-3.4.0/libteec/src/tee_client_api.c:581:8: error: cast increases required alignment of target type [-Werror=cast-align]
  arg = (struct tee_ioctl_invoke_arg *)buf;
        ^
```

jenswi-linaro: That is surprising since `buf` already is supposed to have the alignment of `uint64_t`. Are we missing something? How come `struct tee_ioctl_invoke_arg` needs even greater alignment?
This cast could hide a real problem.
Perhaps it's better to do something like:
```
union {
    struct tee_ioctl_open_session_arg arg;
    uint8_t data[sizeof(struct tee_ioctl_open_session_arg) + TEEC_CONFIG_PAYLOAD_REF_COUNT * sizeof(struct tee_ioctl_param)];
} buf;

arg = &buf.arg;

etienne-lms: Is it related to the structure, or something toolchains are likely to complain when converting opaque references into effective memory pointers?

etienne-lms: I was not found of the `union` I could not find another way aside intermediate cast to `void *`.
I've pushed a fixup with the change proposed by @jenswi-linaro.

etienne-lms:
Superseded.
I've opened https://github.com/OP-TEE/optee_client/pull/150 to focus on the build warning issue. 
I will push a later P-R for the local variables init once #150 completes.

PR #151, opened 2019-03-20
explicitly initialize local variables

etienne-lms:
This change initializes all local variables to prevent build issues
(warnings and/or errors) in OP-TEE client package.

Use memset() to init structured and typed variables. This change
changes ordering in the local variable definition block at function
head. Structured variables are defined below, right above the memset()
call block for their initialization, when possible.

### DIFF
--- a/tee-supplicant/src/sha2.c
+++ b/tee-supplicant/src/sha2.c
@@ -121,12 +121,13 @@ uint32 sha256_k[64] =
 static void sha256_transf(sha256_ctx *ctx, const unsigned char *message,
                           unsigned int block_nb)
 {
-    uint32 w[64];
-    uint32 wv[8];
-    uint32 t1, t2;
-    const unsigned char *sub_block;
-    int i;
-    int j;
+    uint32 w[64] = { 0 };
+    uint32 wv[8] = { 0 };
+    uint32 t1 = 0;
+    uint32 t2 = 0;
+    const unsigned char *sub_block = NULL;
+    int i = 0;
+    int j = 0;
 
     for (i = 0; i < (int) block_nb; i++) {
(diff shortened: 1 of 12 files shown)


etienne-lms:
Followup of https://github.com/OP-TEE/optee_client/pull/146.

tee-supplicant/src/sha2.c
```diff
@@ -175,6 +179,7 @@ void sha256(const unsigned char *message, unsigned int len, unsigned char *diges
 void sha256_init(sha256_ctx *ctx)
 {
     int i;
```
jenswi-linaro: Initialization missing

tee-supplicant/src/tee_socket.c
```diff
@@ -355,7 +355,7 @@ static void ts_add(const struct timespec *a, const struct timespec *b,
 static int ts_diff_to_polltimeout(const struct timespec *a,
 				  const struct timespec *b)
 {
-	struct timespec diff;
+	struct timespec diff = { 0 };
```
jenswi-linaro: This should probably be initialized with `memset()` instead.

tee-supplicant/src/sha2.c
etienne-lms: thanks

tee-supplicant/src/tee_socket.c
etienne-lms: right, thanks.

etienne-lms:
Squashed, tag applied.

jforissier:
Thanks @etienne-lms, I have merged this manually with one very minor change: use an uppercase "E" in the subject line ;-)

etienne-lms:
ok, thanks.

PR #154, opened 2019-04-04
Initialize structured local variable where defined

etienne-lms:
Prefer using = { } rather than explicit memset() for default
initialization of structure local variables. The rational is that
memset() are implemented below variable definition block whereas
such default initialization is more explicit if done exactly where
the variable is define, non structured/enumerated type local variables.

When not possible, use = { 0 } or memset().

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -598,13 +598,13 @@ static void close_mmc_fd(int fd)
 static uint32_t read_ext_csd(int fd, uint8_t *ext_csd)
 {
 	int st = 0;
-	struct mmc_ioc_cmd cmd;
+	struct mmc_ioc_cmd cmd = {
+		.blksz = 512,
+		.blocks = 1,
+		.flags = MMC_RSP_R1 | MMC_CMD_ADTC,
+		.opcode = MMC_SEND_EXT_CSD,
+	};
 
-	memset(&cmd, 0, sizeof(cmd));
-	cmd.blksz = 512;
-	cmd.blocks = 1;
-	cmd.flags = MMC_RSP_R1 | MMC_CMD_ADTC;
-	cmd.opcode = MMC_SEND_EXT_CSD;
 	mmc_ioc_cmd_set_data(cmd, ext_csd);
 
 	st = IOCTL(fd, MMC_IOC_CMD, &cmd);
@@ -621,15 +621,14 @@ static uint32_t rpmb_data_req(int fd, struct rpmb_data_frame *req_frm,
 	int st = 0;
 	size_t i = 0;
 	uint16_t msg_type = ntohs(req_frm->msg_type);
-	struct mmc_ioc_cmd cmd;
-
-	memset(&cmd, 0, sizeof(cmd));
-	cmd.blksz = 512;
-	cmd.blocks = req_nfrm;
-	cmd.data_ptr = (uintptr_t)req_frm;
-	cmd.flags = MMC_RSP_R1 | MMC_CMD_ADTC;
-	cmd.opcode = MMC_WRITE_MULTIPLE_BLOCK;
-	cmd.write_flag = 1;
+	struct mmc_ioc_cmd cmd = {
+		.blksz = 512,
+		.blocks = req_nfrm,
+		.data_ptr = (uintptr_t)req_frm,
+		.flags = MMC_RSP_R1 | MMC_CMD_ADTC,
+		.opcode = MMC_WRITE_MULTIPLE_BLOCK,
+		.write_flag = 1,
+	};
 
 	for (i = 1; i < req_nfrm; i++) {
 		if (req_frm[i].msg_type != msg_type) {
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -584,7 +584,6 @@ static bool process_one_request(struct thread_arg *arg)
 	memset(&request, 0, sizeof(request));
 
 	DMSG("looping");
-	memset(&request, 0, sizeof(request));
 	request.recv.num_params = RPC_NUM_PARAMS;
 
 	/* Let it be known that we can deal with meta parameters */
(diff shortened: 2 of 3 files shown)


jforissier:
> When not possible, use = { 0 } or memset().

When would this happen?

jenswi-linaro:
I'm a bit hesitant to this change. This is a pure user space library and I think it should be kept as portable as possible. Using `{ }` is an extension which may not be supported by the compiler. In user space the world is larger than GCC and Clang.

jforissier:
> In user space the world is larger than GCC and Clang.

Fair point indeed!

etienne-lms:
I'm ok with being portable. I assume all agree.
And I would push for always initializing local variable. 
Looks there are 3 ways: do not always init / init with `= { 0 }` / init with `memset()`.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -546,9 +530,7 @@ static bool find_params(union tee_rpc_invoke *request, uint32_t *func,
 static bool spawn_thread(struct thread_arg *arg)
 {
 	int e = 0;
-	pthread_t tid;
-
-	memset(&tid, 0, sizeof(tid));
+	pthread_t tid = { 0 };
```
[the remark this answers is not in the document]
jenswi-linaro: `pthread_t` is special. The type is implementation defined, so it can be a struct in some implementations (I've seen a few), but it can also be a scalar. The most portable way to initialize this is using `memset()`.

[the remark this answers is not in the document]
jforissier: OK I see. Thanks.

tee-supplicant/src/rpmb.c
```diff
@@ -623,7 +623,6 @@ static uint32_t rpmb_data_req(int fd, struct rpmb_data_frame *req_frm,
 	uint16_t msg_type = ntohs(req_frm->msg_type);
 	struct mmc_ioc_cmd cmd;
 
-	memset(&cmd, 0, sizeof(cmd));
```
jenswi-linaro: We should keep this.

etienne-lms: right, i'll restore.

tee-supplicant/src/rpmb.c
```diff
@@ -601,6 +601,7 @@ static uint32_t read_ext_csd(int fd, uint8_t *ext_csd)
 	struct mmc_ioc_cmd cmd;
 
 	memset(&cmd, 0, sizeof(cmd));
+
```
jenswi-linaro: I don't think this empty line is needed since the assignments below actually relies on first clearing the `cmd`. If we would have assinged `cmd` like
``` C
cmd = (struct mmc_ioc_cmd){ .blksz = 512, .blocks = 1,
                            .flags = MMC_RSP_R1 | MMC_CMD_ADTC,
                            .opcode = MMC_SEND_EXT_CSD, };
```
That would have been a different story.

etienne-lms: see latest fixup change.

tee-supplicant/src/rpmb.c
```diff
@@ -598,14 +598,13 @@ static void close_mmc_fd(int fd)
 static uint32_t read_ext_csd(int fd, uint8_t *ext_csd)
 {
 	int st = 0;
-	struct mmc_ioc_cmd cmd;
-
-	memset(&cmd, 0, sizeof(cmd));
+	struct mmc_ioc_cmd cmd = (struct mmc_ioc_cmd){
```
jenswi-linaro: When initializing instead of assigning there's no need for a compound statement, that is, you can drop `(struct mmc_ioc_cmd)` here.

etienne-lms: done.

jenswi-linaro:

etienne-lms:
Comments addressed.
Added a fixup commit for 2 bad `TEEC_Result` variable init.
IBART failed on xtest 1008.4 + all 60XX. Not found why. Not reproduced. Keep trying on alternate config/platform.

etienne-lms:
(force pushed fix for missing s-o-b in last fixup commit)

jbech-linaro:
> IBART failed on xtest 1008.4 + all 60XX. Not found why. Not reproduced. Keep trying on alternate config/platform.

Seems like the last run also failed on those.

tee-supplicant/src/rpmb.c
```diff
@@ -621,15 +621,14 @@ static uint32_t rpmb_data_req(int fd, struct rpmb_data_frame *req_frm,
...
-	cmd.data_ptr = (uintptr_t)req_frm;
-	cmd.flags = MMC_RSP_R1 | MMC_CMD_ADTC;
-	cmd.opcode = MMC_WRITE_MULTIPLE_BLOCK;
-	cmd.write_flag = 1;
+	struct mmc_ioc_cmd cmd = {
+		.blksz = 512,
+		.blocks = req_nfrm,
+		.data_ptr = (uintptr_t)req_frm,
+		.flags = MMC_RSP_R1 | MMC_CMD_ADTC,
+		.opcode = MMC_SEND_EXT_CSD,
```
etienne-lms: bug: restore `MMC_WRITE_MULTIPLE_BLOCK`.

etienne-lms: Not sure this fix will fix IBART failure report...

etienne-lms:
IBART's happy!
Commit series squashed and commit message updated.
No tag applied, waiting for feedback.

jbech-linaro:
> No tag applied, waiting for feedback.

What feedback?

etienne-lms:
> > No tag applied, waiting for feedback.
>
> What feedback?

Just a "ok, fine!" (the tags are a bit old since changes: fixups and commit message)
Your question sounds as such, I'll apply Jens' tag. Maybe your a-b?

jenswi-linaro:
Fine with me. :-)

jbech-linaro:
`Fine-by: Joakim Bech <joakim.bech@linaro.org>` :)

etienne-lms:
fancy tags applied.

PR #156, opened 2019-04-27
misc updates

vchong:


vchong:
For `config.mk: default CFG_TA_TEST_PATH to n`, https://github.com/OP-TEE/build/pull/363 is submitted to turn it back on to pass xtest 1008, but not too sure about the cmake part so please double check. Thanks!

jbech-linaro:

config.mk
```diff
@@ -20,7 +20,7 @@ CFG_TEE_SUPP_LOG_LEVEL?=1
 
 # CFG_TEE_DATA_PATH
 #   Specify the root path for the TEE data directory.
-CFG_TEE_DATA_PATH ?= /data/vendor/tee
+CFG_TEE_DATA_PATH ?= $(CFG_TEE_FS_PARENT_PATH)/vendor/tee
```
jenswi-linaro: This is a bit more than just cleaning up and making easier to read. Did it end up in the wrong commit?

vchong: Iiuu, based on the description in the file, the hierarchical relationship of the configs is as follows:
```
CFG_TEE_FS_PARENT_PATH ?= /data
CFG_TEE_DATA_PATH ?= $(CFG_TEE_FS_PARENT_PATH)/vendor/tee
CFG_TEE_LOGS_PATH ?= $(CFG_TEE_DATA_PATH)/logs
CFG_TEE_CLIENT_LOG_FILE ?= $(CFG_TEE_LOGS_PATH)/teec.log
```
i.e. the CFG_ defined one line above is used to expand the CFG_ of the current line. That was the reasoning for this change. Does it warrant a separate commit by itself?

jenswi-linaro: It's fine to keep it in this commit. Please update the commit message.

vchong: Pushed a fixup with additional changes, including new commit title and message. Please continue review. Thanks!

jenswi-linaro: It could be worth mentioning that `CFG_TEE_DATA_PATH` has a new default value partially based on `CFG_TEE_FS_PARENT_PATH` and that it's only makefiles you're touching. Because there is a functional change in the commit, it's not just reformatting. When you're scanning the git log it's helpful to know something about what done, "do minor file cleanup" doesn't say much.

vchong: I split these into 3 patches and forced push to hopefully make it easier to review and comment. The first 3 patches are 'new' while the last 3 remain unchanged. Please see if they're ok. Thanks!

vchong: Hrm, sorry for the confusion but on github it's the 1st, 2nd and last patch instead of the first 3.

```
{config,tee_supplicant_android}.mk: improve legibility
config.mk: add more config description
config.mk: redefine CFG_TEE_DATA_PATH
```

tee-supplicant/tee_supplicant_android.mk
```diff
@@ -31,19 +27,21 @@ LOCAL_SRC_FILES += src/handle.c \
...
 ifeq ($(CFG_GP_SOCKETS),y)
 LOCAL_SRC_FILES += src/tee_socket.c
+LOCAL_CFLAGS += -DCFG_GP_SOCKETS=1
 endif
 
-LOCAL_C_INCLUDES := $(LOCAL_PATH)/../public \
-    $(LOCAL_PATH)/../libteec/include \
-    $(LOCAL_PATH)/src
-
-LOCAL_SHARED_LIBRARIES := libteec
```
etienne-lms: remove one of the empty lines.

tee-supplicant/tee_supplicant_android.mk
```diff
@@ -31,19 +27,21 @@ LOCAL_SRC_FILES += src/handle.c \
...
 
 ifeq ($(CFG_TA_GPROF_SUPPORT),y)
 LOCAL_SRC_FILES += src/gprof.c
 LOCAL_CFLAGS += -DCFG_TA_GPROF_SUPPORT
 endif
 
+LOCAL_C_INCLUDES := $(LOCAL_PATH)/../public \
+    $(LOCAL_PATH)/../libteec/include \
+    $(LOCAL_PATH)/src
```
etienne-lms: used indentation as around line 23.

config.mk
```diff
@@ -20,11 +20,16 @@ CFG_TEE_SUPP_LOG_LEVEL?=1
 
 # CFG_TEE_FS_PARENT_PATH
 #   Path to folder that will contain TEE filesystem.
-#   This folder should be created before tee-supplicant start.
+#   This folder can be created with the required permission in an init
+#   script during boot, else it'll be created by the tee-supplicant on
```
etienne-lms: minor: prefer "it will be" rather than "it'll be". Same comment line 31.

config.mk
```diff
@@ -20,11 +20,16 @@ CFG_TEE_SUPP_LOG_LEVEL?=1
...
+#   the first instance the REE FS is required
```
etienne-lms: "on the first instance the REE FS is required": maybe rephrase into "on first REE FS access.".
(unless this was clear English :)

Same comment for line 32.

tee-supplicant/tee_supplicant_android.mk
etienne-lms: Discard this comment: fix by commit "tee_supplicant_android.mk: support RPMB_EMU".

vchong: The extra line should have been removed here and added back in the rpmb patch. Didn't notice it during the previous rebase. Will fix it in the next/final rebase.

tee-supplicant/tee_supplicant_android.mk
vchong: ok

config.mk
vchong: ok

config.mk
vchong: ok

vchong:
> for commit "config.mk: default CFG_TA_TEST_PATH to n":
in commit message, suggest to replace "during 'make' instead" with "by the build environment/recipe."

ok, will do this during rebase

> About commit "tee-supplicant: cmake: fix typo": it is OK RPMB_EMU defaults to ON since this is also a test config rather than a end device feature?

I think maybe the 'logic'(?) here is that RPMB support is off by default, and this is done by setting RPMB_EMU to ON. Maybe @jforissier can comment in better details?

vchong:
Pushed 2 fixup commits to address review comments.

jforissier:
@vchong 
> I think maybe the 'logic'(?) here is that RPMB support is off by default, and this is done by setting RPMB_EMU to ON. Maybe @jforissier can comment in better details?

Well, `RPMB_EMU` is just emulation. So when it's off, it doesn't mean that RPMB is not available, it means the the real RPMB device will be used instead. So yeah, it's kind of a testing mode as @etienne-lms said. It could very well default to off. And we could also rely CFG_RPMB_FS to disable the RPMB code entirely in tee-supplicant.

vchong:
> So when it's off, it doesn't mean that RPMB is not available, it means the the real RPMB device will be used instead.

Yes. I was trying to say that, for the default build, we don't expect any real RPMB device, so rpmb emu is on.

> It could very well default to off. And we could also rely CFG_RPMB_FS to disable the RPMB code entirely in tee-supplicant.

If you all don't mind, can we do this in a different PR since it seems like perhaps further discussion is required to reach a consensus?

jenswi-linaro:

etienne-lms:
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for all.

vchong:
Squashed and added tags. Ok for merge. Thanks!

PR #157, opened 2019-04-27
tee-supplicant: do not 'mkdir /data/tee' on startup

vchong:
'mkdir' only on the first instance the REE FS is required, because:

1. On systems that use RPMB only, this dir isn't necessary so there's
no point in creating an unused dir.

2. AOSP systems that use OP-TEE for disk encryption depends on the
tee-supplicant service to be up before properly mounting /data. If the
tee-supplicant in return depends on 'mkdir /data/tee/' before it can
start, this creates a circular dependency and causes boot to hang.
Removing this tee-supplicant startup dependency on /data is part of the
solution chain to enable the use of OP-TEE for AOSP disk encryption.

3. Most systems seem to take care of the creation of /data/tee with
the proper permission via some sort of init script anyway before
tee-supplicant is even started.

NOTE: /data is the default value assigned to TEE_FS_PARENT_PATH

Tested-by: Victor Chong <victor.chong@linaro.org> (QEMU v8)


tee-supplicant/src/tee_supp_fs.c
```diff
@@ -131,6 +115,27 @@ int tee_supp_fs_init(void)
 	return 0;
 }
 
+static size_t tee_fs_get_absolute_filename(char *file, char *out,
+					   size_t out_size)
+{
+	int s = 0;
+
+	if (!strlen(tee_fs_root)) {
```
jenswi-linaro: I'd rather move this block to `tee_supp_fs_process()`. The result should be more or less the same with the exception that this function remains without side effects as one may expect. It would also give the opportunity to give some return code that is less confusing than `TEEC_ERROR_BAD_PARAMETERS`, perhaps `TEEC_ERROR_GENERIC`. While you're at it please add an error print in case of failure.

vchong: Comments addressed

jenswi-linaro:

vchong:
Tag added and ready for merge. Thanks!

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -610,6 +610,14 @@ TEEC_Result tee_supp_fs_process(size_t num_params,
 	if (!num_params || !tee_supp_param_is_value(params))
 		return TEEC_ERROR_BAD_PARAMETERS;
 
+	if (strlen(tee_fs_root) == 0) {
```
vchong: There's an issue here. When code execution reaches here, `tee-supplicant` should already be up and running. Let's say, on first REE FS access here, `tee_supp_fs_init()` sets `tee_fs_root = "/data/tee"` but fails to `mkdir /data/tee`. Then every subsequent call here will skip this since the string length of `tee_fs_root` isn't 0 anymore and guaranteed to fail later down the path of execution since `/data/tee` doesn't exist. Can we maybe `memset tee_fs_root` back to `0` `if (tee_supp_fs_init() != 0)` below?

tee-supplicant/src/tee_supp_fs.c
```diff
@@ -610,6 +610,14 @@ TEEC_Result tee_supp_fs_process(size_t num_params,
...
+		if (tee_supp_fs_init() != 0) {
+			EMSG("error tee_supp_fs_init: failed to create %s/tee/",
+				TEE_FS_PARENT_PATH);
+			return TEEC_ERROR_GENERIC;
```
vchong: There's another issue here with `TEEC_ERROR_GENERIC`. When `optee_os` wants to load a TA, it will try to load from TA stores in the following order: PTA, early TA, secure storage and REE FS. Assume no PTA and TA is in REE FS. Loading from early TA will fail with item not found and `optee_os` will continue to try to load the TA from secure storage, but in this case, code execution will reach here and in the case of something going wrong, return `TEEC_ERROR_GENERIC` back to https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/kernel/user_ta.c#L881, which will cause `optee_os` to skip trying to load from REE FS, and fail `tee_ta_open_session()`.

Can we return `TEEC_ERROR_ITEM_NOT_FOUND`, or `#define TEEC_ERROR_STORAGE_NOT_AVAILABLE   0xF0100003` in `tee_client_api.h` and return it instead?

vchong:
Pushed a fixup commit for the 2 issues above for review.

tee-supplicant/src/tee_supp_fs.c
jforissier: Adding `TEEC_ERROR_STORAGE_NOT_AVAILABLE` sounds reasonable to me, because it is the code that best describes the error, and we already have done this kind of thing in the past (i.e., taking a code that is in the Internal Core spec and define it with a `TEEC_` prefix in the client API -- `TEEC_ERROR_EXTERNAL_CANCEL` and `TEEC_ERROR_TARGET_DEAD`).

vchong:
@jenswi-linaro any comments?

jenswi-linaro:
LGTM

vchong:
Squashed, rebased and tag already added. Please merge. Thanks!

PR #158, opened 2019-05-05
tee-supplicant: ioctl_emu_read_ctr(): MAC correct data

jenswi-linaro:
Prior to this patch was ioctl_emu_read_ctr() calculating a MAC over
among other things the uninitialized "op_result". "op_result" often
happens to be 0 which is the same as RPMB_RESULT_OK which is returned by
compute_hmac() on success. However, this field is not initialized before
calculating the MAC so it could very well differ resulting in an invalid
MAC. This patch fixes this by initializing "op_result" to RPMB_RESULT_OK
before calculating the MAC.


jforissier:

jenswi-linaro:
Tag applied

PR #160, opened 2019-05-10
simplify use of TEE fs related configs

vchong:
CFG_TEE_DATA_PATH confuses users because it's only used as a prefix for
CFG_TEE_LOGS_PATH and not the actual root path for the TEE data dir as
described. The actual root path is $(TEE_FS_PARENT_PATH)/tee as can be
seen under tee_supp_fs_init() in tee-supplicant/src/tee_supp_fs.c, so
just get rid of CFG_TEE_DATA_PATH.

Next, we move "/tee" from tee_supp_fs_init() to CFG_TEE_FS_PARENT_PATH,
because otherwise, we have the beginning part of the path in a makefile
(config.mk) and the ending in code (tee_supp_fs_init()), which seems
messy to manage and error prone. E.g. CFG_TEE_LOGS_PATH (or any of its
subsequently derived configs) depends on the actual full path of the TEE
fs, so we can just set
CFG_TEE_LOGS_PATH ?= $(CFG_TEE_FS_PARENT_PATH)/logs
rather than
CFG_TEE_LOGS_PATH ?= $(CFG_TEE_FS_PARENT_PATH)/tee/logs
where the "tee" here has to match the "tee" in code.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -6,10 +6,9 @@ LOCAL_PATH := $(call my-dir)
 # set CFG_TEE_CLIENT_LOAD_PATH before include config.mk
 CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib
 
-# set CFG_TEE_DATA_PATH before include config.mk
-CFG_TEE_DATA_PATH ?= /data/vendor/tee
+# set CFG_TEE_FS_PARENT_PATH before include config.mk
 TEEC_TEST_LOAD_PATH ?= /data/vendor/tee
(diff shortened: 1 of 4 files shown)


vchong:
hm.. lots of xtest errors..

vchong:
Fixed. Missed updating `tee-supplicant/CMakeLists.txt`.

vchong:
Commit message updated and tag added. Ok for merge? Thanks!

PR #162, opened 2019-06-14
config.mk: get rid of CFG_TEE_LOGS_PATH

vchong:
CFG_TEE_CLIENT_LOG_FILE was originally /data/teec.log. Commit 9a63135b
("Android.mk & config.mk: move teec.log/teesupp.log under /data/tee")
introduced CFG_TEE_LOGS_PATH which placed teec.log under
/data/tee/logs/, but in nowhere (here or build.git) do we create the
'logs' dir, so fopen /data/tee/logs/teec.log fails and teec.log is
never created. To make things simple, remove CFG_TEE_LOGS_PATH and
just place teec.log under /data/tee.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -25,14 +25,9 @@ CFG_TEE_SUPP_LOG_LEVEL?=1
 #   first REE FS access.
 CFG_TEE_FS_PARENT_PATH ?= /data/tee
 
-# CFG_TEE_LOGS_PATH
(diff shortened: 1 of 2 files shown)


jforissier:
Looks reasonable to me.

vchong:
Tag added and ok for merge. Thanks!

PR #163, opened 2019-06-21
rpmb: return RPMB_RESULT_AUTH_KEY_NOT_PROGRAMMED if key not set 

vchong:


vchong:
Depends on https://github.com/OP-TEE/optee_os/pull/3087.

Let me test this on QEMU first and post an update with the results.

vchong:
For `rpmb emu: increase EMU_RPMB_SIZE_MULT to 2`, should we consider increasing this to 4 MB instead? This is the size of all the real rpmb devices (3) I've seen so far.

vchong:
Link to `optee_os` details related to this PR.
https://github.com/OP-TEE/optee_os/pull/3087#issuecomment-504306575
https://github.com/OP-TEE/optee_os/pull/3087#discussion_r296324166

vchong:
> The defines you're adding are never used except RPMB_RESULT_AUTH_KEY_NOT_PROGRAMMED so I don't really see the point.
I'd rather add only the one that is used in the commit that uses it.

Removed unused defines.

> Please remove the defines from the commit message.

Ok, will remove when adding tag.

vchong:
All comments addressed

vchong:
Added a patch to change AOSP RPMB device path.

tee-supplicant/src/rpmb.c
```diff
@@ -341,6 +341,11 @@ static bool is_hmac_valid(struct rpmb_emu *mem, struct rpmb_data_frame *frm,
 	return true;
 }
 
+static uint16_t gen_msb1st_result(uint8_t byte)
+{
+	return (uint16_t)(byte << 8);
```
etienne-lms: prefer cast before shifting:  `return (uint16_t)byte << 8;`

vchong: Done. Thanks!

vchong:
Added patch to debug print `dev_id`. Please continue review. Thanks!

jforissier:
For "rpmb: add DMSG to print dev_id":

jenswi-linaro:
Please squash all commits with their fixups.

vchong:
@jenswi-linaro I'm still missing R-b tags for "rpmb: add missing defines for RPMB results", "rpmb emu: increase EMU_RPMB_SIZE_MULT to 2" and "rmpb: update AOSP RPMB device path".

jforissier:
@vchong it doesn't matter ;-) I suppose Jens would like to see what we have here without having to figure out the fixups. In fact, for relatively small patches, I find it easier to review the whole series again in the latest form. For bigger patches however, it's good to work incrementally.

vchong:
@jenswi-linaro @jforissier Rebased, squashed and added tags wherever appropriate. Please continue review. Thanks!

jenswi-linaro:
Looks good. For the commits:
```
rpmb: add missing defines for RPMB results
rpmb emu: return RPMB_RESULT_AUTH_KEY_NOT_PROGRAMMED if key not set
rpmb emu: remove gratuitous line
rpmb: add DMSG to print dev_id
```
Please apply:

and for the commits:
```
rpmb emu: fix ioctl return type
rpmb emu: increase EMU_RPMB_SIZE_MULT to 2
rmpb: update AOSP RPMB device path
```

vchong:
Rebased and tags applied. Ok to merge.

tee-supplicant/src/rpmb.c
```diff
@@ -93,6 +93,7 @@ struct rpmb_data_frame {
 #define RPMB_RESULT_GENERAL_FAILURE		0x01
 #define RPMB_RESULT_AUTH_FAILURE		0x02
 #define RPMB_RESULT_ADDRESS_FAILURE		0x04
+#define RPMB_RESULT_AUTH_KEY_NOT_PROGRAMMED	0x07
```
etienne-lms: dummy question: why not 0x3 or 0x5? below 0x7 are reserved?

jforissier: 0x3, 0x5, 0x6 are defined by the eMMC spec but are not used in the code, that's why they do not need to be here.

etienne-lms: ok, thanks. I should have re-read to comment few lines above.

PR #166, opened 2019-07-30
Travis: install package libc6-dev-armhf-cross

jforissier:
Travis CI have switched their default distribution from Ubuntu 14.04.5
to 16.04.6. Now we have the following build error:

   CC      src/tee_client_api.c
 In file included from /usr/include/errno.h:28:0,
                  from src/tee_client_api.c:28:
 /usr/include/features.h:367:25: fatal error: sys/cdefs.h: No such file or directory

The missing file is in the libc6-dev-armhf-cross package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ notifications:
 # Install the cross-compiler
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

PR #170, opened 2019-09-12
[RFC] tee-supplicant: add support for accessing custom REE service

divneil:
Signed-off-by: Divneil Rai Wadhawan <divneil.r.wadhawan@intel.com>

### DIFF
--- a/libteec/CMakeLists.txt
+++ b/libteec/CMakeLists.txt
@@ -22,9 +22,9 @@ set (CFG_TEE_CLIENT_LOG_FILE "/data/tee/teec.log" CACHE STRING "Location of libt
 # Source files
 ################################################################################
 set (SRC
+	src/ree_service_api.c
 	src/tee_client_api.c
 	src/teec_trace.c
-	src/ree_service_api.c
 )
 
 if (CFG_TEE_BENCHMARK)
--- a/libteec/CMakeLists.txt
+++ b/libteec/CMakeLists.txt
@@ -24,6 +24,7 @@ set (CFG_TEE_CLIENT_LOG_FILE "/data/tee/teec.log" CACHE STRING "Location of libt
 set (SRC
 	src/tee_client_api.c
 	src/teec_trace.c
+	src/ree_service_api.c
 )
 
 if (CFG_TEE_BENCHMARK)
--- a/libteec/Makefile
+++ b/libteec/Makefile
@@ -18,7 +18,8 @@ LIB_MAJ_MIN	:= $(LIB_NAME).$(MAJOR_VERSION).$(MINOR_VERSION)
 LIB_MAJ_MIN_P	:= $(LIB_NAME).$(MAJOR_VERSION).$(MINOR_VERSION).$(PATCH_VERSION)
 
 TEEC_SRCS	:= tee_client_api.c \
-		   teec_trace.c
+		   teec_trace.c \
+		   ree_service_api.c
 ifeq ($(CFG_TEE_BENCHMARK),y)
 TEEC_SRCS	+= teec_benchmark.c
 endif
--- a/libteec/Makefile
+++ b/libteec/Makefile
@@ -17,9 +17,9 @@ LIB_MAJOR	:= $(LIB_NAME).$(MAJOR_VERSION)
 LIB_MAJ_MIN	:= $(LIB_NAME).$(MAJOR_VERSION).$(MINOR_VERSION)
 LIB_MAJ_MIN_P	:= $(LIB_NAME).$(MAJOR_VERSION).$(MINOR_VERSION).$(PATCH_VERSION)
 
-TEEC_SRCS	:= tee_client_api.c \
-		   teec_trace.c \
-		   ree_service_api.c
+TEEC_SRCS	:= ree_service_api.c \
+		   tee_client_api.c \
+		   teec_trace.c
 ifeq ($(CFG_TEE_BENCHMARK),y)
 TEEC_SRCS	+= teec_benchmark.c
 endif
--- a/libteec/src/ree_service_api.c
+++ b/libteec/src/ree_service_api.c
@@ -1,41 +1,63 @@
+/*
+ * Copyright (C) 2019 Intel Corporation All Rights Reserved
+ */
 
+#include <errno.h>
 #include <sys/types.h>
 #include <sys/ipc.h>
 #include <sys/msg.h>
 #include <stdlib.h>
-#include <errno.h>
 
-#include <tee_client_api.h>
-#include <teec_trace.h>
 #include <ree_service_api.h>
+#include <teec_trace.h>
+#include <tee_client_api.h>
 
 #ifndef __aligned
 #define __aligned(x) __attribute__((__aligned__(x)))
 #endif
 #include <linux/tee.h>
 
-struct service {
+/*
+ * Internal structure to represent Message Queue Service.
+ * struct msgq_service
+ * @msgqid: Message Queue received from system
+ * @buf   : Buffer associated with service
+ * @buf_sz: Size of @buf
+ */
+struct msgq_service {
 	int msgqid;
 	void *buf;
 	size_t buf_sz;
 };
 
-static bool is_param_type_value(uint64_t param_type)
+/**
+ * param_type_is_value() - returns true if param type is value
+ */
+static bool param_type_is_value(uint64_t param_type)
 {
-	if (param_type == TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INPUT ||
-			param_type == TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_OUTPUT ||
-			param_type == TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INOUT)
+	switch(param_type) {
+	case TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INPUT:
+	case TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_OUTPUT:
+	case TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INOUT:
 		return true;
-	return false;
+	default:
+		return false;
+	}
 }
 
-static bool is_param_type_memref(uint64_t param_type)
+/**
+ * param_type_is_memref() - returns true if param type is memory reference
+ */
+static bool param_type_is_memref(uint64_t param_type)
 {
-	if (param_type == TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_INPUT ||
-			param_type == TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_OUTPUT ||
-			param_type == TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_INOUT)
+	switch (param_type) {
+	case TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_INPUT:
+	case TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_OUTPUT:
+	case TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_INOUT:
 		return true;
-	return false;
+	default:
+		return false;
+	}
 }
 
 /**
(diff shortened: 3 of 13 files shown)


libteec/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set (CFG_TEE_CLIENT_LOG_FILE "/data/tee/teec.log" CACHE STRING "Location of libt
 set (SRC
 	src/tee_client_api.c
 	src/teec_trace.c
+	src/ree_service_api.c
```
jforissier: Alphabetical order preferred, please

libteec/Makefile
```diff
@@ -18,7 +18,8 @@ LIB_MAJ_MIN	:= $(LIB_NAME).$(MAJOR_VERSION).$(MINOR_VERSION)
 LIB_MAJ_MIN_P	:= $(LIB_NAME).$(MAJOR_VERSION).$(MINOR_VERSION).$(PATCH_VERSION)
 
 TEEC_SRCS	:= tee_client_api.c \
-		   teec_trace.c
+		   teec_trace.c \
+		   ree_service_api.c
```
jforissier: Alphabetical order preferred, please

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
+
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/msg.h>
+#include <stdlib.h>
+#include <errno.h>
+
+#include <tee_client_api.h>
+#include <teec_trace.h>
+#include <ree_service_api.h>
```
jforissier: Alphabetical order preferred, please

public/tee_client_api.h
```diff
@@ -270,6 +270,8 @@ typedef struct {
 	uint8_t clockSeqAndNode[8];
 } TEEC_UUID;
 
+typedef TEEC_UUID REEC_UUID;
```
jforissier: Not needed

tee-supplicant/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ set (SRC
 	src/tee_supp_fs.c
 	src/tee_supplicant.c
 	src/teec_ta_load.c
+	src/tee_service.c
+	src/tee_service_handle.c
```
jforissier: Ordering, and name could be better. `ree_service.c` and `ree_service_handle.c` maybe?

tee-supplicant/src/optee_msg_supplicant.h
```diff
@@ -183,6 +183,7 @@
  */
 #define OPTEE_MSG_RPC_CMD_FTRACE	11
 
+#define OPTEE_MSG_RPC_CMD_GENERIC	30
```
jforissier: Name is too generic. `OPTEE_MSG_RPC_CMD_REE_SERVICE`? (or `OPTEE_MSG_RPC_CMD_REE_SRVC`).

tee-supplicant/src/optee_msg_supplicant.h
```diff
@@ -259,4 +260,25 @@
...
+ */
+#define OPTEE_MRC_GENERIC_OPEN		1
+
+/*
+ * Close REE Service
+ *
+ * [in]     param[0].u.value.a  OPTEE_MRC_GENERIC_CLOSE
+ * [in]     param[0].u.value.b  TA instance id
+ */
+#define OPTEE_MRC_GENERIC_CLOSE		2
```
jforissier: `OPTEE_MRC_REE_SERVICE_OPEN`, `OPTEE_MRC_REE_SERVICE_CLOSE`

tee-supplicant/src/tee_service.c
```diff
@@ -0,0 +1,502 @@
...
+#include <sys/ipc.h>
+#include <sys/msg.h>
+#include <errno.h>
+#include <dlfcn.h>
+
+#include <tee_client_api.h>
+#include <teec_trace.h>
+#include <tee_supplicant.h>
+#include <optee_msg_supplicant.h>
+#include <ree_service_api.h>
```
jforissier: Ordering

public/ree_service_api.h
```diff
@@ -0,0 +1,76 @@
...
+	} u;
+};
+
+TEEC_Result uuid_to_str(REEC_UUID *uuid, char *uuid_str, size_t size);
+TEEC_Result ree_service_init(REEC_UUID *uuid, void **service);
+void ree_service_exit(void *service);
+TEEC_Result ree_rcv_params(void *service, size_t *num_params,
+					struct tee_params *params);
+TEEC_Result ree_snd_params(void *service, size_t num_params,
+				struct tee_params *params, int32_t error);
```
jforissier: `ree_service_rcv()` and `ree_service_snd()` maybe (having service here is more useful than params IMO).

More general question: where is the actual service executed, in which context? As far as I can tell, you have two modes:
1. A dynamic library is loaded by tee-supplicant and a callback is invoked. Code therefore runs in the context of the tee-supplicant RPC loop.
2. A message queue interface. Core executes in a different process.
Is this correct? I think these function should be commented so it's clear who should use them and for which scenario.

What if someone needs to run the service in kernel space? (i.e., TA wants to talk to a Linux driver). That's a use case I've heard before.

tee-supplicant/src/tee_service.c
```diff
@@ -0,0 +1,502 @@
...
+
+	return result;
+}
+
+static TEEC_Result process_dlib_params(void *dl, size_t num_params,
+									struct tee_ioctl_param *params)
+{
+	size_t i;
+	TEEC_Result res = TEEC_SUCCESS;
+	TEEC_Result (*process_tee_params)(size_t num_params, struct tee_params *params);
```
jforissier: This API has to be declared and documented in some header file, because it needs to be implemented by the code that actually provides the service, right?

public/tee_client_api.h
divneil: I added to differentiate whose UUID it is. Currently, with the name it always seems it is of TEE. I will remove it.

tee-supplicant/CMakeLists.txt
divneil: I used `tee_` because similar data flow direction exists in other files like `teec_ta_load.c`  (TA binary -> optee_os). Please confirm if you want to change the names.

tee-supplicant/src/optee_msg_supplicant.h
divneil: Sure. Will move to `OPTEE_MSG_RPC_CMD_REE_SERVICE`

public/ree_service_api.h
divneil: > `ree_service_rcv()` and `ree_service_snd()` maybe (having service here is more useful than params IMO).

Sure.

> 1. A dynamic library is loaded by tee-supplicant and a callback is invoked. Code therefore runs in the context of the tee-supplicant RPC loop.
> 2. A message queue interface. Code executes in a different process.

Yes. The example code is in https://github.com/linaro-swg/optee_examples/pull/63. I will fork out a new directory in `optee_examples`. Currently, in ` hello_world/host/main.c`, there's a code in #if0 (line number 107). It enables Message Queue based params processing.
 
>think these function should be commented so it's clear who should use them and for which scenario.

Sure.

> What if someone needs to run the service in kernel space? (i.e., TA wants to talk to a Linux driver). That's a use case I've heard before.

Hmm, for now, I didn't had this requirement, but as I understand, it will involve patching `teepriv`. A convoluted way could be to write a `ree_service` and have that service talk to driver.

tee-supplicant/src/tee_service.c
divneil: My bad. Since, dynamic library loading finds the symbol, so, it doesn't complain. I will fix it.

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+	size_t buf_sz;
+};
+
+static bool is_param_type_value(uint64_t param_type)
+{
+	if (param_type == TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INPUT ||
+			param_type == TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_OUTPUT ||
+			param_type == TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INOUT)
+		return true;
+	return false;
```
etienne-lms: minor style: prefer a switch/case, here and in `is_param_type_memref()`.
```C
	switch (param_type) {
	case TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INPUT:
	case TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_OUTPUT:
	case TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INOUT:
		return true;
	default:
		return false;
	}
```


Also rename `is_param_type_{value|memref}()` into `param_type_is_{value|memref}()`.

Same comment for equivalent functions in `tee_service.c`.

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+	char filename[64];
+	key_t msgqkey = 0;
+	TEEC_Result result;
+	char uuid_str[48];
+
+	result = uuid_to_str(uuid, uuid_str, sizeof(uuid_str));
+	if (result != TEEC_SUCCESS)
+		return result;
+
+	struct service *s = malloc(sizeof(struct service));
```
etienne-lms: style: prefer declaration at function entry.

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+	if (!s)
+		return -ENOMEM;
+
+	/* Create a file in /data/<uuid> */
+	snprintf(filename, sizeof(filename), "/data/%s", uuid_str);
+	fp = fopen(filename, "w");
+	if (!fp) {
+		printf("Failed to create a file for token\n");
```
etienne-lms: prefer the optee log macros `EMSG()`  for error traces.

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+		goto err;
+	}
+
+	size = fwrite(uuid_str, 1, strlen(uuid_str), fp);
+	if (size != strlen(uuid_str)) {
+		printf("Failed to write to %s\n", filename);
+		result = TEEC_ERROR_GENERIC;
```
etienne-lms: should `fclose(fp);` here.

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+	if (!s)
+		return;
+
+	if (s->msgqid != -1) {
+		if (msgctl(s->msgqid, IPC_RMID, NULL) == -1)
+			printf("Failed to delete msgq, try using ipcrm\n");
+	}
+
+	if (s->buf)
+		free(s->buf);
```
etienne-lms: minor style: remove the test, calling `free(NULL);` is safe,

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+			ptr += attr_sz;
+
+			params[idx].u.memref.size = *(size_t *)ptr;
+			params[idx].u.memref.buffer = ptr +
+					sizeof(params[idx].u.memref.size);
+			ptr = (char *)params[idx].u.memref.buffer +
+					params[idx].u.memref.size;
+		}
+		idx++;
+		if (idx == 4)
```
etienne-lms: `TEE_PARAM_NUM`

libteec/src/ree_service_api.c
```diff
@@ -0,0 +1,255 @@
...
+TEEC_Result ree_rcv_params(void *service, size_t *num_params,
+					struct tee_params *params)
+{
+	int ret, idx = 0;
+	char *buf = NULL, *ptr;
+	struct service *s = service;
+	long msg_size[2] = {0};
+	size_t size, attr_sz, value_sz, mtype_sz = sizeof(long);
+
+	if (!s || !num_params || !params)
```
etienne-lms: add ` || (num_params > TEE_PARAM_NUM)`

libteec/src/ree_service_api.c
divneil: I have modified optee-examples and optee_os, and came here. I used `is` in the beginning to have the function name as question: "Is it a memref?"

etienne-lms: For consistency with other OP-TEE OS source, I suggest to use `foo_is_fine()` rather than `is_foo_find()`.

jenswi-linaro: It's just a static function so anything within reason goes. I think that `is_param_type_memref()` looks nice, but I don't have a strong opinion in this.

tee-supplicant/src/tee_service.c
```diff
@@ -0,0 +1,502 @@
...
+err:
+	return res;
+}
+
+/**
+ * tee_service_process() - called from tee-supplicant
+ * This functions finds the service for tee based on UUID
+ * based on either message queue or dynamic lib.
+ * o Message queue is useful when we want to directly pass
+ *   some date to CA.
```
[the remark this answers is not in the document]
divneil: Missed in this patch. I will update in next commit

PR #172, opened 2019-09-24
Remove obsolete toolchain variables reference in config.mk

AleksandrBulyshchenko:
ARM_TOOLCHAIN_DIR and ARM_GCC_PREFIX aren't used for a long time and only mentioned in config.mk now.
This could be confusing, thus it's better to remove them.


jforissier:

Thanks!

jenswi-linaro:

PR #175, opened 2019-10-30
Fix Open() permission is not enough in ARM.

leesagacious:
When executing xtest 2001 on arm, a test case appears failed,
when the owner of tee-supplicant is non-root,

Use the command chmod -R 777 /data/tee to fix,
Must be 777

### DIFF
--- a/tee-supplicant/src/tee_supp_fs.c
+++ b/tee-supplicant/src/tee_supp_fs.c
@@ -37,6 +37,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


jforissier:
No, `tee-supplicant` doesn't have to run as `root` and `/data/tee` doesn't have to be `0777`.
See https://github.com/OP-TEE/build/blob/master/br-ext/package/optee_client/optee_client.mk#L22:L31 and https://github.com/OP-TEE/build/blob/master/br-ext/package/optee_client/S30optee#L8:L52 for example.

PR #176, opened 2019-11-07
ftrace, gprof: fix format of UUID string in file name

jforissier:
Function prof_process() creates a file in /tmp which is named after
the UUID passed to the function. The format of the UUID is not the
usual one; a dash is missing. Add it.

Signed-off-by: Jerome Forissier <jerome@forissier.org>


jenswi-linaro:

PR #177, opened 2020-01-29
travis: run checkpach on patches

jenswi-linaro:
Adds checks with checkpatch on new patches.


jenswi-linaro:
Tag applied

PR #179, opened 2020-02-04
libckteec: invocation api function to register shared memory

etienne-lms:
Add new API function ckteec_register_shm() to register in OP-TEE OS
a piece of Linux userland memory which a PKCS11 TA service will access
to, as an input data or output data buffer in a cryptography ciphering
sequence.


etienne-lms:
tag applied.

PR #180, opened 2020-02-04
libckteec: assert api init/release functions return value compliance

etienne-lms:
For sanity of the library and service implementation, assert the
CKR_* return value before returning to caller from CK API functions.

This change asserts return values for C_Initialize() and C_Finalize().

### DIFF
--- a/libckteec/src/ck_helpers.c
+++ b/libckteec/src/ck_helpers.c
@@ -3,6 +3,11 @@
  * Copyright (c) 2020, Linaro Limited
  */
 
+#include <assert.h>
+#include <pkcs11.h>
+#include <stdio.h>
+#include <tee_client_api.h>
+
 #include "ck_helpers.h"
 
 CK_RV teec2ck_rv(TEEC_Result res)
@@ -20,3 +25,20 @@ CK_RV teec2ck_rv(TEEC_Result res)
 		return CKR_FUNCTION_FAILED;
 	}
 }
+
+#ifdef DEBUG
+void ckteec_assert_expected_rv(const char *function, CK_RV rv,
+			       const CK_RV *expected_rv, size_t expected_count)
+{
+	size_t n = 0;
+
+	for (n = 0; n < expected_count; n++)
+		if (rv == expected_rv[n])
+			return;
+
+	fprintf(stderr, "libckteec: unexpected return value 0x%lx for %s\n",
+		rv, function);
+
+	assert(0);
+}
+#endif
--- a/libckteec/src/ck_helpers.h
+++ b/libckteec/src/ck_helpers.h
@@ -9,6 +9,23 @@
 #include <pkcs11.h>
 #include <tee_client_api.h>
 
+#include "local_utils.h"
+
+#ifdef DEBUG
+#define ASSERT_CK_RV(_rv, ...)						\
+	do {								\
+		const CK_RV ref[] = { __VA_ARGS__ };			\
+		size_t count = ARRAY_SIZE(ref);				\
+									\
+		ckteec_assert_expected_rv(__func__, (_rv), ref, count);	\
+	} while (0)
+
+void ckteec_assert_expected_rv(const char *function, CK_RV rv,
+			       const CK_RV *expected_rv, size_t expected_count);
+#else
+#define ASSERT_CK_RV(_rv, ...)		(void)0
+#endif /*DEBUG*/
+
 /*
  * Convert IDs between PKCS11 TA and Cryptoki.
  */
--- a/libckteec/src/pkcs11_api.c
+++ b/libckteec/src/pkcs11_api.c
@@ -8,6 +8,7 @@
 #include <stddef.h>
 
 #include "invoke_ta.h"
+#include "ck_helpers.h"
 
 static const CK_FUNCTION_LIST libckteec_function_list = {
 	.version = {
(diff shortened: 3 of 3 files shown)


libckteec/src/pkcs11_api.c
```diff
@@ -26,18 +27,53 @@ static bool lib_initiated(void)
...
 {
+	CK_RV rv = 0;
+
 	/* Argument currently unused as per the PKCS#11 specification */
 	(void)pInitArgs;
 
-	return ckteec_invoke_init();
+	rv = ckteec_invoke_init();
+
+	switch (rv) {
```
jenswi-linaro: This seems to be a recurring pattern. How about using a helper function for this instead? I think it would be nice to have the offending return value printed too.

etienne-lms: good idea. thanks.

libckteec/src/ck_helpers.c
```diff
@@ -20,3 +25,25 @@ CK_RV teec2ck_rv(TEEC_Result res)
...
+
+	for (n = 0; n < expected_rv_count; n++)
+		if (rv == expected_rv[n])
+			break;
+
+	expected_cryptoki_return_value = (n < expected_rv_count);
+
+	if (!expected_cryptoki_return_value)
+		fprintf(stderr, "libckteec: unexpected retrun value 0x%lx for %s\n",
+			rv, function);
```
etienne-lms: Together with the assert below, this build a trace like:
```
libckteec: unexpected retrun value 0xffffffff for C_Initialize
xtest: src/ck_helpers.c:48: ckteec_assert_expected_ckr: Assertion `expected_cryptoki_return_value' failed.
Aborted
```

libckteec/src/ck_helpers.c
```diff
@@ -20,3 +25,25 @@ CK_RV teec2ck_rv(TEEC_Result res)
...
+void ckteec_assert_expected_ckr(CK_RV rv, const CK_RV *expected_rv,
+				size_t expected_rv_count,
+				const char *function)
+{
+	size_t n = 0;
+	bool expected_cryptoki_return_value = false;
+
+	for (n = 0; n < expected_rv_count; n++)
+		if (rv == expected_rv[n])
+			break;
```
jenswi-linaro: How about returning here instead? That way you can get rid of all tests below.

libckteec/src/ck_helpers.c
```diff
@@ -20,3 +25,25 @@ CK_RV teec2ck_rv(TEEC_Result res)
...
+
+	expected_cryptoki_return_value = (n < expected_rv_count);
+
+	if (!expected_cryptoki_return_value)
+		fprintf(stderr, "libckteec: unexpected retrun value 0x%lx for %s\n",
```
jenswi-linaro: Stopped using `EMSG()`?
`s/retrun/return/`

libckteec/src/pkcs11_api.c
```diff
@@ -26,18 +28,40 @@ static bool lib_initiated(void)
...
+	};
+	CK_RV rv = 0;
+
 	/* Argument currently unused as per the PKCS#11 specification */
 	(void)pReserved;
 
-	return ckteec_invoke_terminate();
+	rv = ckteec_invoke_terminate();
+
+	CKTEEC_ASSERT_EXPECTED_CRV(rv, expected_rv);
```
jenswi-linaro: This macro could be a varargs macro instead, which would construct the array passed to `ckteec_assert_expected_ckr()` internally instead. Since we'll have this in all the coming C_ functions it would be nice to have it as streamlined as possible.

libckteec/src/ck_helpers.c
etienne-lms: right.
I though also about forcinf `CKR_OK` at first test, likely the most used.
But I guess with `DEBUG` we don't expect any performances.

libckteec/src/ck_helpers.c
etienne-lms: `EMSG()` mandates hacky `BINARY_PREFIX` at source file top. I don't want to spread on all the 
Since this trace is companion to the `assert()` trace below, I think it's fair using `stderr` straight.

> s/retrun/return/
thanks. sorry.

libckteec/src/pkcs11_api.c
etienne-lms: ok :)

etienne-lms:
bad update, will push again...

etienne-lms:
i squashed the fixup commits into a new series since content fully changed.

libckteec/src/ck_helpers.h
```diff
@@ -7,8 +7,28 @@
 #define LIBCKTEEC_CK_HELPERS_H
 
 #include <pkcs11.h>
+#include <stdio.h>
 #include <tee_client_api.h>
 
+#include "local_utils.h"
+
+#define ASSERT_CK_RV(_rv, _count, ...)	\
+	ckteec_assert_expected_rv(__func__, (_rv), (_count), \
```
jenswi-linaro: How about something like:
``` C
#define ASSERT_CK_RV(_rv, ...) do { \
    const CK_RV _rvs[] = { ## __VA_ARGS__ }; \
    ckteec_assert_expected_rv(__func__, rv, rvs, ARRAY_SIZE(rvs)); \
} while (0)

Comment (etienne-lms) [reply libckteec/src/ck_helpers.h]: smart

File: libckteec/src/ck_helpers.c [hunk 7]
```diff
@@ -20,3 +25,20 @@ CK_RV teec2ck_rv(TEEC_Result res)
...
+	size_t n = 0;
+
+	for (n = 0; n < expected_count; n++)
+		if (rv == expected_rv[n])
+			return;
+
+	fprintf(stderr, "libckteec: unexpected return value 0x%lx for %s\n",
+		rv, function);
+
+	assert(n < expected_count);
```
jenswi-linaro: assert(0)?

etienne-lms: not strong opinion. `assert(0);` looks more explicit. I'll update.

etienne-lms:
tag applied. thanks.

PR #185, opened 2020-02-10
ci: travis: Use proper construct for full clone

d3zd3z:
As per the Travis docs at
https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth to
avoid performing git clones with limited depth, the depth should be set
to "false", not to a large magic number.

Signed-off-by: David Brown <david.brown@linaro.org>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ notifications:
   - email: true
 
 git:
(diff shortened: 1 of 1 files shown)


jenswi-linaro:

PR #187, opened 2020-02-11
libckteec: use PRIu32 instead of casts

etienne-lms:
Use PRIu32 specifiers in debug trace instead of casting the
traced values.


jenswi-linaro:

etienne-lms:
Tag applied.
IBART fails, but not seems related to this P-R.

PR #188, opened 2020-02-11
libckteec: handle pkcs11 return value from TA

etienne-lms:


etienne-lms:
Checkpatch warns on macro `TA2CK()` but implementation is fine.
```
CHECK: Macro argument reuse 'conv' - possible side-effects?
#64: FILE: libckteec/src/ck_helpers.c:49:
+#define TA2CK(out, in, conv)		ta2ck(out, in, conv, ARRAY_SIZE(conv))
total: 0 errors, 0 warnings, 1 checks, 148 lines checked
```

libckteec/src/ck_helpers.c
```diff
@@ -5,25 +5,124 @@
...
 #include <stdio.h>
+#include <string.h>
 #include <tee_client_api.h>
 
 #include "ck_helpers.h"
+#include "local_utils.h"
 
-CK_RV teec2ck_rv(TEEC_Result res)
+/*
+ * Helpers for conversion of CK IDs to/from PKCS11 TA IDs: tables of identifiers
```
jenswi-linaro: Why do we need to convert anything? Can't we use the correct value from the beginning?

etienne-lms: The idea was to use private IDs in the TA so that TA internals can evolve independently from the Client interface that is Cryptoki API, hence a need for conversion.

Currently I use the TA private IDs in the TA API and Client lib has to do the conversion.

An alternate is to have TA API using the Client IDs and do the conversion from/to its internal IDs.

Another alternative could be to use PKCS#11 IDs in the TA. No conversion, but would make TA internals even more tied to the Client interface.

jenswi-linaro: I'd prefer PKCS#11 IDs, then it's tied to the specification which should be OK for a PKCS#11 TA.

etienne-lms: https://github.com/OP-TEE/optee_os/pull/3582 also updated accordingly.

jenswi-linaro:
Please squash the commits to make it easier to review.

etienne-lms:
fixup commits squashed and rebased.
Deprecated commit "libckteec: convert return code from PKCS11 TA" removed.

etienne-lms:
tag applied

PR #191, opened 2020-03-04
libckteec: better handle null size buffer in C_GetSlotList()

etienne-lms:
When caller of `C_GetSlotList()` provides a non-NULL buffer reference with a count value of zero, the TA gets invoked with a null buffer reference in the OP-TEE shared memory, hence it will return `CKR_OK` status since it legitimately only returns the count information.

This change makes ckteec library to return `CKR_BUFFER_TOO_SMALL` as per PKCS#11 specification.

This change also fixes `C_GetSlotList()` that, prior to this change, would have returned `CKR_ARGUMENTS_BAD` upon `NULL` buffer and non-zero count value whereas specification says:
> If pSlotList is NULL_PTR, (...). The contents of *pulCount on entry to C_GetSlotList has no meaning in this case (...).

Found the issue from https://github.com/OP-TEE/optee_os/pull/3644#discussion_r387592239.

### DIFF
--- a/libckteec/src/pkcs11_token.c
+++ b/libckteec/src/pkcs11_token.c
@@ -65,15 +65,23 @@ CK_RV ck_slot_get_list(CK_BBOOL present,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	TEEC_SharedMemory *shm = NULL;
 	uint32_t *slot_ids = NULL;
+	size_t client_count = 0;
 	size_t size = 0;
 
 	/* Discard @present: all slots reported by TA are present */
 	(void)present;
 
-	if (!count || (*count && !slots))
+	if (!count)
 		return CKR_ARGUMENTS_BAD;
 
-	size = *count * sizeof(*slot_ids);
+	/*
+	 * As per spec, if @slots is NULL, "The contents of *pulCount on
+	 * entry to C_GetSlotList has no meaning in this case (...)"
+	 */
+	if (slots)
+		client_count = *count;
+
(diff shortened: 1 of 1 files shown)


etienne-lms:
xtest update for the specific case through https://github.com/OP-TEE/optee_test/pull/400.

libckteec/src/pkcs11_token.c
```diff
@@ -83,6 +92,19 @@ CK_RV ck_slot_get_list(CK_BBOOL present,
...
 	if (rv == CKR_OK || rv == CKR_BUFFER_TOO_SMALL) {
+		/*
+		 * When @slot != NULL and @count[0] == 0, library shall
+		 * return CKR_BUFFER_TOO_SMALL if provided buffer is too
+		 * small whereas TA would have returned CKR_OK since
+		 * allocating a zero sized shm instance results in a NULL
+		 * shm buffer reference.
+		 */
+		if (size && slots && !client_count) {
+			assert(rv == CKR_OK);
```
jenswi-linaro: I'm not convinced that we should have this assert().

etienne-lms: True, I can remove.
(will do)

jenswi-linaro:

etienne-lms:
Tag applied.
Failure in IBART seems not related to these changes:
`23805:  error: RPC failed; curl 56 GnuTLS recv error (-110): The TLS connection was non-properly terminated.`

PR #193, opened 2020-03-10
libckteec: get mechanism list/info

etienne-lms:
Related to https://github.com/OP-TEE/optee_os/pull/3667 and https://github.com/OP-TEE/optee_test/pull/403.


libckteec/src/pkcs11_token.c
```diff
@@ -270,3 +270,125 @@ CK_RV ck_token_get_info(CK_SLOT_ID slot, CK_TOKEN_INFO_PTR info)
...
+		goto bail;
+	}
+	if (rv)
+		goto bail;
+
+	mecha_ids = out->buffer;
+	for (n = 0; n < *count; n++)
+		mechanisms[n] = mecha_ids[n];
+
+bail:
```
jenswi-linaro: I'd prefer `out:` since we may use this label when returning OK too.

etienne-lms: ok

etienne-lms:
Fixup commits squashed, tag applied, P-R reference fixed.

Checkpatch taints on BIT() non-use and missing fallthru already faced in this track.
```
160cfb6 libckteec: sync with pkcs11 ta api for mechanism info
CHECK: Prefer using the BIT macro
#40: FILE: libckteec/include/pkcs11_ta.h:298:
+#define PKCS11_CKFM_HW				(1U << 0)

(...)

WARNING: Possible switch case/default not preceded by break or fallthrough comment
#131: FILE: libckteec/src/ck_debug.c:208:
+	default:

(...)
```

PR #195, opened 2020-03-26
libteec: Update TEEC_OpenSession() to pass connection data based on method

vesajaaskelainen:
TEE Client API specification states that only with login methods
TEEC_LOGIN_GROUP and TEEC_LOGIN_GROUP_APPLICATION connection data is
non-NULL.

In current Linux kernel API interfece it is not possible to handle custom
connection data so every login method needs to be manually passed thru
clnt_uuid field and handled specially both in user space and in kernel to
allow kernel to do proper authorization checking when opening an session.

This is related to: https://github.com/OP-TEE/optee_os/issues/3642

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -502,6 +502,59 @@ static void uuid_to_octets(uint8_t d[TEE_IOCTL_UUID_LEN], const TEEC_UUID *s)
 	memcpy(d + 8, s->clockSeqAndNode, sizeof(s->clockSeqAndNode));
 }
 
+static void setup_client_data(struct tee_ioctl_open_session_arg *arg,
+			      uint32_t connection_method,
+			      const void *connection_data)
+{
+	arg->clnt_login = connection_method;
+
+	switch (connection_method) {
+	case TEE_IOCTL_LOGIN_PUBLIC:
+		/* No connection data to pass */
+		break;
+	case TEE_IOCTL_LOGIN_USER:
+		/* Kernel auto-fills UID and forms client UUID */
+		break;
+	case TEE_IOCTL_LOGIN_GROUP:
+		/*
+		 * Connection data for group login is uint32_t and rest of
+		 * clnt_uuid is set as zero.
+		 *
+		 * Kernel verifies group membership and then forms client UUID.
+		 */
+		memcpy(arg->clnt_uuid, connection_data, sizeof(gid_t));
+		break;
+	case TEE_IOCTL_LOGIN_APPLICATION:
+		/*
+		 * Kernel auto-fills application identifier and forms client
+		 * UUID.
+		 */
+		break;
+	case TEE_IOCTL_LOGIN_USER_APPLICATION:
+		/*
+		 * Kernel auto-fills application identifier, UID and forms
+		 * client UUID.
+		 */
+		break;
+	case TEE_IOCTL_LOGIN_GROUP_APPLICATION:
+		/*
+		 * Connection data for group login is uint32_t rest of
+		 * clnt_uuid is set as zero.
+		 *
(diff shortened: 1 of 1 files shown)


libteec/src/tee_client_api.c
```diff
@@ -533,6 +531,52 @@ TEEC_Result TEEC_OpenSession(TEEC_Context *ctx, TEEC_Session *session,
 	uuid_to_octets(arg->uuid, destination);
 	arg->clnt_login = connection_method;
 
+	switch (connection_method) {
```
jenswi-linaro: The documentation is nice, but it makes you lose the overview of this function.
Please move this into a helper function.

jenswi-linaro:
> In current Linux kernel API interface it is not possible to handle custom
connection data so every login method needs to be manually passed thru
clnt_uuid field and handled specially both in user space and in kernel to
allow kernel to do proper authorization checking when opening an session.

I'd argue that it is possible since this is what you're doing in this patch. :-)
It's even done automatically by `TEEC_OpenSession()` with this.

libteec/src/tee_client_api.c
vesajaaskelainen: Ok.

vesajaaskelainen:
> > In current Linux kernel API interface it is not possible to handle custom
> > connection data so every login method needs to be manually passed thru
> > clnt_uuid field and handled specially both in user space and in kernel to
> > allow kernel to do proper authorization checking when opening an session.
> 
> I'd argue that it is possible since this is what you're doing in this patch. :-)
> It's even done automatically by `TEEC_OpenSession()` with this.

I would say it is hackish. It does not allow login extensions that could provide free string as input.

jenswi-linaro:
Whatever is passed needs to fit into a `TEE_Identity` when retrieved by the TA.

libteec/src/tee_client_api.c
vesajaaskelainen: moved and updated the PR

vesajaaskelainen:
Just rebased this PR to latest master.

jforissier:
Looks like this PR is mostly OK, @jenswi-linaro are you requesting some editing of the commit message? Speaking of which, there is a typo: `s/interfece/interface`.

jenswi-linaro:
I'd like to have the commit message to describe what's added. The last paragraph doesn't add anything useful.

vesajaaskelainen:
> I'd like to have the commit message to describe what's added. The last paragraph doesn't add anything useful.

Ok. Perhaps:
```
    libteec: Update TEEC_OpenSession() to pass connection data based on method
    
    Adds handling and descriptions for login methods TEE_IOCTL_LOGIN_PUBLIC,
    TEE_IOCTL_LOGIN_USER, TEE_IOCTL_LOGIN_GROUP, TEE_IOCTL_LOGIN_APPLICATION,
    TEE_IOCTL_LOGIN_USER_APPLICATION and TEE_IOCTL_LOGIN_GROUP_APPLICATION.
    
    TEE Client API specification states that only with login methods
    TEEC_LOGIN_GROUP and TEEC_LOGIN_GROUP_APPLICATION connection data is
    non-NULL and in these cases connection data is group identifier (gid_t).
```

jenswi-linaro:
Sounds good, with that:

vesajaaskelainen:
> Sounds good, with that:
> `Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>`

Thanks for the review!

Updated the PR with those.

jforissier:
Thanks @vesajaaskelainen.

I don't know what's wrong with checkpatch (Travis CI), it seems to be confused with the non-ASCII characters in your name perhaps. I checked locally, all is fine, and I could run our `checkpatch.sh` with no error.

PR #196, opened 2020-03-27
libckteec: fix TA API for token flags and user ID

etienne-lms:


etienne-lms:
fix commit message header line for 1st commit: s/ta: pkcs11:/libckteec:/.

etienne-lms:
tag applied

PR #198, opened 2020-03-30
tee-supplicant: cast to __u64 to avoid potential overflow

jforissier:
If request->send.num_params is assigned a large value, there could be
an overflow when computing data.buf_len. Cast to __u64 to avoid a
32-bit computation.

Signed-off-by: Jerome Forissier <jerome@forissier.org>


jenswi-linaro:

Issue #178, opened 2020-01-31
#138 seems to have broken zynqmp


veriktig:
[RFC] libsks for Pkcs#11 services through OP-TEE SKS TA #138

Build now fails when linking libckteec with: "cannot find -lteec"

Workaround:
diff --git a/zynqmp/optee-client/optee-client.bb b/zynqmp/optee-client/optee-client.bb
index ee8dbea..e1c7bbb 100644
--- a/zynqmp/optee-client/optee-client.bb
+++ b/zynqmp/optee-client/optee-client.bb
@@ -1,5 +1,6 @@
 OPTEE_VERSION ??= "latest"
-SRCREV ??= "${AUTOREV}"
+# SRCREV ??= "${AUTOREV}"
+SRCREV = "82ec680ece99e7b345765fe680bd1c143b993da0"
 BRANCH ??= "master"
 
 DESCRIPTION = "OP-TEE Client"

jenswi-linaro:
From the error description I can only assume that `-L<path to libteec>` somehow isn't passed to the linker when linking libckteec. If it's an error in a Makefile or CMakeLists.txt I can't tell.

Emantor:
@veriktig does your client recipe use cmake or the plain makefiles?

PR #199, opened 2020-03-30
libckteec/libteec: move LFLAGS to end of call

Emantor:
At least Yocto and ptxdist provide the option to invoke the linker with
-Wl,--as-needed, which will only mark the libraries as needed if they
are used by the binary or library. However this assumes that the object
files are passed before the shared libraries, move the LFLAGS for
libctkteec and libteec to the end of linker invocation to support this.

Signed-off-by: Rouven Czerwinski <rouven@czerwinskis.de>

### DIFF
--- a/libteec/Makefile
+++ b/libteec/Makefile
@@ -48,7 +48,7 @@ libteec: $(TEEC_LIBRARY) $(OUT_DIR)/libteec.a
 
 $(TEEC_LIBRARY): $(TEEC_OBJS)
 	@echo "  LINK    $@"
-	$(VPREFIX)$(CC) -shared -Wl,-soname,$(LIB_MAJOR) $(TEEC_LFLAGS) -o $@ $+
+	$(VPREFIX)$(CC) -shared -Wl,-soname,$(LIB_MAJOR) -o $@ $+ $(TEEC_LFLAGS)
 	@echo ""
 
 $(OUT_DIR)/libteec.a: $(TEEC_OBJS)
--- a/libckteec/Makefile
+++ b/libckteec/Makefile
@@ -47,7 +47,7 @@ libckteec: $(OUT_DIR)/$(LIBCKTEEC_SO_LIBRARY)
 
 $(OUT_DIR)/$(LIBCKTEEC_SO_LIBRARY): $(LIBCKTEEC_OBJS)
 	@echo "  LINK    $@"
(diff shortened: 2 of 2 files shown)


libteec/Makefile
```diff
@@ -48,7 +48,7 @@ libteec: $(TEEC_LIBRARY) $(OUT_DIR)/libteec.a
 
 $(TEEC_LIBRARY): $(TEEC_OBJS)
 	@echo "  LINK    $@"
-	$(VPREFIX)$(CC) -shared -Wl,-soname,$(LIB_MAJOR) $(TEEC_LFLAGS) -o $@ $+
+	$(VPREFIX)$(CC) -shared -Wl,-soname,$(LIB_MAJOR) -o $@ $+ $(TEEC_LFLAGS)
```
jforissier: I don't understand why this changes anything wrt. `-Wl,--as-needed`. As long as the flag is given before the shared libraries that should be pulled as needed, the result should be the same, no?

Emantor: At least during my testing the library is linked correctly, but is not dynamically linked to the required library. This results i.e. in libckteec not being linked to libteec, which results in an error on execution (since the TEEC_* symbols can't be resolved). 
I'm still searching for the proper LD documentation, however all I can currently come up with are stackoverflow posts and these mention that the objects need to be before the shared library flags to properly detect that the shared library is needed, [1](https://stackoverflow.com/questions/7826448/linking-libraries-with-gcc-order-of-arguments), [2](https://stackoverflow.com/questions/8207464/linking-math-library-in-gcc-4-6-1-ubuntu-11-10).

Emantor: The [llvm-ld](https://linux.die.net/man/1/llvm-ld) man page mentions the same under library linkage.

jenswi-linaro: Sometimes we use `--start-group` and `--end-group` to work around this. In this case it's clear that the libraries should be given after the object files.

jforissier: Got it now. Indeed objects should always be before libraries. Now I understand why adding `-Wl,--as-needed` triggers the issue -- without it the DT_NEEDED dependencies are always added.

Emantor:
Tag applied.

Emantor:
This should also fix #178, added as Fixes to the commit.

jforissier:
> This should also fix #178, added as Fixes to the commit.

@veriktig can you confirm?

Emantor:
On second look at #178 this looks more like a problem with `-L…` and his underlying build system. I removed the fixes flag.

jforissier:
> On second look at #178 this looks more like a problem with `-L…` and his underlying build system. I removed the fixes flag.

Agreed :wink:

PR #200, opened 2020-03-31
libteec: fix null pointer dereference

HuKeping:
The context of the given session which used to close a session
or invoke command may be NULL.
    
Typically, we would not call TEEC_InvokeCommand or TEEC_CloseSession
if TEEC_Opensession or TEEC_InitializeContext failed, but we can not
stop people from doing that either.
    
Actually we failed some test in our product which keep calling
`invokeCommand` even if `Opensession` has failed, and it finally turns
out a hang task because the `session->ctx` is NULL.
    
Signed-off-by: Xiao Wang <xiaowang3@huawei.com>
Tested-by: Xiao Wang <xiaowang3@huawei.com> (Hisilicon)
Signed-off-by: Hu Keping <hukeping@huawei.com>

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -568,7 +568,7 @@ void TEEC_CloseSession(TEEC_Session *session)
 
 	memset(&arg, 0, sizeof(arg));
 
-	if (!session)
+	if (!session || !session->ctx)
 		return;
 
 	arg.session = session->session_id;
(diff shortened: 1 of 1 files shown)


jforissier:
Hi @HuKeping,

I understand why you want to do this, but as discussed in the past (here: https://github.com/OP-TEE/optee_client/pull/60) we don't like it :wink:

HuKeping:
Okay, i'll take it as "Programmer Error"  if  passing invalid parameters . 

Thanks @jforissier

PR #202, opened 2020-04-07
libckteec: fixup changes from Ricardo work

etienne-lms:
A series of fixes from @ricardosalveti back ported  https://github.com/ricardosalveti/optee-sks.


etienne-lms:
After a second thought, URLs provided as reference may not be permanent links (@ricardosalveti ?). Maybe better remove them.

jforissier:
@etienne-lms the links are not really useful indeed. Ricardo is the author and first S-o-b, and you have added your own S-o-b, so I'd say all is fine.
On minor thing: regarding the `[backport]` comment in the tags, I would suggest not to add anything if the patch applies cleanly (just keep you S-o-b to record that you carried the patch upstream). You should add a comment only if you need to change the patch in some way for the patch to apply cleanly, in this case you would briefly describe the change.

etienne-lms:
None of the patches did apply straight. I needed to adapt them in some way or another, at least to update file paths that do not match. I mainly wanted the original author with related S-o-b to be set and added my S-o-b since my updates.

Ok if I remove the `Link:` and `[backport]` lines?

jforissier:
> None of the patches did apply straight. I needed to adapt them in some way or another, at least to update file paths that do not match. I mainly wanted the original author with related S-o-b to be set and added my S-o-b since my updates.

OK. That's good.

> Ok if I remove the `Link:`

For me, yes

> and `[backport]` lines?

For me again, yes although I would prefer any change to be briefly mentioned, it is more correct with respect to the original author. Example in the linux kernel: https://github.com/linaro-swg/linux/commit/5ee858975b13a9b40db00f456989a689fdbb296c

etienne-lms:
Ok, i'll put some very brief description.

ricardosalveti:
I'm happy with link/backport removed as well, no worries from my side.

Thanks a lot for doing this work, hope to follow with additional changes soon!

jenswi-linaro:

etienne-lms:
Tag applied. 
Thanks for the feedback. Tell me if the commit messages are not fine.

PR #203, opened 2020-04-08
libteec: report out of bound memory references from client library

etienne-lms:
Change teec_pre_process_partial() to reject out of bound memref
invocation parameters before TEE is invoked.

This change is needed in case dynamic SHM is disabled in which case
memory reference used as TEE invocation parameters may have been
allocated using a wider buffer than the one provided by client. In
such case, TEE may not able to detect out of bound references
since original client buffer very location (start and size) do not
reach the TEE.

Reported-by: Sumit Garg <sumit.garg@linaro.org>
Suggested-by: Jens Wiklander <jens.wiklander@linaro.org>

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -271,6 +271,10 @@ static TEEC_Result teec_pre_process_partial(uint32_t param_type,
 	if ((shm->flags & req_shm_flags) != req_shm_flags)
 		return TEEC_ERROR_BAD_PARAMETERS;
 
+	if ((memref->offset + memref->size < memref->offset) ||
+	    (memref->offset + memref->size > shm->size))
+		return TEEC_ERROR_BAD_PARAMETERS;
+
 	/*
 	 * We're using a shadow buffer in this reference, copy the real buffer
 	 * into the shadow buffer if needed. We'll copy it back once we've


etienne-lms:
Updated with missing ref to https://github.com/OP-TEE/optee_test/issues/410 in commit message.

jforissier:

libteec/src/tee_client_api.c
```diff
@@ -271,6 +271,9 @@ static TEEC_Result teec_pre_process_partial(uint32_t param_type,
 	if ((shm->flags & req_shm_flags) != req_shm_flags)
 		return TEEC_ERROR_BAD_PARAMETERS;
 
+	if (memref->offset + memref->size > shm->size)
```
jenswi-linaro, line 274: Check for overflow?

etienne-lms: good idea

etienne-lms:
updated

jenswi-linaro:

etienne-lms:
squashed. tags applied.

jforissier:
Why does `xtest 1014` fails with IBART?

b49020:
> Why does `xtest 1014` fails with IBART?

It seems we also need to extend return origin for `xtest 1014` [here](https://github.com/OP-TEE/optee_test/blob/master/host/xtest/sdp_basic.c#L573) to accept `TEEC_ORIGIN_API` too for invalid memory references.

etienne-lms:
@b49020, agree.
I should have run the full xtest before submit this.
Will do. sorry...

jforissier:
@etienne-lms I have a patch ready now and can create a PR in xtest, you don't need to do it ;-)

etienne-lms:
ok, thanks.

b49020:
`Reviewed-by: Sumit Garg <sumit.garg@linaro.org>`

etienne-lms:
@b49020 tag applied

PR #204, opened 2020-04-14
libteec: registration shared memory for a null buffer is invalid

etienne-lms:
Fix libteec to return TEEC_ERROR_BAD_PARAMETERS when client registers
a shared memory reference for a NULL buffer as per GPD specification
v1.0 that states "The buffer field MUST point to the memory region to
be shared, and MUST not be NULL.

Prior this change TEEC_RegisterSharedMemory() called with a NULL buffer
reference returned TEEC_ERROR_OUT_OF_MEMORY when dynamic shared memory
is enabled and successfully registered a memory reference when dynamic
shared memory is disabled. With this change the API function returns
TEEC_ERROR_BAD_PARAMETERS.


jenswi-linaro:
Would it make sense to mention any commit in a `Fixes:` tag?
Anyway:

etienne-lms:
This change does not fix commit 20b567068a37 ("libutee: flag NULL pointer using invalid shm id"). Prior this commit, `TEEC_RegisterSharedMemory()` already had the behavior described in commit message of current change.

etienne-lms:
tag applied

PR #207, opened 2020-04-17
Sync tee.h with Linux kernel v5.6

jenswi-linaro:
It seems that tee.h never was synced with the kernel after the patches finally landed upstream.

The union in struct tee_ioctl_param is now removed. Some parts of the code seems to become a bit harder to read. I'm not sure what's the best way of addressing that. Any idea?

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -88,6 +88,12 @@ struct thread_arg {
 	pthread_mutex_t mutex;
 };
 
+struct param_value {
+	uint64_t a;
+	uint64_t b;
+	uint64_t c;
+};
+
 static pthread_mutex_t shm_mutex = PTHREAD_MUTEX_INITIALIZER;
 static struct tee_shm *shm_head;
 
@@ -122,7 +128,7 @@ static size_t num_waiters_dec(struct thread_arg *arg)
 }
 
 static int get_value(size_t num_params, struct tee_ioctl_param *params,
-		     const uint32_t idx, struct tee_ioctl_param_value **value)
+		     const uint32_t idx, struct param_value **value)
 {
 	if (idx >= num_params)
 		return -1;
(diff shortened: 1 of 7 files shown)


jforissier:
> It seems that tee.h never was synced with the kernel after the patches finally landed upstream.
> 
> The union in struct tee_ioctl_param is now removed. Some parts of the code seems to become a bit harder to read. I'm not sure what's the best way of addressing that. Any idea?

Use accessor macros for memrefs perhaps?

```C
#define MEMREF_SHM_ID(p) ((p)->c)
#define MEMREF_SHM_OFFS(p) ((p)->a)
#define MEMREF_SIZE(p) ((p)->b)
```

Then `bufsize = params[2].u.memref.size;` would become `bufsize = MEMREF_SIZE(params + 2)` [or `bufsize = MEMREF_SIZE(&params[2])`] instead of `bufsize = params[2].b;`.

jenswi-linaro:
That's a good idea. I'll update with that.

jenswi-linaro:
Update

tee-supplicant/src/tee_supplicant.c
```diff
@@ -743,19 +753,18 @@ void *tee_supp_param_to_va(struct tee_ioctl_param *param)
...
 
-	end_offs = param->u.memref.size + param->u.memref.shm_offs;
-	if (end_offs < param->u.memref.size ||
-	    end_offs < param->u.memref.shm_offs)
+	end_offs = MEMREF_SIZE(param) + MEMREF_SHM_OFFS(param);
+	if (end_offs < MEMREF_SIZE(param) || end_offs < MEMREF_SHM_OFFS(param))
 		return NULL;
 
-	tshm = find_tshm(param->u.memref.shm_id);
+	tshm = find_tshm(param->c);
```
jforissier, line 760: `MEMREF_SHM_ID`

jenswi-linaro:
Comment addressed, squashed and tag applied

PR #208, opened 2020-04-22
libckteec: fix in C_GetMechanismList and prevent segfault in mecha/slot info countings

etienne-lms:


libckteec/src/pkcs11_token.c
```diff
@@ -281,7 +281,12 @@ CK_RV ck_token_mechanism_ids(CK_SLOT_ID slot,
 	if (!count || (*count && !mechanisms))
```
jenswi-linaro, line 281: I suppose we should remove ` || (*count && !mechanisms)` also.

jenswi-linaro:
For "￼libckteec: prevent segfault in C_GetSlotList()", there's a stray "fixup" at the end of the commit message.

libckteec/src/pkcs11_token.c
etienne-lms: you're right. I will remove it.

etienne-lms:
updated (commit log not fixed yet)

jenswi-linaro:

etienne-lms:
fixup commit squashed.
commit log message fixed.
tag applied.
thanks

PR #211, opened 2020-05-31
flags: do not override CFLAGS from host

zandrey:
Some build systems set CFLAGS in environment before the build, but they
got overridden by CFLAGS immediate assignment operator.

Replace immediate assignment with append directive, so the former
definition of CFLAGS gets properly expanded.

Signed-off-by: Andrey Zhizhikin <andrey.z@gmail.com>

### DIFF
--- a/flags.mk
+++ b/flags.mk
@@ -6,7 +6,7 @@ CROSS_COMPILE   ?= arm-linux-gnueabihf-
 CC              ?= $(CROSS_COMPILE)gcc
 AR		?= $(CROSS_COMPILE)ar
 
-CFLAGS          := -Wall -Wbad-function-cast -Wcast-align \
+override CFLAGS += -Wall -Wbad-function-cast -Wcast-align \
 		   -Werror-implicit-function-declaration -Wextra \
 		   -Wfloat-equal -Wformat-nonliteral -Wformat-security \
 		   -Wformat=2 -Winit-self -Wmissing-declarations \
@@ -16,13 +16,13 @@ CFLAGS          := -Wall -Wbad-function-cast -Wcast-align \
 		   -Wswitch-default -Wunsafe-loop-optimizations \
 		   -Wwrite-strings -D_FILE_OFFSET_BITS=64
 ifeq ($(CFG_WERROR),y)
-CFLAGS		+= -Werror
+override CFLAGS	+= -Werror
 endif
-CFLAGS          += -c -fPIC
+override CFLAGS += -c -fPIC
 
 DEBUG       ?= 0
 ifeq ($(DEBUG), 1)
-CFLAGS          += -DDEBUG -O0 -g
+override CFLAGS += -DDEBUG -O0 -g
 endif
 
 RM              := rm -f


flags.mk
```diff
@@ -6,7 +6,7 @@ CROSS_COMPILE   ?= arm-linux-gnueabihf-
 CC              ?= $(CROSS_COMPILE)gcc
 AR		?= $(CROSS_COMPILE)ar
 
-CFLAGS          := -Wall -Wbad-function-cast -Wcast-align \
+CFLAGS          += -Wall -Wbad-function-cast -Wcast-align \
```
jbech-linaro, line 9: Shouldn't it be `?=` instead? Or at least the default with `?=`, then add the rest with `+=`?

zandrey: @jbech-linaro According to GNU Make manual - either way should work, as when `CFLAGS` is undefined, then append operator acts as immediate assignment.

In case if command line or build environment does not provide `CFLAGS` - then only those defined in *OP-TEE* are going to be set, which resembles the previous logic.

zandrey: Sorry, forgot the reference link, section 3.7 of [GNU Make manual](https://www.gnu.org/software/make/manual/make.html) references above behavior.

-- andrey

jforissier: > According to GNU Make manual - either way should work, as when `CFLAGS` is undefined, then append operator acts as immediate assignment.

No, citing the GNU manual:

"For the append operator ‘+=’, the right-hand side is considered immediate if the variable was previously set as a simple variable (‘:=’ or ‘::=’), and deferred otherwise."

So when the variable is not set initially, this will create a deferred assignment. Which is not an issue as far as I can tell.

However, have you tried this change? Does it work? What is the final value of CFLAGS when: (1) `CFLAGS=foo make` and (2) `make CFLAGS=foo`?
I'm not sure about the first case (not tried) but I'm almost certain the second case will result in CFLAGS being set to `foo` with all the added flags ignored.

zandrey: @jforissier Thanks a lot for reviewing the patch and your feedback!

> 
> 
> > According to GNU Make manual - either way should work, as when `CFLAGS` is undefined, then append operator acts as immediate assignment.
> 
> No, citing the GNU manual:
> 
> "For the append operator ‘+=’, the right-hand side is considered immediate if the variable was previously set as a simple variable (‘:=’ or ‘::=’), and deferred otherwise."
> 
> So when the variable is not set initially, this will create a deferred assignment. Which is not an issue as far as I can tell.

This is exactly what I meant with my reply above. Deferring assignment should not cause any issue here, even if `CFLAGS` are not set from outside.

> 
> However, have you tried this change? Does it work? What is the final value of CFLAGS when: (1) `CFLAGS=foo make` and (2) `make CFLAGS=foo`?
> I'm not sure about the first case (not tried) but I'm almost certain the second case will result in CFLAGS being set to `foo` with all the added flags ignored.

(1) resolves proper with OP-TEE flags appended to the `foo`
(2) indeed drops all flags but `foo`.

I've also tried out the original implementation, and in case of (2) internal flags are also dropped out, leaving only `foo` set. (1) does however discard `foo` and proceeds only with internal flags set.

I haven't managed yet to solve (2), but this assignment has to be changes IMHO since it does not work as expected anyway.

If you have some suggestions here on how to solve it - they would be really appreciated, since I'm a bit stuck with it.

Thanks again!

-- andrey

jforissier: > I haven't managed yet to solve (2), but this assignment has to be changes IMHO since it does not work as expected anyway.
> 
> If you have some suggestions here on how to solve it - they would be really appreciated, since I'm a bit stuck with it.

Like this:

```make
override CFLAGS += -Wall -Wbad-function-cast -Wcast-align ...
```
See [here](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html).

zandrey: > 
> 
> > I haven't managed yet to solve (2), but this assignment has to be changes IMHO since it does not work as expected anyway.
> > If you have some suggestions here on how to solve it - they would be really appreciated, since I'm a bit stuck with it.
> 
> Like this:
> 
> ```makefile
> override CFLAGS += -Wall -Wbad-function-cast -Wcast-align ...
> ```

Spot on! 👍 Thanks a lot, this really solved both cases above, and the issue I was trying to tackle as well!

Thanks a lot @jforissier!

I'll update the patch in this PR.

> 
> See [here](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html).

Guess I should do my *RTFM* more thorough... 😉

zandrey:
Travis shown an interesting failure, which led me to read the Make manual more. It turn out, that once `override` is used on a variable - all consecutive appends are getting ignored.

That is the reason of failed CI check, which I (hopefully) solved with adding `override` to all occurrences of `CFLAGS` in the file.

Strangely enough - this didn't fail my Yocto build. I guess this is due to `-fPIC` is set in the compiler string.

Now the expansion should be complete. I'll re-submit and check if CI goes green.

@jforissier I've added a link to GNU Make Manual into the commit description as well.

-- andrey

zandrey:
OK, Travis complained about the URL in commit message, looks like [bdc48fa](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144) is only out with `v5.7` tree 😃

@jforissier @jbech-linaro Would you accept it or should I re-format the description again?

-- andrey

jforissier:
> OK, Travis complained about the URL in commit message, looks like [bdc48fa](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144) is only out with `v5.7` tree 


That's for code, not sure this will apply to commit descriptions (AFAIK they should be <= 75 columns).

> @jforissier @jbech-linaro Would you accept it or should I re-format the description again?

Fine by me, FYI you can get rid of the warning by moving the URL to a Link: tag.
```
Link: [1] URL
```
Then no more overflow check.

jforissier:
> Travis shown an interesting failure, which led me to read the Make manual more. It turn out, that once `override` is used on a variable - all consecutive appends are getting ignored.

Ah, that I didn't know! Come to think of it, it is not really surprising.

Another option would be to use a different name (`MY_CFLAGS := $(FLAGS) blah blah` then `MY_CFLAGS += blah...`), but this way is fine IMO.

zandrey:
> 
> 
> > OK, Travis complained about the URL in commit message, looks like [bdc48fa](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144) is only out with `v5.7` tree
> 
> That's for code, not sure this will apply to commit descriptions (AFAIK they should be <= 75 columns).

I guess and hope GIT would pick it up as well. 😉 

> 
> > @jforissier @jbech-linaro Would you accept it or should I re-format the description again?
> 
> Fine by me, FYI you can get rid of the warning by moving the URL to a Link: tag.
> 
> ```
> Link: [1] URL
> ```
> 
> Then no more overflow check.

That is a very nice tip! 👍 Didn't know about this one, since usually I try to cut the line in commit messages.

-- andrey

jbech-linaro:

jforissier:
@zandrey would you please add the Reviewed-by: tags?

zandrey:
@jforissier Done, sorry for that. I thought once the merging is done - you would apply those lines. This is how I got it in Linux kernel.

I've also added a `Link:` tag to make Travis happy. 😃 

-- andrey

jforissier:
> @jforissier Done, sorry for that. I thought once the merging is done - you would apply those lines. This is how I got it in Linux kernel.

No worries. I could apply the tags myself but it would be a bit more work. If you do it, I just need to click a button, otherwise I need to `git fetch`/`git rebase`/`git commit --amend`/ and `git push`...

PR #212, opened 2020-06-25
Update tarball makeing procedure

HuKeping:
When creating a release tarball(we're using `git archive`, if I'm right),
it's pointless to include .gitignore and scripts for travisCI.

This patch adds a few entries in .gitattributes to specify files that
should never end up in a distribution tarball.

Signed-off-by: Hu Keping <hukeping@huawei.com>


jforissier:
Hello @HuKeping,

Thanks for the patch. This looks OK to me. In fact I didn't know about "git archive" and .gitattributes so I'm glad I am learning something today! :)
I have some comments on the commit message however:

> When creating a release tarball(we're using `git archive`, if I'm right),

- Missing space before (
- Remove "if I'm right" (who knows better than you what you are using??? certainly not the person that will read your commit message! ;-))

> it's pointless to include .gitignore and scripts for travisCI.
> 
> This patch adds a few entries in .gitattributes to specify files that

"This patch adds a .gitattributes file to specify..."


With that:

jforissier:
Please also change the commit subject to "Add .gitattributes".

HuKeping:
Commit message updated, PTAL @jforissier

HuKeping:
should I add the ack of @etienne-lms to the commit message and re-push it?

PR #217, opened 2020-09-04
libteec: fix TEEC_RegisterSharedMemory() with a fallback option

jenswi-linaro:
Adds a fallback option to TEEC_RegisterSharedMemory() where
TEE_IOC_SHM_REGISTER fails, likely due to the caller supplying read-only
memory. Instead of failing and burden the caller to handle this,
implement it here instead. If TEE_IOC_SHM_REGISTER fails, malloc() a
buffer and register that as a shadow buffer instead.

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -56,6 +56,12 @@
 #define MEMREF_SHM_OFFS(p)	((p)->a)
 #define MEMREF_SIZE(p)		((p)->b)
 
+/*
+ * Internal flags of TEEC_SharedMemory::internal.flags
+ */
+#define SHM_FLAG_BUFFER_ALLOCED		(1u << 0)
+#define SHM_FLAG_SHADOW_BUFFER_ALLOCED	(1u << 1)
+
 static pthread_mutex_t teec_mutex = PTHREAD_MUTEX_INITIALIZER;
 
 static void teec_mutex_lock(pthread_mutex_t *mu)
@@ -739,6 +745,7 @@ void TEEC_RequestCancellation(TEEC_Operation *operation)
 
 TEEC_Result TEEC_RegisterSharedMemory(TEEC_Context *ctx, TEEC_SharedMemory *shm)
 {
+	TEEC_Result res = TEEC_SUCCESS;
 	int fd = 0;
 	size_t s = 0;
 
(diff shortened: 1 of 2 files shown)


jenswi-linaro:
Fixes the error exposed with https://github.com/OP-TEE/optee_test/pull/445

libteec/src/tee_client_api.c
```diff
@@ -756,10 +757,38 @@ TEEC_Result TEEC_RegisterSharedMemory(TEEC_Context *ctx, TEEC_SharedMemory *shm)
 		s = 8;
 	if (ctx->reg_mem) {
 		fd = teec_shm_register(ctx->fd, shm->buffer, s, &shm->id);
-		if (fd < 0)
+		if (fd >= 0) {
+			shm->registered_fd = fd;
+			goto out;
```
etienne-lms, line 762: I think `shm->shadow_buffer = NULL;` is missing in this case.

jenswi-linaro: You're right, thanks.

jenswi-linaro:
Update

public/tee_client_api.h
```diff
@@ -299,6 +299,7 @@ typedef struct {
 	void *shadow_buffer;
 	int registered_fd;
 	bool buffer_allocated;
+	bool shadow_buffer_allocated;
```
jforissier, line 302: ABI breakage? It seems it happened a couple of times already :cry:

```
$ git blame -s -L 290,+13 3.10.0 tee_client_api.h | awk '{system("git describe --always " $1" | tr \\\\n \~"); print}' | column -t -s~
04ac465           04ac465f 290) typedef struct {
2.0.0             88acd6bd 291) 	void *buffer;
2.0.0             88acd6bd 292) 	size_t size;
04ac465           04ac465f 293) 	uint32_t flags;
0.2.0-2-gcaf0745  caf07451 294) 	/*
2.0.0             88acd6bd 295) 	 * Implementation-Defined
0.2.0-2-gcaf0745  caf07451 296) 	 */
2.0.0             88acd6bd 297) 	int id;
2.0.0             88acd6bd 298) 	size_t alloced_size;
2.0.0             88acd6bd 299) 	void *shadow_buffer;
2.2.0-2-ge305d73  e305d73a 300) 	int registered_fd;
2.5.0-4-g27888d7  27888d73 301) 	bool buffer_allocated;
04ac465           04ac465f 302) } TEEC_SharedMemory;
```
This tells me that minor version 2.3.0 was incompatible with 2.2.0 and same for 2.6.0 vs. 2.5.0.
 
Is it time to switch to 4.0.0, and perhaps rework the layout of this structure for future-proofing?

etienne-lms (commented):

public/tee_client_api.h
jenswi-linaro: I'm not sure this causes ABI breakage in practise since there's likely some implicit padding upto at least an even 4 byte boundary after `buffer_allocated`, so `shadow_buffer_allocated` should fit too. Assuming a `bool` is just a byte.

If we don't want to take any chances I guess we have two options. I rework this to not grow `TEEC_SharedMemory` or we switch to `4.0.0`. Switching to `4.0.0` over this seems a bit much to me.

If I'm not to add `bool shadow_buffer_allocated`. What are my options?
1. replace `size_t allocated_size` with `ssize_t allocated_size` and use the sign to indicate "shadow_buffer_allocated"
2. replace `bool buffer_allocated` with
``` C
union {
         bool dummy;
         uint8_t internal_flags;
};
```

jforissier: > I'm not sure this causes ABI breakage in practise since there's likely some implicit padding upto at least an even 4 byte boundary after `buffer_allocated`, so `shadow_buffer_allocated` should fit too. Assuming a `bool` is just a byte.

Makes sense, but I'd rather make this more explicit as you are suggesting below.

> 
> If we don't want to take any chances I guess we have two options. I rework this to not grow `TEEC_SharedMemory` or we switch to `4.0.0`. Switching to `4.0.0` over this seems a bit much to me.

Agreed.

> 
> If I'm not to add `bool shadow_buffer_allocated`. What are my options?
> 
> 1. replace `size_t allocated_size` with `ssize_t allocated_size` and use the sign to indicate "shadow_buffer_allocated"
> 2. replace `bool buffer_allocated` with
> 
> ```c
> union {
>          bool dummy;
>          uint8_t internal_flags;
> };
> ```

That looks good to me :+1:

jenswi-linaro:
Updated as requested. @etienne-lms, OK with this?

jforissier (approved):

libteec/src/tee_client_api.c
```diff
@@ -56,6 +56,12 @@
 #define MEMREF_SHM_OFFS(p)	((p)->a)
 #define MEMREF_SIZE(p)		((p)->b)
 
+/*
+ * Internal flags of TEEC_SharedMemory::internal.flags
+ */
+#define SHM_FLAG_BUFFER_ALLOCED		0x1
```
etienne-lms, line 62: prefer `BIT(0)` and `BIT(1)` as these are flags.
Seems `BIT()` is not support in **libteec**. I suggest `(1 << 0)` and `(1 << 1)`.

etienne-lms (commented):
looks good to me with a minor comment.

jenswi-linaro:
Squashed and fixed up as suggested. Tags applied.

PR #221, opened 2020-10-14
libteec: Handle NULL pointer when using TEEC_TempMemoryReference

ruchi393:
If TEE or TEE driver doesn't support the capability "TEE_GEN_CAP_MEMREF_NULL",
"memref_null" is not set in the TEEC_Context. In such cases, while
handling parameters of type TEEC_TempMemoryReference, a call to
TEEC_RegisterSharedMemory() will return an error for NULL pointers.
Using NULL pointer is a valid use case and should not result in errors.
Use TEEC_AllocateSharedMemory() to avoid errors in such cases.

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -202,13 +202,20 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 	}
 	shm->size = tmpref->size;
 
-	if (!tmpref->buffer && ctx->memref_null) {
+	if (!tmpref->buffer) {
 		if (tmpref->size)
 			return TEEC_ERROR_BAD_PARAMETERS;
 
-		/* Null pointer, indicate no shared memory attached */
-		MEMREF_SHM_ID(param) = TEE_MEMREF_NULL;
-		shm->id = -1;
+		if (ctx->memref_null) {
+			/* Null pointer, indicate no shared memory attached */
+			MEMREF_SHM_ID(param) = TEE_MEMREF_NULL;
+			shm->id = -1;
+		} else {
+			res = TEEC_AllocateSharedMemory(ctx, shm);
+			if (res != TEEC_SUCCESS)
+				return res;
+			MEMREF_SHM_ID(param) = shm->id;
+		}
 	} else {
 		shm->buffer = tmpref->buffer;
 		res = TEEC_RegisterSharedMemory(ctx, shm);


libteec/src/tee_client_api.c
```diff
@@ -209,6 +209,14 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 		/* Null pointer, indicate no shared memory attached */
 		MEMREF_SHM_ID(param) = TEE_MEMREF_NULL;
 		shm->id = -1;
+	} else if (!tmpref->buffer) {
+		res = TEEC_AllocateSharedMemory(ctx, shm);
+                if (res != TEEC_SUCCESS)
+                        return res;
+
+		memcpy(shm->buffer, tmpref->buffer, tmpref->size);
```
jenswi-linaro, line 217: What if `tmpref->size` isn't `0`?

libteec/src/tee_client_api.c
```diff
@@ -209,6 +209,14 @@ static TEEC_Result teec_pre_process_tmpref(TEEC_Context *ctx,
 		/* Null pointer, indicate no shared memory attached */
 		MEMREF_SHM_ID(param) = TEE_MEMREF_NULL;
 		shm->id = -1;
+	} else if (!tmpref->buffer) {
```
jenswi-linaro, line 212: How about something like this (starting replacing from line 205 above):
``` C
        if (!tmpref->buffer) {
                if (tmpref->size)

                 if (ctx->memref_null) {
                        /* Null pointer, indicate no shared memory attached */
                        MEMREF_SHM_ID(param) = TEE_MEMREF_NULL;
                        shm->id = -1;
                } else {
                        res = TEEC_AllocateSharedMemory(ctx, shm);
                        if (res != TEEC_SUCCESS)
                                return res;
                        MEMREF_SHM_ID(param) = shm->id;
                }
        } else {
```

ruchi393: Sure, this makes sense.  Previous implementation of this flow without the memref_null set had the memcpy, though it didn't make any logical sense to have that memcpy there. I left it there inorder to avoid any other hidden failures.

I will make this change and re-submit the patch.

jenswi-linaro (commented):

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`

In commit log, maybe remove  the last sentence "but ensure the size is set to 0.".

jforissier:
Thanks @ruchi393 for working on this on such a short notice :+1:

PR #230, opened 2020-11-02
libckteec: replace { } initializer by { 0 }

ffontaine:
Replace { } initializer by { 0 } to fix the following build failure with gcc 4.8:

```
libckteec/src/pkcs11_processing.c: In function 'ck_create_object':
libckteec/src/pkcs11_processing.c:22:9: error: missing initializer for field 'buffer' of 'struct serializer' [-Werror=missing-field-initializers]
  struct serializer obj = { };
         ^
```
Fixes: http://autobuild.buildroot.org/results/a3d663adb943aee814180f01d6e153b3309be962
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>
Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>

### DIFF
--- a/libckteec/src/pkcs11_processing.c
+++ b/libckteec/src/pkcs11_processing.c
@@ -19,7 +19,7 @@ CK_RV ck_create_object(CK_SESSION_HANDLE session, CK_ATTRIBUTE_PTR attribs,
 		       CK_ULONG count, CK_OBJECT_HANDLE_PTR handle)
 {
 	CK_RV rv = CKR_GENERAL_ERROR;
-	struct serializer obj = { };
+	struct serializer obj = { 0 };
 	size_t ctrl_size = 0;
 	TEEC_SharedMemory *ctrl = NULL;
 	TEEC_SharedMemory *out_shm = NULL;
@@ -114,7 +114,7 @@ CK_RV ck_encdecrypt_init(CK_SESSION_HANDLE session,
 {
 	CK_RV rv = CKR_GENERAL_ERROR;
 	TEEC_SharedMemory *ctrl = NULL;
-	struct serializer obj = { };
+	struct serializer obj = { 0 };
 	uint32_t session_handle = session;
 	uint32_t key_handle = key;
 	size_t ctrl_size = 0;
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -39,7 +39,7 @@ static CK_RV serialize_indirect_attribute(struct serializer *obj,
 	CK_ATTRIBUTE_PTR attr = NULL;
 	CK_ULONG count = 0;
 	CK_RV rv = CKR_GENERAL_ERROR;
-	struct serializer obj2 = { };
+	struct serializer obj2 = { 0 };
 
 	switch (attribute->type) {
 	/* These are serialized each separately */
@@ -263,7 +263,7 @@ static CK_RV serialize_mecha_aes_iv(struct serializer *obj,
 CK_RV serialize_ck_mecha_params(struct serializer *obj,
 				CK_MECHANISM_PTR mechanism)
 {
-	CK_MECHANISM mecha = { };
+	CK_MECHANISM mecha = { 0 };
 	CK_RV rv = CKR_GENERAL_ERROR;
 
 	memset(obj, 0, sizeof(*obj));
--- a/libckteec/src/serializer.c
+++ b/libckteec/src/serializer.c
@@ -14,7 +14,7 @@
 
 CK_RV init_serial_object(struct serializer *obj)
 {
-	struct pkcs11_object_head head = { };
+	struct pkcs11_object_head head = { 0 };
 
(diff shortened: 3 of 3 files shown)


etienne-lms:
Hi Fabrice,
I'm a bit concerned with this change. I've tested it using GCC 9.2 and it works fine. However, `{ 0 }` is neither a universal zeroing initializer for gcc 4.8. I known **optee_os** sources quite use `{ }` initializers. Indeed building **optee_os** with GCC 4.8 produced the same issues you pointed. However, changing `{ }` to `{ 0 }` does not solve this issue. I faced this following kind of errors (when a struct 1st field is a struct or array):
```shell
(...)
core/arch/arm/kernel/unwind_arm32.c: In function ‘unw_get_kernel_stack’:
core/arch/arm/kernel/unwind_arm32.c:62:9: warning: missing braces around initializer [-Wmissing-braces]
  struct unwind_state_arm32 state = { 0 };
         ^
core/arch/arm/kernel/unwind_arm32.c:62:9: warning: (near initialization for ‘state.registers’) [-Wmissing-braces]
(...)
core/arch/arm/kernel/rpc_io_i2c.c: In function ‘rpc_io_i2c_transfer’:
core/arch/arm/kernel/rpc_io_i2c.c:21:9: warning: missing braces around initializer [-Wmissing-braces]
  struct thread_param p[4] = { 0 };
         ^
```

I wonder if we shall mandate gcc-7.x (IIRC gcc-7 was fine with `{ }`, to check) or above for optee_client and other OP-TEE packages. 

Or maybe, focusing only on optee_client, we could change it to have specific initializers (strictly adapted) in the the source to support older toolchains.

jenswi-linaro:
> Or maybe, focusing only on optee_client, we could change it to have specific initializers (strictly adapted) in the the source to support older toolchains.

I'm also open for a more pragmatic approach to the problem in optee_client since that's likely to be compiled in a wider range of environments than any of the other gits. The diff as it is looks like a good approach to me.

etienne-lms:
Ok, thanks @jenswi-linaro for the feedback.
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`.

Travis CI failure is a false positive.

@ffontaine, could you change the commit log so that it's more consistent with other optee_client commit logs:
```patch
-fix build with gcc 4.8
+libckteec: replace { } initializer by { 0 }
 
 Replace { } initializer by { 0 } to fix the following build failure with
 gcc 4.8:
 
 /home/giuliobenetti/autobuild/run/instance-2/output-1/build/optee-client-3.11.0/libckteec/src/pkcs11_processing.c: In function 'ck_create_object':
 /home/giuliobenetti/autobuild/run/instance-2/output-1/build/optee-client-3.11.0/libckteec/src/pkcs11_processing.c:22:9: error: 
 missing initializer for field 'buffer' of 'struct serializer' [-Werror=missing-field-initializers]
   struct serializer obj = { };
          ^
 
-Fixes:
- - http://autobuild.buildroot.org/results/a3d663adb943aee814180f01d6e153b3309be962
-
+Fixes: http://autobuild.buildroot.org/results/a3d663adb943aee814180f01d6e153b3309be962
 Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>
```

etienne-lms:
Thanks @ffontaine.
I forgot to mention: can you also add the review tag (here `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`) to the commit log so that maintainer can merge this change?

etienne-lms (approved):

jforissier:
The checkpatch warning should go away if you use relative paths in the error messages, in other words remove everything up to e.g., `libckteec` like so:
```
 libckteec/src/pkcs11_processing.c: In function 'ck_create_object':
 libckteec/src/pkcs11_processing.c:22:9: error: missing initializer for field 'buffer' of 'struct serializer' [-Werror=missing-field-initializers]
   struct serializer obj = { };
          ^
```

Note, I am on vacation this week. I will occasionally check for PRs that are ready for merging but expect slower response time from me.

etienne-lms:
> The checkpatch warning should go away if you use relative paths in the error messages, in other words remove everything up to e.g., libckteec like so:

I guess trace messages can always be quite long and there's no generic way to get checkout not complaining unless there's a trace tag which would be a bit cumbersome IMO. Not bad there are still things to be handled manually ;)

There not hurry for merging. Anyway thanks for the feedback, have a nice vacation time.

jforissier:
@etienne-lms no guarantee indeed, but in case of compiler errors, overly long lines tend to be mostly the ones with the file paths followed by the error message, not the copy of the source code lines. So the checkpatch heuristic of iognoring lines starting with what looks like a path is actually quite good, unfortunately it identifies only relative paths not absolute ones (because that's what happens with the Linux kernel).

etienne-lms:
@ffontaine, looks like you duplicated the S-o-b tag.
@jforissier, thanks for the explanations. Checkpatch always surprises me!

PR #232, opened 2020-11-03
Pkcs11 : Add Signing and Verification functions for HMAC

ruchi393:


libckteec/include/pkcs11_ta.h
```diff
@@ -341,6 +341,84 @@ enum pkcs11_ta_cmd {
...
 	PKCS11_CMD_DECRYPT_ONESHOT = 24,
+
+	/*
+	 * PKCS11_CMD_SIGN_INIT - Initialize a signature computation processing
+	 * PKCS11_CMD_VERIFY_INIT - Initialize a signature verification processing
+	 *
+	 * [in]  memref[0] = [
+	 *              32bit session handle,
+	 *              32bit key handle,
+	 *              (struct pkcs11_attribute_head)mechanism + mecha params,
```
jbech-linaro, line 352: s/mecha/mechanism/

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
 }
+
+CK_RV ck_signverify_init(CK_SESSION_HANDLE session,
+			 CK_MECHANISM_PTR mechanism,
+			 CK_OBJECT_HANDLE key,
+			 int sign)
+{
+	CK_RV rv = CKR_GENERAL_ERROR;
+	TEEC_SharedMemory *ctrl = NULL;
+	struct serializer obj;
```
jbech-linaro, line 353: Initialize please.

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+
+	/* Shm io0: (in/out) ctrl = [session-handle] / [status] */
+	ctrl = ckteec_alloc_shm(sizeof(session_handle), CKTEEC_SHM_INOUT);
+	if (!ctrl) {
+		rv = CKR_HOST_MEMORY;
+		goto bail;
+	}
+	memcpy(ctrl->buffer, &session_handle, sizeof(session_handle));
+
+	/* Shm io1: intput data */
```
jbech-linaro, line 419: s/intput/input/

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+	if (in_len) {
```
jbech-linaro, line 420: Do we also need to set an upper bound here?

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+	/* Shm io0: (in/out) ctrl = [session-handle] / [status] */
+	ctrl = ckteec_alloc_shm(sizeof(session_handle), CKTEEC_SHM_INOUT);
+	if (!ctrl) {
+		rv = CKR_HOST_MEMORY;
+		goto bail;
+	}
+	memcpy(ctrl->buffer, &session_handle, sizeof(session_handle));
+
+	/* Shm io1: input payload */
+	if (in_len) {
```
jbech-linaro, line 464: Same as previous function, do we need to set an upper bound that we allow?

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+			goto bail;
+		}
+	}
+
+	/*
+	 * Shm io2: input signature (if verifying) or null sized shm
+	 * or
+	 * Shm io2: output signature (if signing) or null sized shm
+	 */
+	if (sign_len && *sign_len) {
```
jbech-linaro, line 477: Upper bound?

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+		goto bail;
+	}
+	memcpy(ctrl->buffer, &session_handle, sizeof(session_handle));
+
+	/*
+	 * Shm io1: input signature (if verifying) or null sized shm
+	 * or
+	 * Shm io1: output signature (if signing) or null sized shm
+	 */
+	if (sign_len && *sign_len)
```
jbech-linaro, line 535: Upper bound needed?

libckteec/src/pkcs11_processing.h
```diff
@@ -36,4 +36,28 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+			    CK_BYTE_PTR out,
+			    CK_ULONG_PTR out_len,
+			    int sign);
+
+CK_RV ck_signverify_final(CK_SESSION_HANDLE session,
+			  CK_BYTE_PTR out,
+			  CK_ULONG_PTR out_len,
+			  int sign);
+
+
```
etienne-lms, line 62: remove at least one of the  empty lines.

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+		rv = CKR_HOST_MEMORY;
+		goto bail;
+	}
+
+	rv = ckteec_invoke_ta(sign ? PKCS11_CMD_SIGN_ONESHOT :
+			      PKCS11_CMD_VERIFY_ONESHOT, ctrl,
+			      io1, io2, &out_size, NULL, NULL);
+
+	if (sign && (rv == CKR_OK || rv == CKR_BUFFER_TOO_SMALL))
+		*sign_len = out_size;
```
etienne-lms, line 495: Seems null `sign_len` is supported even is `sign == CK_TRUE`. In such case this would segfault.
I think we should add return `CKR_ARGUMENTS_BAD` at this function entry if `sign && !sign_len`.

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+
+	if (!io) {
+		rv = CKR_HOST_MEMORY;
+		goto bail;
+	}
+
+	rv = ckteec_invoke_ctrl_out(sign ? PKCS11_CMD_SIGN_FINAL : PKCS11_CMD_VERIFY_FINAL,
+					    ctrl, io, &out_size);
+
+	if (sign && sign_len && (rv == CKR_OK || rv == CKR_BUFFER_TOO_SMALL))
```
etienne-lms, line 549: here also, I wonder if `sign && !sign_len` should return `CKR_ARGUMENTS_BAD` at function entry.

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,220 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+	else
+		io = ckteec_alloc_shm(0, sign ? CKTEEC_SHM_OUT : CKTEEC_SHM_IN);
+
+	if (!io) {
+		rv = CKR_HOST_MEMORY;
+		goto bail;
+	}
+
+	rv = ckteec_invoke_ctrl_out(sign ? PKCS11_CMD_SIGN_FINAL : PKCS11_CMD_VERIFY_FINAL,
+					    ctrl, io, &out_size);
```
etienne-lms, line 547: indentation

libckteec/src/pkcs11_processing.c
ruchi393: You are right. I will add the check

libckteec/src/pkcs11_processing.c
ruchi393: I will add the check

libckteec/src/pkcs11_processing.c
ruchi393: Ok

libckteec/src/pkcs11_processing.c
ruchi393: @etienne-lms , are you saying that in_len check is not needed here or upper bound check is not required ?

etienne-lms: In the TA, we can strictly check no input memref is provided when not needed. So here, register/pass input memref only when it is expected/used. This (these) API are used by all sign/verify mechanisms. Overkillling?

jbech-linaro: I've already had a chat with Ruchika separately, but the reason for asking initially was that we directly or indirectly get the information from user provided calls. I.e., my concern was that someone tries to abuse the API and try to exploit whatever happens, crash, out-of-memory, buffer overflow. I suppose that even if the spec doesn't mention it, there is probably a limit somewhere there it no longer makes sense to allow greater "xyz_length". I.e., is it likely that a user calls `C_VerifyUpdate(...) -> ck_signverify_update(..)` with `ulPartLen / in_len` = ... 1MB, 100MB, 1GB? :)

etienne-lms: Such limit may be related to a mechanism. I think only the TA at some point knows what make sense.

jbech-linaro: OK, well, you for sure know this better than I do, so I'll take your word here.

ruchi393: > In the TA, we can strictly check no input memref is provided when not needed. So here, register/pass input memref only when it is expected/used. This (these) API are used by all sign/verify mechanisms. Overkillling?
@etienne-lms , are you suggesting I change this check to 
if (!in)

As removing any check all-together from here will result in ckteec_register_shm returning error.

Also the scenario where in is a valid pointer but length passed is 0,  is this a valid scenario or an error case ?

etienne-lms: From spec 2.40-e01 section "5.2 Conventions for functions returning output in a variable-length buffer":
> 1. If pBuf is NULL_PTR, then (...) return (in *pulBufLen) a number of bytes (...). CKR_OK is returned by the function.
> 2. If pBuf is not NULL_PTR, then *pulBufLen MUST contain the size in bytes of the buffer (...). If
the buffer is not large enough, then CKR_BUFFER_TOO_SMALL is returned.

This is were returning OK and TOO_SMALL should follow spec for the TA interface consistency, I think.

ruchi393: You are right, this is for output parameters where if NULL pointer is passed, TA is supposed to return the size of the output buffer.

The above case with in_len  is for input parameter. For an update operation for sign/verify, input buffer and length are being passed. The only valid scenario here is if input buffer is not NULL and length is not zero. I think for rest, we should return CKR_BAD_ARGUMENT error, which is what happens with the current code.

etienne-lms: We can't register NULL buffers, but we can register zero sized memory (from a valid memory pointer), right?
In the end, I think @jbech-linaro comment should be addressed with @ruchi393 suggestion:
Replace memref register conditions `if (in_len)` with `if (in)`, on the NULL pointer case, not the length.

ruchi393: @etienne-lms , there is one issue with replacing in_len with in. If the in is a valid pointer and in_len is 0, call to TEEC_RegisterSharedMemory() results in allocation of a buffer of length 8 bytes. As a result, the TA considers it a valid operation and update passes which is not the right sceanrio.

Looking at all the cases, I think if (in_len) is the right check here.

Given below is my understanding of the checks and error scenarios. Please review.
1. in != NULL, in_len != NULL -> Valid scenario, update should happen
2. in != NULL in_len = 0 -> Error scenario, BAD ARGUMENTS
3. in == NULL in_len != 0 -> Error scenaio -> BAD ARGUMENTS
4. in == NULL and in_len == 0 -> Error sceanrio -> BAD ARGUMENTS

libckteec/src/pkcs11_processing.c
ruchi393: Done

libckteec/src/pkcs11_processing.c
ruchi393: Done

libckteec/src/pkcs11_processing.c
ruchi393: Adding upper bound check would also be valid for other API's in this file. Will need to evaluate it more. Will submit a separate patch for this.

libckteec/include/pkcs11_ta.h
```diff
@@ -341,6 +341,84 @@ enum pkcs11_ta_cmd {
...
+	 *	 ]
+	 * [out] memref[0] = 32bit return code, enum pkcs11_rc
+	 *
+	 * These commands relate to the PKCS#11 API functions C_SignInit() and
+	 * C_VerifyInit().
+	 */
+	PKCS11_CMD_SIGN_INIT = 136,
```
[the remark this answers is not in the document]
ruchi393: Done

libckteec/src/pkcs11_processing.h
ruchi393: Done

libckteec/src/pkcs11_processing.c
ruchi393: added

libckteec/src/pkcs11_processing.c
ruchi393: Added

libckteec/src/pkcs11_processing.c
ruchi393: Done

libckteec/src/pkcs11_processing.c
```diff
@@ -342,3 +342,221 @@ CK_RV ck_encdecrypt_final(CK_SESSION_HANDLE session,
...
+	/*
+	 * Shm io2: input signature (if verifying) or null sized shm
+	 * or
+	 * Shm io2: output signature (if signing) or null sized shm
+	 */
+	if (sign_len && *sign_len) {
+		io2 = ckteec_register_shm(sign_ref, *sign_len,
+					  sign ? CKTEEC_SHM_OUT : CKTEEC_SHM_IN);
+	} else {
+		/* Query size if ouput signature */
```
etienne-lms, line 482: s/ouput/output/

libckteec/src/pkcs11_processing.c
ruchi393: @etienne-lms , were you able to review this ?

libckteec/include/pkcs11_ta.h
```diff
@@ -413,8 +412,7 @@ enum pkcs11_ta_cmd {
 	PKCS11_CMD_SIGN_ONESHOT = 31,
 
 	/*
-	 * PKCS11_CMD_VERIFY_ONESHOT - Update and finalize a signature
-	 *                             verification
+	 * PKCS11_CMD_VERIFY_ONESHOT - Compute a signature
```
etienne-lms, line 415: "Compute and compare a signature"

libckteec/src/pkcs11_processing.c
etienne-lms: I agree that if `in_len!=0` then we expect in input buffer to be registered.
If `in_len==0`, there's no input data. Let's the secure world return an error if the input is really expected.

etienne-lms:
@ruchi393, I think you can squash the fixup commits and set all comments as resolved in this P-R.

ruchi393:
All review comments till now have been addresses, patches squashed and branch pushed for review.

2 things are pending which will be taken care in separate PR's as they apply to generic approach and would also affect the C_Encrypt/Decrypt functions:
1. Upper bound checks
2. Update scenario with valid input pointer and 0 length. Current patch set returns CKR_BAD_ARGUMENT in this scenario. We might want to just do nothing and return CKR_OK from here. Specification is not very clear on this scenario.

ricardosalveti:
Only concerns I had is what is described at the comment above, so fine to move those 2 points to another pr.

`Reviewed-by: Ricardo Salveti <ricardo@foundries.io>
`

ruchi393:
Rebased and tags applied

PR #233, opened 2020-11-16
Use pvalloc for shared memory

jenswi-linaro:


jforissier:
- What issue may happen if two shared memory buffers share a page?
- `man pvalloc` mentions that the function is obsolete. Could it be written differently?

jenswi-linaro:
> * What issue may happen if two shared memory buffers share a page?

Without a hypervisor at S-EL2 the pages are just aliased and shouldn't cause any trouble except that it might be a bit harder to track down buffer overruns in shared memory buffers. With FF-A and a hypervisor at S-EL2 (post ARM v8.4 architecture) there's trouble since this isn't permitted.

> * `man pvalloc` mentions that the function is obsolete. Could it be written differently?

Sure, it was only very convenient since it does exactly what's needed and has always been there. I'll try to replace it with something portable.

jforissier:
> > * What issue may happen if two shared memory buffers share a page?
> 
> Without a hypervisor at S-EL2 the pages are just aliased and shouldn't cause any trouble except that it might be a bit harder to track down buffer overruns in shared memory buffers. With FF-A and a hypervisor at S-EL2 (post ARM v8.4 architecture) there's trouble since this isn't permitted.

I'm not sure I understand 100% :-/ Are we talking about secure world pages? I understand that you want to avoid mapping the same physical memory in two different pages in secure world. It could happen if two shm buffers have physical addresses that land in the same page in secure world. By using `pvalloc()` you would make sure each buffer starts at a page boundary so that would avoid the issue. But, what if page sizes are different in normal and secure world? (`pvalloc()` could use a smaller granularity than secure world, no?).

jenswi-linaro:
> > > * What issue may happen if two shared memory buffers share a page?
> > 
> > 
> > Without a hypervisor at S-EL2 the pages are just aliased and shouldn't cause any trouble except that it might be a bit harder to track down buffer overruns in shared memory buffers. With FF-A and a hypervisor at S-EL2 (post ARM v8.4 architecture) there's trouble since this isn't permitted.
> 
> I'm not sure I understand 100% :-/ Are we talking about secure world pages?

Yes, sorry being more clear.

> I understand that you want to avoid mapping the same physical memory in two different pages in secure world. It could happen if two shm buffers have physical addresses that land in the same page in secure world. By using `pvalloc()` you would make sure each buffer starts at a page boundary so that would avoid the issue. But, what if page sizes are different in normal and secure world? (`pvalloc()` could use a smaller granularity than secure world, no?).

In theory I we could have some conflict here, but if that would ever happen we'd have much larger problem. The shared memory handling currently depend on OP-TEE being able to map with the required granularity.

jforissier:
>> But, what if page sizes are different in normal and secure world? (`pvalloc()` could use a smaller granularity than secure world, no?).
> 
> In theory I we could have some conflict here, but if that would ever happen we'd have much larger problem. The shared memory handling currently depend on OP-TEE being able to map with the required granularity.

OK.

If we increase the alignment of the shm buffers we will waste some space, could it be significant?
Is it not possible to be smarter when mapping the buffers in secure world in order to avoid mapping them twice?

jenswi-linaro:
> > > But, what if page sizes are different in normal and secure world? (`pvalloc()` could use a smaller granularity than secure world, no?).
> > 
> > 
> > In theory I we could have some conflict here, but if that would ever happen we'd have much larger problem. The shared memory handling currently depend on OP-TEE being able to map with the required granularity.
> 
> OK.
> 
> If we increase the alignment of the shm buffers we will waste some space, could it be significant?

Increase where? System wide? Yes we'd waste some memory.

> Is it not possible to be smarter when mapping the buffers in secure world in order to avoid mapping them twice?

No, that's impossible or at least it's impossible to be smart enough as you'd at a minimum would need to be able to make accurate predictions of the future.

jforissier:
I mean by using `pvalloc()` (or `posix_memalign()` with page alignment) instead of `malloc()` for small buffers there will be more fragmentation and unused space in the heap I think. Maybe not much, maybe not an issue.

When secure world receives a physical address for a shared buffer, doesn't it have a way to find out if there is already some mapping for it? A `phys_to_virt()` function more or less. With that is should be possible to avoid multiple mappings?

libteec/src/tee_client_api.c
```diff
@@ -74,6 +74,16 @@ static void teec_mutex_unlock(pthread_mutex_t *mu)
 	pthread_mutex_unlock(mu);
 }
 
+static void *teec_paged_aligned_alloc(size_t sz)
+{
+	void *p = NULL;
+
+	if (!posix_memalign(&p, sysconf(_SC_PAGESIZE), sz))
```
jforissier, line 81: I have no idea of the performance impact of calling `sysconf` every time, but since the result will not change, the value could perhaps be cached.

jenswi-linaro:
> I mean by using `pvalloc()` (or `posix_memalign()` with page alignment) instead of `malloc()` for small buffers there will be more fragmentation and unused space in the heap I think. Maybe not much, maybe not an issue.
> 

Keep in mind that this is only used for memory shared with secure world so the number of buffers will not be very large.

> When secure world receives a physical address for a shared buffer, doesn't it have a way to find out if there is already some mapping for it? A `phys_to_virt()` function more or less. With that is should be possible to avoid multiple mappings?

If all buffers where just a single page that would be possible. But assume more than one page is shared in one shared memory object and we have multiple such object. At some point yet another range of memory is shared which happens to use half the last page of some of the other earlier shared memory object. How can that be handled assuming that all the earlier shared memory objects are mapped, which they will be with a hypervisor at S-EL2, only they are mapped as IPA.

Any way, in the FF-A specification it's not allowed to share a physical page more than once so the one sharing the memory must make sure to avoid that.

libteec/src/tee_client_api.c
jenswi-linaro: On Linux I expect that `sysconf(_SC_PAGESIZE)` doesn't result in a syscall so the impact should be fairly low.
In glibc `pvalloc()` doesn't cache the value, instead it's determined with `GLRO (dl_pagesize)` which `sysconf(_SC_PAGESIZE)` also ends up doing in `__getpagesize()`.

This function may be called concurrently so if we implement some caching we'll need synchronization.

jforissier:
What if the client application calls `TEEC_RegisterSharedMemory()` on two buffers that share the same page? (say two small buffers allocated in a row with `malloc()`, they are quite likely to share a page).

libteec/src/tee_client_api.c
jforissier: OK. Now that you're mentioning Linux... will it work on Android? I'm not sure `sysconf(_SC_PAGESIZE)` is implemented in bionic.

jenswi-linaro: It seems to be implemented there too.

jenswi-linaro:
That's where this code kicks in:
https://github.com/OP-TEE/optee_client/blob/61d8105953c1f08c34bc28b1b726acd48abe2b5f/libteec/src/tee_client_api.c#L790-L816

Since I've linked to upstream we're still using `malloc()` instead of `teec_paged_aligned_alloc()` though.

jforissier:
You mean `teec_shm_register()` would fail with the second buffer?

jenswi-linaro:
Yes, that's the idea. So this part of the code would serve two purposes.
In the prototype we're crashing a bit later instead, but hopefully it shouldn't be too hard to return an early error. :-)

jforissier:
> In the prototype we're crashing a bit later instead, but hopefully it shouldn't be too hard to return an early error. :-)

Again I don't totally get it :-/ I think it is too soon for such a change. There is currently no problem with this code with existing configurations, and it's hard to tell if it is the right way to go without seeing all the related code, I mean secure world, plus Linux driver changes perhaps (as I understand it).

libteec/src/tee_client_api.c
jforissier: Yep found it now, good.

jenswi-linaro:
> > In the prototype we're crashing a bit later instead, but hopefully it shouldn't be too hard to return an early error. :-)
> 
> Again I don't totally get it :-/

Quite understandable, the spec https://static.docs.arm.com/den0077/a/DEN0077A_PSA_Firmware_Framework_Arm_v8-A_1.0_EAC.pdf takes a while to grok.

> I think it is too soon for such a change. There is currently no problem with this code with existing configurations, and it's hard to tell if it is the right way to go without seeing all the related code, I mean secure world, plus Linux driver changes perhaps (as I understand it).

Affected parts are: Linux kernel driver, Non-secure hypervisor (Hafnium in the prototype), TF-A, Secure hypervisor/SPM (secure world configuration of Hafnium in the prototype), and finally OP-TEE. 

I'm afraid it will take some effort to understand all that code. I can give links to all the code if you think it would help. I can give a bit more high level details about the crash as I understand it if you're interested. It's probably easier check the specification since that sets the boundaries for the code. 


This recovery mechanism was a happy coincidence, the code is already there and if it works in the new case too that's nice. If it doesn't no harm done in upstream as there's no change (except the replacement of `malloc()`)

When it comes to making sure that that all shared memory buffers are page aligned it will have a rather small impact on current code.

1. No aliasing of mapped shared memory in secure world
    easier to debug eventual buffer overruns since changes only occurs in one secure virtual memory range or shared memory object
2. Non-secure user space uses slightly more memory for shared memory since each buffer starts on a new page
   a. virtually no impact on a client which is already has a somewhat efficient shared memory usage
   b. in a pathological case where a client allocates lots of small and long lived shared memory buffers there will be wasted memory

An alternative is to add runtime configuration if this changed shared memory allocation policy is needed or not. The impact on the current code is in my opinion too small to justify runtime configuration.

The benefit with merging this patch now is that it's one patch/git less to keep track of in the prototype.

jforissier:
>> You mean `teec_shm_register()` would fail with the second buffer?
> Yes, that's the idea. So this part of the code would serve two purposes.
...
> The benefit with merging this patch now is that it's one patch/git less to keep track of in the prototype.

OK, I'm fine with this provided that:
1. It has no adverse side-effect on current users (your point 2b above is what I am worried about, but if we don't get more comments we can probably assume it's OK)
2. The change is consistent with the rest of the code even before the FFA stuff is merged, and currently it means that the `TEE_IOC_SHM_REGISTER` ioctl should fail when called with a non paged-aligned pointer or a pointer already registered, because you can't claim this patch does what it is supposed to do ("doesn't accidentally share a page of memory") without that. Or am I missing something?

jenswi-linaro:
> > > You mean `teec_shm_register()` would fail with the second buffer?
> > > Yes, that's the idea. So this part of the code would serve two purposes.
> > > ...
> > > The benefit with merging this patch now is that it's one patch/git less to keep track of in the prototype.
> 
> OK, I'm fine with this provided that:
> 
> 1. It has no adverse side-effect on current users (your point 2b above is what I am worried about, but if we don't get more comments we can probably assume it's OK)

OK

> 2. The change is consistent with the rest of the code even before the FFA stuff is merged, and currently it means that the `TEE_IOC_SHM_REGISTER` ioctl should fail when called with a non paged-aligned pointer or a pointer already registered, because you can't claim this patch does what it is supposed to do ("doesn't accidentally share a page of memory") without that. Or am I missing something?

Currently `TEE_IOC_SHM_REGISTER` can fail as described in the comment (around line 801).  With FF-A stuff it could also fail because the hypervisor refuses to register/share a physical page twice. Note that the first condition can still cause a failure with FF-A. There are no checks to see if a supplied buffer is page aligned. They don't have to be page aligned, but it's a very convenient way of avoiding using the same page twice is to make the allocations page aligned.

Regardless of why registration of a buffer failed, the fall back is to allocate a shadow buffer and use that instead. To make sure the that shadow buffer doesn't cause any conflicts we're making a page aligned allocation.

jforissier:
Sounds reasonable, although I still don't see why this could not be hidden from user space altogether. You said "With FF-A stuff it could also fail because the hypervisor refuses to register/share a physical page twice". In this situation the hypervisor has a mapping already so why not just use it (and maybe reference-count it)?

jenswi-linaro:
Earlier mentioned that the FF-A 1.0 specification forbids this, now I've found the place, see "Table 36: Share memory transaction state machine". The reason is to keep complexity out of the non-secure hypervisor and SPMC (secure "hypervisor").

Until now in OP-TEE we've maintained a very flexible policy towards shared memory as a way of keeping complexity out of secure world. In this case it's the other way around, we add restrictions in order to minimize complexity. There's a big difference between keeping track of how different ranges are shared with other SPs or VMs compared to manage this per physical page with added reference counters. Even more so if you consider the corner cases that can be avoided. I suppose memory footprint also is an aspect.

Since additional complexity at more privileged levels can be avoided by user space being nice I don't see why that shouldn't be required given that it doesn't affect the API (TEE Client API that is).

jforissier:
> Earlier mentioned that the FF-A 1.0 specification forbids this, now I've found the place, see "Table 36: Share memory transaction state machine".

Wow! That looks complex... :open_mouth:

>  In this case it's the other way around, we add restrictions in order to minimize complexity.

OK, overall I tend to agree but I'm thinking this somewhat breaks the kernel/user space ABI. Consider for example a client application linked against an older `libteec`. Then the user upgrades the firmware and kernel to a brand new configuration with FF-A. Now the `TEE_IOC_SHM_REGISTER` `ioctl()` may fail because of the new constraints on the buffer. We could even consider the case of an application not using our `libteec`, although I admit it is not very likely.

Isn't the OP-TEE driver supposed to maintain full backward compatibility?

jenswi-linaro:
> > Earlier mentioned that the FF-A 1.0 specification forbids this, now I've found the place, see "Table 36: Share memory transaction state machine".
> 
> Wow! That looks complex... 

Yeah, it is, but so is the use case.

> 
> > In this case it's the other way around, we add restrictions in order to minimize complexity.
> 
> OK, overall I tend to agree but I'm thinking this somewhat breaks the kernel/user space ABI. Consider for example a client application linked against an older `libteec`. Then the user upgrades the firmware and kernel to a brand new configuration with FF-A. Now the `TEE_IOC_SHM_REGISTER` `ioctl()` may fail because of the new constraints on the buffer. We could even consider the case of an application not using our `libteec`, although I admit it is not very likely.
> 
> Isn't the OP-TEE driver supposed to maintain full backward compatibility?

Yes in this case it does that for non-FF-A systems. This is a reasonable compromise for FF-A systems in my opinion. I think we have these options:
1. This share a page only once restriction in a FF-A system
2. Adding a flag that prevents older libraries to work in a FF-A system
3. Convince Arm to update the specification with to allow more sharing the same page twice with all the new corner cases that brings plus adding extra complexity in Hypervisor and SPMC

Note that for non-FF-A systems there will not be any changes at all in the kernel/user space ABI.

If you can see a better way of dealing with this please let me know.

jforissier:
Well if you say it is too complex to deal with this in EL2 and/or EL1 and/or SEL2 and/or SEL1, then I think the best option is indeed this PR, preferably with a shared memory allocation function that makes a difference between FF-A and non-FF-A. Say, rename `paged_aligned_alloc()` to `shared_mem_alloc()` and enforce alignment only for FF-A. By doing so there would be zero impact on non FF-A systems, otherwise some memory can be wasted.
As for FF-A systems:
- Those that have never been non-FFA are fine of course
- Those transitioning from non-FF-A to FF-A will have to use the newer `libteec`...

jenswi-linaro:
I'm not so keen on adding a flag in the ABI or a config flags for this.

In tee-supplicant we have only short lived buffers with a minimum size normally not smaller that a page so `paged_aligned_alloc()` should be harmless there.

In `libteec` the problematic function is `TEEC_AllocateSharedMemory()`. The other case, allocate shadow buffer in `TEEC_RegisterSharedMemory()` we're already trying to recover.

For `TEEC_RegisterSharedMemory()` to become a pathological case it needs to be called to register lots of small buffers close to each other in memory or anywhere in read-only memory. I'm not sure that is a realistic use case.

How about letting `TEEC_AllocateSharedMemory()` try first with `malloc()` and if that fails to register try again with `teec_paged_aligned_alloc()`?

jforissier:
> How about letting `TEEC_AllocateSharedMemory()` try first with `malloc()` and if that fails to register try again with `teec_paged_aligned_alloc()`?

Yes, sounds good.

I would still prefer to see how the driver changes are accepted on the LKML before engaging in this direction.

jenswi-linaro:
It's still relevant. However, the driver changes are still under review.

jforissier:
@jenswi-linaro is there anything holding this off now?

PR #236, opened 2020-11-22
pkcs11: Add TEE Identity based authentication support

vesajaaskelainen:
Please see https://github.com/OP-TEE/optee_os/pull/4222 for testing and usage instructions.

Implements:
https://github.com/OP-TEE/optee_os/issues/3946

### DIFF
--- a/libckteec/src/invoke_ta.c
+++ b/libckteec/src/invoke_ta.c
@@ -39,7 +39,7 @@ bool ckteec_invoke_initiated(void)
 
 TEEC_SharedMemory *ckteec_alloc_shm(size_t size, enum ckteec_shm_dir dir)
 {
-	TEEC_SharedMemory *shm;
+	TEEC_SharedMemory *shm = NULL;
 
 	switch (dir) {
 	case CKTEEC_SHM_IN:
@@ -72,7 +72,7 @@ TEEC_SharedMemory *ckteec_alloc_shm(size_t size, enum ckteec_shm_dir dir)
 TEEC_SharedMemory *ckteec_register_shm(void *buffer, size_t size,
 				       enum ckteec_shm_dir dir)
 {
-	TEEC_SharedMemory *shm;
+	TEEC_SharedMemory *shm = NULL;
 
 	switch (dir) {
 	case CKTEEC_SHM_IN:
@@ -194,11 +194,11 @@ CK_RV ckteec_invoke_ta(unsigned long cmd, TEEC_SharedMemory *ctrl,
 
 static CK_RV ping_ta(void)
 {
-	TEEC_Operation op;
-	uint32_t origin;
-	TEEC_Result res;
-	uint32_t ta_version[3];
-	uint32_t status;
+	TEEC_Operation op = { 0 };
+	uint32_t origin = 0;
+	TEEC_Result res = TEEC_SUCCESS;
+	uint32_t ta_version[3] = { 0 };
+	uint32_t status = 0;
 
 	memset(&op, 0, sizeof(op));
 	op.params[0].tmpref.buffer = &status;
@@ -237,7 +237,7 @@ CK_RV ckteec_invoke_init(void)
 	uint32_t origin = 0;
 	TEEC_Result res = TEEC_SUCCESS;
 	CK_RV rv = CKR_CRYPTOKI_ALREADY_INITIALIZED;
-	int e;
+	int e = 0;
 
 	e = pthread_mutex_lock(&ta_ctx.init_mutex);
 	if (e) {
@@ -290,7 +290,7 @@ out:
 CK_RV ckteec_invoke_terminate(void)
 {
 	CK_RV rv = CKR_CRYPTOKI_NOT_INITIALIZED;
-	int e;
+	int e = 0;
 
 	e = pthread_mutex_lock(&ta_ctx.init_mutex);
 	if (e) {
--- a/libckteec/src/invoke_ta.c
+++ b/libckteec/src/invoke_ta.c
@@ -8,6 +8,7 @@
 #define BINARY_PREFIX		"ckteec"
 #endif
 
+#include <errno.h>
 #include <inttypes.h>
 #include <pkcs11.h>
 #include <pkcs11_ta.h>
@@ -237,7 +238,47 @@ CK_RV ckteec_invoke_init(void)
 	uint32_t origin = 0;
 	TEEC_Result res = TEEC_SUCCESS;
 	CK_RV rv = CKR_CRYPTOKI_ALREADY_INITIALIZED;
+	const char *login_type_env = NULL;
+	const char *login_gid_env = NULL;
+	uint32_t login_method = TEEC_LOGIN_PUBLIC;
+	void *login_data = NULL;
+	gid_t login_gid = 0;
+	unsigned long tmpconv = 0;
+	char *endp = NULL;
 	int e = 0;
+
+	login_type_env = getenv("CKTEEC_LOGIN_TYPE");
+
+	if (login_type_env) {
+		if (strcmp(login_type_env, "public") == 0) {
+			login_method = TEEC_LOGIN_PUBLIC;
+		} else if (strcmp(login_type_env, "user") == 0) {
+			login_method = TEEC_LOGIN_USER;
+		} else if (strcmp(login_type_env, "group") == 0) {
+			login_gid_env = getenv("CKTEEC_LOGIN_GID");
+			if (!login_gid_env || !strlen(login_gid_env)) {
+				EMSG("missing CKTEEC_LOGIN_GID");
+				rv = CKR_ARGUMENTS_BAD;
+				goto out;
+			}
+
+			login_method = TEEC_LOGIN_GROUP;
+			tmpconv = strtoul(login_gid_env, &endp, 10);
+			if (errno == ERANGE || tmpconv > (gid_t)-1 ||
+			    (login_gid_env + strlen(login_gid_env) != endp)) {
+				EMSG("failed to convert CKTEEC_LOGIN_GID");
+				rv = CKR_ARGUMENTS_BAD;
+				goto out;
+			}
+
+			login_gid = (gid_t)tmpconv;
+			login_data = &login_gid;
+		} else {
+			EMSG("invalid value for CKTEEC_LOGIN_TYPE");
+			rv = CKR_ARGUMENTS_BAD;
+			goto out;
+		}
+	}
 
 	e = pthread_mutex_lock(&ta_ctx.init_mutex);
 	if (e) {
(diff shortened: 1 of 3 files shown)


libckteec/src/invoke_ta.c
```diff
@@ -237,8 +238,49 @@ CK_RV ckteec_invoke_init(void)
...
+	void *login_data = NULL;
+	gid_t login_gid = 0;
+	unsigned long int tmpconv = 0;
+	char *endp = NULL;
 	int e;
 
+	login_type_env = getenv("CKTEEC_LOGIN_TYPE");
+
+	if (login_type_env) {
+		if (strcmp(login_type_env, "public") == 0) {
```
etienne-lms, line 253: maybe move this into a local function that outputs `login_method` & `login_data`.

etienne-lms (commented):
Only minor comments.
For both commits, prefer start log header line with `libckteec: ` and add `()` to function labels. Also add a description line between header line and S-o-b tag in commit "C_Login: Allow NULL_PTR PIN entry.". i.e.
```
libckteec: Allow NULL_PTR PIN entry in C_Login()

Allow C_Login() to transport a null PIN reference as when protected
authentication path is used.

```

vesajaaskelainen:
> Only minor comments.
> For both commits, prefer start log header line with `libckteec: ` and add `()` to function labels. Also add a description line between header line and S-o-b tag in commit "C_Login: Allow NULL_PTR PIN entry.". i.e.
> 
> ```
> libckteec: Allow NULL_PTR PIN entry in C_Login()
> 
> Allow C_Login() to transport a null PIN reference as when protected
> authentication path is used.
> 
> Signed-off-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>
> ```

OK

libckteec/src/invoke_ta.c
vesajaaskelainen: That could require additional alloc as one need to handle login_data somehow. Here it is just referenced for duration of the call.

If that is what is wanted I can of course do that.

libckteec/src/pkcs11_token.c
```diff
@@ -698,7 +698,7 @@ CK_RV ck_login(CK_SESSION_HANDLE session, CK_USER_TYPE user_type,
 	size_t ctrl_size = 0;
 	char *buf = NULL;
 
-	if (!pin)
+	if (!pin && pin_len)
```
vesajaaskelainen, line 701: Co-worker of mine noticed that this only allows empty string and not NULL_PTR (which should be internally the same).

Will fix this when I test it out with test app that NULL_PTR works too.

Internally they should be the same thru the API's as we do not have NULL passing concept.

vesajaaskelainen:
> > Only minor comments.
> > For both commits, prefer start log header line with `libckteec: ` and add `()` to function labels. Also add a description line between header line and S-o-b tag in commit "C_Login: Allow NULL_PTR PIN entry.". i.e.
> > ```
> > libckteec: Allow NULL_PTR PIN entry in C_Login()
> > 
> > Allow C_Login() to transport a null PIN reference as when protected
> > authentication path is used.
> > 
> > Signed-off-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>
> > ```
> 
> OK

Updated commit messages.

libckteec/src/invoke_ta.c
etienne-lms: You're right, it would need extra implementation to make it clean. Maybe when there will be other alternate paths to support.
Can keep as-is, at least as for me.

libckteec/src/invoke_ta.c
```diff
@@ -237,8 +238,49 @@ CK_RV ckteec_invoke_init(void)
...
+			login_gid_env = getenv("CKTEEC_LOGIN_GID");
+			if (!login_gid_env || !strlen(login_gid_env)) {
+				EMSG("missing CKTEEC_LOGIN_GID");
+				rv = CKR_ARGUMENTS_BAD;
+				goto out;
+			}
+
+			login_method = TEEC_LOGIN_GROUP;
+			tmpconv = strtoul(login_gid_env, &endp, 10);
+
```
etienne-lms, line 267: minor: can remove this empty line.

libckteec/src/invoke_ta.c
```diff
@@ -237,8 +238,49 @@ CK_RV ckteec_invoke_init(void)
...
 	TEEC_Result res = TEEC_SUCCESS;
 	CK_RV rv = CKR_CRYPTOKI_ALREADY_INITIALIZED;
+	const char *login_type_env = NULL;
+	const char *login_gid_env = NULL;
+	uint32_t login_method = TEEC_LOGIN_PUBLIC;
+	void *login_data = NULL;
+	gid_t login_gid = 0;
+	unsigned long int tmpconv = 0;
+	char *endp = NULL;
 	int e;
```
etienne-lms, line 248: By the way, could you initialize `e` to 0 here: `int e = 0;`?

vesajaaskelainen: Ok. Will do the same for other variables in separate commit.

libckteec/src/invoke_ta.c
vesajaaskelainen: OK

libckteec/src/invoke_ta.c
vesajaaskelainen: OK

vesajaaskelainen:
@etienne-lms I believe I have addressed all your comments.

Also tested (and fixed) PIN == NULL_PTR cases for methods that use PIN as input.

vesajaaskelainen:
I was thinking about could xtest be used here somehow but seems rather challenging as there is no API to reset the token back to initial state without performing clearing secure storage and reboot 😞. Also switching to different identities and somehow having common identities can be a bit of problem.

etienne-lms:
Maybe a debug command to reset a token even without the SO credentials?
The TA supports multi tokens, a token could be reserved for debug purpose, identified from its token label.

vesajaaskelainen:
> Maybe a debug command to reset a token even without the SO credentials?
> The TA supports multi tokens, a token could be reserved for debug purpose, identified from its token label.

If it would have some magic name then perhaps that could enable reset token feature. This magic name would just be something that normally would not be used.

What I would like to have is that we can run xtest (and other tests) on to be released TA binaries to verify that everything is still looking good and if automated tests pass then mark release gate for those TA's as tests has passed. Eg. so that the binary would be intact and no need to compile it differently.

etienne-lms:
Currently **xtest** lists the slots and picks the last last to run its tests.
We could have **xtest** to check the token label is "Test token for OP-TEE xtest        " so we can safely scratch all even without rights.

libckteec/src/invoke_ta.c
```diff
@@ -237,7 +238,47 @@ CK_RV ckteec_invoke_init(void)
...
+			if (!login_gid_env || !strlen(login_gid_env)) {
+				EMSG("missing CKTEEC_LOGIN_GID");
+				rv = CKR_ARGUMENTS_BAD;
+				goto out;
+			}
+
+			login_method = TEEC_LOGIN_GROUP;
+			tmpconv = strtoul(login_gid_env, &endp, 10);
+			if ((errno == ERANGE) || (tmpconv > (gid_t)-1) ||
+			    (login_gid_env + strlen(login_gid_env) != endp)) {
```
etienne-lms, line 268: CI/checkpatch issue report:
```
4e56bdb libckteec: Use environment variables to determine OP-TEE TA login
CHECK: Unnecessary parentheses around 'errno == ERANGE'
#68: FILE: libckteec/src/invoke_ta.c:267:
+			if ((errno == ERANGE) || (tmpconv > (gid_t)-1) ||
+			    (login_gid_env + strlen(login_gid_env) != endp)) {
total: 0 errors, 0 warnings, 1 checks, 63 lines checked
```
I.e.:
```c
			if (errno == ERANGE || tmpconv > (gid_t)-1 ||
			    (login_gid_env + strlen(login_gid_env) != endp)) {
				...
			}

**Review by etienne-lms — COMMENTED** [review]
@ruchi393, does this change look ok to you?
@vesajaaskelainen, LGTM (with checkpatch report addressed) but I wait for https://github.com/OP-TEE/optee_os/pull/4222 before posting my R-b tag.
I think this is a nice feature, thanks for sharing.

File: libckteec/src/invoke_ta.c (hunk 5 above)
Comment (vesajaaskelainen) [reply libckteec/src/invoke_ta.c:268]:
```shell
$ CHECKPATCH=../linux/scripts/checkpatch.pl ../optee_os/scripts/checkpatch.sh 0eaddfc570b557c39148ff2e7faccaba8e07d55a
Checking commit(s):
0eaddfc libckteec: Use environment variables to determine OP-TEE TA login
total: 0 errors, 0 warnings, 110 lines checked

"[PATCH] libckteec: Use environment variables to determine OP-TEE TA" has no obvious style problems and is ready for submission.
```

Where as linux == v5.10-rc1.

Then wondering the difference:
```
$ cp ../optee_os/scripts/checkpatch.sh scripts/
$ CHECKPATCH=../linux/scripts/checkpatch.pl scripts/checkpatch.sh 0eaddfc570b557c39148ff2e7faccaba8e07d55a
Checking commit(s):
0eaddfc libckteec: Use environment variables to determine OP-TEE TA login
CHECK: Unnecessary parentheses around 'errno == ERANGE'
#121: FILE: libckteec/src/invoke_ta.c:267:
+			if ((errno == ERANGE) || (tmpconv > (gid_t)-1) ||
+			    (login_gid_env + strlen(login_gid_env) != endp)) {

total: 0 errors, 0 warnings, 1 checks, 110 lines checked
```

Shall we copy the checkpatch.sh also to this repo?

vesajaaskelainen:
Just rebased on top of master.

libckteec/src/invoke_ta.c
```diff
@@ -194,11 +194,11 @@ CK_RV ckteec_invoke_ta(unsigned long cmd, TEEC_SharedMemory *ctrl,
 
 static CK_RV ping_ta(void)
 {
-	TEEC_Operation op;
-	uint32_t origin;
-	TEEC_Result res;
-	uint32_t ta_version[3];
-	uint32_t status;
+	TEEC_Operation op = { };
```
etienne-lms, line 198: s/`{ }`/`{ 0 }`/ 
optee_client uses `{ 0 }` as unified initializer to support more toolchains.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment on 1st commit.

It would be nice we have something in optee_test/host/xtest/pkcts11_1000.c to test a bit this but it may tricky to test invalid identities.

libckteec/src/invoke_ta.c
vesajaaskelainen: OK

vesajaaskelainen:
> `Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` with comment on 1st commit.

Thanks for the review!

Fixed and applied tags.

vesajaaskelainen:
> It would be nice we have something in optee_test/host/xtest/pkcts11_1000.c to test a bit this but it may tricky to test invalid identities.

I believe we need to improve the testing support for tokens in overall first and then we can try to add tests also for these. This is kinda integration testing not unit testing level problem as we would need to navigate to different identities in a way that people can still run xtest successfully.

I would prefer that we do these as separate things.

PR #237, opened 2020-11-30
libckteec: Allow 0 length input buffer  for update operations.

ruchi393:
If a valid input buffer with size 0 is passed in an update
operation like C_SignUpdate()/C_VerifyUpdate()/C_EncryptUpdate()/
C_DecryptUpdate(), current implementation return error
CKR_BAD_ARGUMENTS. This is actually not an error scenario.
Ideally, we should not perform any cryptographic operation
and return CKR_OK if other parameters are valid. Hence register
a shared memory buffer in such scenarios also and pass it to the TA.

Suggested-by: Etienne Carriere <etienne.carriere@linaro.org>


etienne-lms (commented):

jenswi-linaro:

ruchi393:
Tags applied

PR #238, opened 2020-12-08
tee-supplicant: add a new thread to trigger deferred works in OP-TEE

anisyanka:
optee core contains the deferred work framework, which
allows to add any function to queue of pending works
(see 'core/kernel/deferred_work.c' file in optee_os).

These works must be executed, when Rich OS has booted
and the tee-supplicant has been started.

The following cases might be the deferred works in OP-TEE:
 - Run some specific RPC as soon as possible, when RichOS will start;
 - Remove REE based key store or any other fs operations;
 - Save TEE boot process log

This commit adds new posix thread to trigger execution of
all deferred works in OP-TEE. The thread will terminate,
when all deferred works will be done or when all attempts
will end. Last case is possible, if TEE always returns errors.

Signed-off-by: Aleksandr Anisimov <a.anisimov@omprussia.ru>

### DIFF
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -26,6 +26,7 @@ set (SRC
 	src/tee_supp_fs.c
 	src/tee_supplicant.c
 	src/teec_ta_load.c
+	src/tee_deferred_work.c
 )
 
 if (CFG_GP_SOCKETS)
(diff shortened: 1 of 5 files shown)


anisyanka:
The feature related to [this](https://github.com/OP-TEE/optee_os/pull/4247) PR.

PR #239, opened 2020-12-08
tee-supplicant: add a framework for loadable plugins

anisyanka:
This framework makes the supplicant a bit more flexible
in terms of providing services. Any external TEE services
can be designed as a tee-supplicant plugin.
It makes it easy to:
 - add new features in the supplicant that aren't needed in upstream,
   e.g. Rich OS specific services;
 - sync upstream version with own fork;

To create a plugin developers should implement the
'struct plugin_method' and a bind function.
See public/tee_plugin_method.h file.
As an example this patch implements a 'syslog' plugin
and its Makefile. The plugin helps to log any
information from TEE to system journal.

The plugin framework is based on new RPC call - 'OPTEE_MSG_RPC_CMD_PLUGIN'.
This RPC is an unified interface between TEE and any plugins.
Every plugin has own name based on UUID. TEE has access to
plugins only by the name (UUID).

Every plugin should be placed into "/usr/lib/tee-supplicant/plugins/".
The supplicant opens and binds all plugins in the directory during
startup process using libdl. After this any requests to plugins
from TEE will be processed in the common RPC handler.

Signed-off-by: Aleksandr Anisimov <a.anisimov@omprussia.ru>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,6 +129,7 @@ distclean: clean
 
 copy_export: build
 	mkdir -p $(DESTDIR)$(SBINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
+	cp config.mk $(DESTDIR)/optee_client_config.mk
 	cp -a ${O}/libteec/libteec.so* $(DESTDIR)$(LIBDIR)
 	cp -a ${O}/libteec/libteec.a $(DESTDIR)$(LIBDIR)
 	cp ${O}/tee-supplicant/tee-supplicant $(DESTDIR)$(SBINDIR)
--- a/public/tee_plugin_method.h
+++ b/public/tee_plugin_method.h
@@ -1,0 +1,20 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2020, Open Mobile Platform LLC
+ */
+
+#ifndef TEE_PLUGIN_METHOD_H
+#define TEE_PLUGIN_METHOD_H
+
+#include <stddef.h>
+#include <tee_client_api.h>
+
+struct plugin_method {
+	const char *name; /* short friendly name of the plugin */
+	TEEC_UUID uuid;
+	TEEC_Result (*init)(void);
+	TEEC_Result (*invoke)(unsigned int cmd, unsigned int sub_cmd,
+			      void *data, size_t in_len, size_t *out_len);
+};
+
+#endif /* TEE_PLUGIN_METHOD_H */
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -14,6 +14,8 @@ set (CFG_TEE_CLIENT_LOAD_PATH "/lib" CACHE STRING "Location of libteec.so")
 set (CFG_TEE_FS_PARENT_PATH "/data/tee" CACHE STRING "Location of TEE filesystem (secure storage)")
 # FIXME: Why do we have if defined(CFG_GP_SOCKETS) && CFG_GP_SOCKETS == 1 in the c-file?
 set (CFG_GP_SOCKETS "1" CACHE STRING "Enable GlobalPlatform Socket API support")
+set (CFG_TEE_PLUGIN_LOAD_PATH "/usr/lib/tee-supplicant/plugins/" CACHE STRING "tee-supplicant's plugins path")
+set (CMAKE_INSTALL_RPATH "${CFG_TEE_PLUGIN_LOAD_PATH}")
 
 ################################################################################
 # Source files
@@ -26,6 +28,7 @@ set (SRC
 	src/tee_supp_fs.c
 	src/tee_supplicant.c
 	src/teec_ta_load.c
+	src/plugin.c
 )
 
 if (CFG_GP_SOCKETS)
@@ -49,6 +52,7 @@ target_compile_definitions (${PROJECT_NAME}
 	PRIVATE -DTEEC_LOAD_PATH="${CFG_TEE_CLIENT_LOAD_PATH}"
 	PRIVATE -DTEE_FS_PARENT_PATH="${CFG_TEE_FS_PARENT_PATH}"
 	PRIVATE -DBINARY_PREFIX="TSUP"
+	PRIVATE -DTEE_PLUGIN_LOAD_PATH="${CFG_TEE_PLUGIN_LOAD_PATH}"
 )
 
 ################################################################################
@@ -86,7 +90,8 @@ target_include_directories(${PROJECT_NAME} PRIVATE src)
 
 target_link_libraries (${PROJECT_NAME}
 	PRIVATE teec
-	PRIVATE optee-client-headers)
+	PRIVATE optee-client-headers
+	PRIVATE dl)
 
 ################################################################################
 # Install targets
--- a/tee-supplicant/Makefile
+++ b/tee-supplicant/Makefile
@@ -18,7 +18,8 @@ TEES_SRCS	:= tee_supplicant.c \
 		   teec_ta_load.c \
 		   tee_supp_fs.c \
 		   rpmb.c \
-		   handle.c
+		   handle.c \
+		   plugin.c
 
 
 ifeq ($(CFG_GP_SOCKETS),y)
@@ -43,7 +44,8 @@ TEES_CFLAGS	:= $(addprefix -I, $(TEES_INCLUDES)) $(CFLAGS) \
 		   -DDEBUGLEVEL_$(CFG_TEE_SUPP_LOG_LEVEL) \
 		   -DBINARY_PREFIX=\"TEES\" \
 		   -DTEE_FS_PARENT_PATH=\"$(CFG_TEE_FS_PARENT_PATH)\" \
-		   -DTEEC_LOAD_PATH=\"$(CFG_TEE_CLIENT_LOAD_PATH)\"
+		   -DTEEC_LOAD_PATH=\"$(CFG_TEE_CLIENT_LOAD_PATH)\" \
+		   -DTEE_PLUGIN_LOAD_PATH=\"$(CFG_TEE_PLUGIN_LOAD_PATH)\"
 
 ifeq ($(CFG_GP_SOCKETS),y)
 TEES_CFLAGS	+= -DCFG_GP_SOCKETS=1
@@ -68,6 +70,10 @@ endif
 TEES_LFLAGS	+= -lpthread
 # Needed to get clock_gettime() for for glibc versions before 2.17
 TEES_LFLAGS	+= -lrt
+# Needed to dynamically load user plugins
+TEES_LFLAGS	+= -ldl
+# Needed for dlopen()
+TEES_LFLAGS += -Wl,-rpath=$(CFG_TEE_PLUGIN_LOAD_PATH)
 
 tee-supplicant: $(TEES_FILE)
 
--- a/tee-supplicant/src/optee_msg_supplicant.h
+++ b/tee-supplicant/src/optee_msg_supplicant.h
@@ -183,6 +183,10 @@
  */
 #define OPTEE_MSG_RPC_CMD_FTRACE	11
 
+/*
+ * Plugin commands
+ */
+#define OPTEE_MSG_RPC_CMD_PLUGIN	12
 
 /*
  * Define protocol for messages with .cmd == OPTEE_MSG_RPC_CMD_SOCKET
@@ -257,6 +261,40 @@
 
 /*
  * End of definitions for messages with .cmd == OPTEE_MSG_RPC_CMD_SOCKET
+ */
+
+/*
+ * Define protocol for messages with .cmd == OPTEE_MSG_RPC_CMD_PLUGIN
+ */
+
+/*
+ * Invoke a tee-supplicant plugin.
+ *
+ * [in]     param[0].u.value.a	OPTEE_INVOKE_PLUGIN
+ * [in]     param[0].u.value.b	uuid.d1
+ * [in]     param[0].u.value.c	uuid.d2
+ * [in]     param[1].u.value.a	uuid.d3
+ * [in]     param[1].u.value.b	uuid.d4
+ * [in]     param[1].u.value.c	cmd for plugin
+ * [in]     param[2].u.value.a	sub_cmd for plugin
+ * [out]    param[2].u.value.b  length of the outbuf (param[3]),
+ *                              if out is needed.
+ * [in/out] param[3].u.tmem	buffer holding data for plugin
+ *
+ * UUID serialized into octets:
+ * b0  b1  b2  b3   b4  b5  b6  b7   b8  b9  b10  b11   b12  b13  b14  b15
+ *       d1       |       d2       |        d3        |         d4
+ *
+ * The endianness of words d1, d2, d3 and d4 from SWd is little-endian.
+ * d1 word contains [b3 b2 b1 b0]
+ * d2 word contains [b7 b6 b5 b4]
+ * d3 word contains [b11 b10 b9 b8]
+ * d4 word contains [b15 b14 b13 b12]
+ */
+#define OPTEE_INVOKE_PLUGIN	0
+
+/*
+ * End of definitions for messages with .cmd == OPTEE_MSG_RPC_CMD_PLUGIN
  */
 
 #endif /*__OPTEE_MSG_SUPPLICANT_H*/
--- a/tee-supplicant/src/plugin.c
+++ b/tee-supplicant/src/plugin.c
@@ -1,0 +1,221 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) 2020, Open Mobile Platform LLC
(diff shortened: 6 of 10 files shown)


anisyanka:
The PR relates to:
optee backend: https://github.com/OP-TEE/optee_os/pull/4248
example of using: https://github.com/linaro-swg/optee_examples/pull/79

jenswi-linaro:
Why doesn't tee-supplicant read from the ringbuffer directly?
Assuming that this relates to https://github.com/OP-TEE/optee_os/issues/4230
I have a feeling that this will not be very suitable for the logging framework since we may want to use all kinds of tricks for efficiency.

anisyanka:
> Why doesn't tee-supplicant read from the ringbuffer directly?
> Assuming that this relates to [OP-TEE/optee_os#4230](https://github.com/OP-TEE/optee_os/issues/4230)
> I have a feeling that this will not be very suitable for the logging framework since we may want to use all kinds of tricks for efficiency.

Sorry, if I confused you. I have not started to code the logging framework discussed in [OP-TEE/optee_os#4230](https://github.com/OP-TEE/optee_os/issues/4230) yet.

This PR does not apply to the logger, because it is only about plugins.
The plugins are needed just to do the tee-supplicant a bit more flexible in terms of providing services.
In our system there are several OS specific services, which are needed from optee code or TA code.
For example os specific logging mechanisms (not the syslog) and entropy gathering service.
Of course, we can add new RPCs to optee and the supplicant and always modify its code for working with our services,
but, seems, it looks overkill. I think the other optee users may have the same experience.
That's why we proposed to use plugins like an universal framework to add new specific feature to the supplicant.

As an example I added a plugin named `syslog` (file `tee-supplicant/plugins/syslog/syslog_plugin.c`).
We can consider the 'syslog' plugin as a proof-of-concept that the plugin framework works.
Also It maybe useful if you want to have a simple logger from TAs code already now,
but this plugin is not what we discussed, when we talked about the logger :)

tee-supplicant/Makefile
```diff
@@ -81,9 +88,13 @@ $(TEES_OBJ_DIR)/%.o: $(TEES_SRC_DIR)/%.c
 	@echo "  CC      $<"
 	$(VPREFIX)$(CC) $(TEES_CFLAGS) $(TEES_CFLAGS_$(notdir $<)) -c $< -o $@
 
+plugins:
```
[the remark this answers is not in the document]
anisyanka: Ok, I agree.
I will move the `syslog` plugin to `optee_examples` and implement tests for `xtest`.

anisyanka:
All comments have been addressed

tee-supplicant/src/plugin.c
```diff
@@ -0,0 +1,211 @@
...
+				 size_t *out_len)
+{
+	int res;
+	struct plugin *p;
+
+	p = find_plugin(u);
+	if (!p)
+		return TEEC_ERROR_ITEM_NOT_FOUND;
+
+	if (!p->method->invoke)
```
jenswi-linaro, line 96: This "can't happen" since it was tested in `load_plugin()` above. An `assert()` should be enough.

public/tee_plugin_method.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef __TEE_PLUGIN_METHOD_H
+#define __TEE_PLUGIN_METHOD_H
+
+#include <stddef.h>
+#include <tee_client_api.h>
+
+struct plugin_method {
+	const char *name; /* short friendly name of the plugin */
+	TEEC_UUID uuid;
+	int (*init)(void);
```
jenswi-linaro, line 15: Any reason why we're not returning a `TEEC_Result` instead here and for `invoke()` below?

tee-supplicant/src/plugin.c
```diff
@@ -0,0 +1,211 @@
...
+				break;
+			}
+
+			/* Init the plugin */
+			if (p->method->init) {
+				res = p->method->init();
+				if (res != 0) {
+					free(p);
+					EMSG("init the <%s> plugin failed with orig = %d",
+					     p->method->name, res);
```
jenswi-linaro, line 156: Use after free of `p`.

tee-supplicant/src/optee_msg_supplicant.h
```diff
@@ -259,4 +263,28 @@
...
 
+/*
+ * Define protocol for messages with .cmd == OPTEE_MSG_RPC_CMD_PLUGIN
+ */
+
+/*
+ * Invoke a tee-supplicant plugin.
+ *
+ * [in]     param[0].u.value.a	OPTEE_INVOKE_PLUGIN
+ * [in]     param[0].u.value.b	uuid.d1
```
jenswi-linaro, line 274: We need to define byte order of the UUID. Consider the case when Secure world is little endian but Normal world is Big endian. These value words will then be swapped (in the driver, not supported now by the way) before transmission.

I think this should to be defined as in https://developer.arm.com/documentation/den0028/c/ section "5.3 Unique Identification format".

tee-supplicant/src/plugin.c
anisyanka: thanks

public/tee_plugin_method.h
anisyanka: I think we can replace `int` to ``TEEC_Result``

anisyanka:
All comments have been addressed

tee-supplicant/src/plugin.c
```diff
@@ -42,6 +43,14 @@ static int uuid_cmp(TEEC_UUID *u1, TEEC_UUID *u2)
 	return 1;
 }
 
+static void uuid_from_octets(TEEC_UUID *d, const uint8_t s[TEE_IOCTL_UUID_LEN])
+{
+	d->timeLow = (s[0] << 24) | (s[1] << 16) | (s[2] << 8) | s[3];
```
jenswi-linaro, line 48: Does work (will the `uint8_t` automatically be promoted to `unsigned int` here)?
It seems more robust to either use a helper macro like `SHIFT_U32()` in https://github.com/OP-TEE/optee_os/blob/b68aca61f70193c8715a49748d13c77afe43aa1c/lib/libutils/ext/include/util.h#L117

public/tee_plugin_method.h
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2020, Open Mobile Platform LLC
+ */
+
+#ifndef __TEE_PLUGIN_METHOD_H
```
jenswi-linaro, line 6: Please drop the `__` prefix. Those are reserved for libc and friends in user space.
We're using it in some files already, but let's not make matters worse.

tee-supplicant/src/plugin.c
```diff
@@ -0,0 +1,221 @@
...
+TEEC_Result plugin_load_all(void)
+{
+	DIR *dir = NULL;
+	int res = PLUGIN_OK;
+	TEEC_Result teec_res = TEEC_SUCCESS;
+	struct dirent *entry = NULL;
+
+	dir = opendir(TEE_PLUGIN_LOAD_PATH);
+	if (!dir) {
+		EMSG("could not open directory %s", TEE_PLUGIN_LOAD_PATH);
```
jenswi-linaro, line 118: I suppose this isn't an error, please use `IMSG()` instead.

tee-supplicant/src/plugin.h
```diff
@@ -0,0 +1,34 @@
...
+#include <stdint.h>
+#include <tee_client_api.h>
+#include <tee_plugin_method.h>
+
+struct tee_ioctl_param;
+
+/* This structure describes one plugin for the supplicant */
+struct plugin {
+	void *handle;
+	struct plugin_method *method; /* implemented by plugins' authors */
```
jenswi-linaro, line 18: Implemented in the plugin

tee-supplicant/src/tee_supplicant.c
```diff
@@ -711,6 +715,11 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	if (plugin_load_all() != 0) {
+		EMSG("failed to load user's tee plugins");
```
jenswi-linaro, line 719: Failed to load plugins.

tee-supplicant/src/plugin.h
```diff
@@ -0,0 +1,34 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2020, Open Mobile Platform LLC
+ */
+
+#ifndef __PLUGIN_H
```
jenswi-linaro, line 6: Please drop the `__` prefix.

tee-supplicant/src/plugin.c
anisyanka: I agree.
It seems we need common `util.h` file in `optee-client` project.
We can put into one all helpers which already exists in the tee_supplicant.c file (e.g. `uuid_from_octets()`), in the libcktee (file `local_utils.h`) and also `SHIFT_U32()`. But it should be done in separate MR.

So far I did the casting each `s[x]` to `uint32_t`.

anisyanka:
All comments have been addressed

tee-supplicant/src/plugin.c
```diff
@@ -0,0 +1,222 @@
...
+		    TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INPUT ||
+	    (params[1].attr & TEE_IOCTL_PARAM_ATTR_TYPE_MASK) !=
+		    TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INPUT ||
+	    (params[2].attr & TEE_IOCTL_PARAM_ATTR_TYPE_MASK) !=
+		    TEE_IOCTL_PARAM_ATTR_TYPE_VALUE_INOUT ||
+	    (params[3].attr & TEE_IOCTL_PARAM_ATTR_TYPE_MASK) !=
+		    TEE_IOCTL_PARAM_ATTR_TYPE_MEMREF_INOUT)
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	uuid_words[0] = le32toh(params[0].b);
```
jenswi-linaro, line 196: I'm sorry I was wrong about this swapping to make it little. I Recently got clarifications from Arm on this.
See https://www.spinics.net/lists/arm-kernel/msg868040.html
So please drop the `le32toh()` and we should be good.

tee-supplicant/src/plugin.c
```diff
@@ -0,0 +1,222 @@
...
+
+#include "optee_msg_supplicant.h"
+
+#ifndef __aligned
+#define __aligned(x) __attribute__((__aligned__(x)))
+#endif
+#include <linux/tee.h>
+
+/* internal possible returned values */
+enum {
```
jenswi-linaro, line 29: Please give a name to this enum so that we can use the correct type when it's used.

tee-supplicant/src/plugin.c
anisyanka: Thank you for the clarification

tee-supplicant/src/plugin.c
anisyanka: done

anisyanka:
All comments have been addressed

Makefile
```diff
@@ -128,11 +130,15 @@ checkpatch-all-files: checkpatch-pre-req
 distclean: clean
 
 copy_export: build
-	mkdir -p $(DESTDIR)$(SBINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
+	mkdir -p $(DESTDIR)$(SBINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR) $(DESTDIR)$(PLUGINDIR)
+	mkdir -p $(MKDIR)
+	cp config.mk $(MKDIR)
```
jenswi-linaro, line 135: When building with buildroot for ARMv7 `$(MKDIR)` seems to become `host/arm-buildroot-linux-gnueabihf/sysroot/mk`. This is far too likely to clash with other packages. I'd prefer something like:
``` shell
cp config.mk $(DESTDIR)/optee_client_config.mk
```

Makefile
```diff
@@ -128,11 +130,15 @@ checkpatch-all-files: checkpatch-pre-req
...
 	cp -a ${O}/libteec/libteec.so* $(DESTDIR)$(LIBDIR)
 	cp -a ${O}/libteec/libteec.a $(DESTDIR)$(LIBDIR)
 	cp ${O}/tee-supplicant/tee-supplicant $(DESTDIR)$(SBINDIR)
 	cp public/*.h $(DESTDIR)$(INCLUDEDIR)
 	cp libckteec/include/*.h $(DESTDIR)$(INCLUDEDIR)
 	cp -a ${O}/libckteec/libckteec.so* $(DESTDIR)$(LIBDIR)
 	cp -a ${O}/libckteec/libckteec.a $(DESTDIR)$(LIBDIR)
+	$(eval SUPP_PLUGINS := $(shell find ./tee-supplicant/plugins -name *.plugin))
```
jenswi-linaro, line 143: We do we need this when we don't have any plugins in this git?

anisyanka: I think with these lines the plugin framework looks complete and ready to using.
I'd prefer keep ones here in case if someone will add own plugins in own fork.
In this case they won't have to edit the Makefile to build the plugins. 

And also, these lines don't generate errors during building, if there are no plugins.

jenswi-linaro: I hope we can agree that this is dead code here in upstream.

What happens downstream we can control, perhaps using a separate git for plugins is the preferred solution. We're demonstrating that in https://github.com/linaro-swg/optee_examples/pull/79 so it seems at bit strange to promote another approach here.

anisyanka:
All comments have been addressed.

Makefile
anisyanka: You're right.
I'll drop these lines

tee-supplicant/Makefile
```diff
@@ -68,8 +71,12 @@ endif
 TEES_LFLAGS	+= -lpthread
 # Needed to get clock_gettime() for for glibc versions before 2.17
 TEES_LFLAGS	+= -lrt
+# Needed to dynamically load user plugins
+TEES_LFLAGS	+= -ldl
+# Needed for dlopen()
+TEES_LFLAGS += -Wl,-rpath=$(CFG_TEE_PLUGIN_LOAD_PATH)
 
-tee-supplicant: $(TEES_FILE)
+tee-supplicant: $(TEES_FILE) plugins
```
jenswi-linaro, line 79: I don't suppose we need the `plugins` target any longer. Please remove it here and elsewhere.

tee-supplicant/plugins/Makefile
```diff
@@ -0,0 +1,11 @@
+# Add new plugin dirs here
```
jenswi-linaro, line 1: This file together with the `tee-supplicant/plugins` directory can be removed.

tee-supplicant/CMakeLists.txt
```diff
@@ -86,7 +90,8 @@ target_include_directories(${PROJECT_NAME} PRIVATE src)
 
 target_link_libraries (${PROJECT_NAME}
 	PRIVATE teec
-	PRIVATE optee-client-headers)
+	PRIVATE optee-client-headers
+	PRIVATE dl)
```
jenswi-linaro, line 94: Is this how the `-ldl` parameter get added when linking?
How is the `-Wl,-rpath=$(CFG_TEE_PLUGIN_LOAD_PATH)` parameter added?

anisyanka: > Is this how the -ldl parameter get added when linking?

Yes, it is.

> How is the -Wl,-rpath=$(CFG_TEE_PLUGIN_LOAD_PATH) parameter added?

I'have used [CMAKE_INSTALL_RPATH](https://github.com/OP-TEE/optee_client/pull/239/files/52ed2ea5b6b36a911c90ab98a133caa27722ce53#diff-2498ed101052b351eb80970e8034c349068084c52046fe269a2c60608a61fbfdR18). It is a variable embedded in cmake.
If I understood cmake [specification](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_RPATH.html) right, the variable is needed to set a run-time search path.

tee-supplicant/plugins/Makefile
anisyanka: Done.

tee-supplicant/CMakeLists.txt
jenswi-linaro: OK, thanks

jenswi-linaro:
Please squash in the fixup commits

jenswi-linaro:
I'm sorry I didn't see the update before. It seems that github sometimes doesn't notify of a force-push. The safest is to add a comment once a force-push has been performed. Just a plain "Update" is good enough.

jenswi-linaro (commented):

anisyanka:
Updated

ruchi393:
@anisyanka , can you please add documentation related with this feature by raising a PR in https://github.com/OP-TEE/optee_docs.

anisyanka:
@ruchi393, yes, good idea, thanks!

PR #250, opened 2021-01-20
libckteec: serialize: fix check for ULONG type entries

ruchi393:
It is fine for callee to allocate more memory than what is
needed for any attribute. Earlier for attributes of type
ulong, a strict check was there which returned error
if callee passed buffer greater than the required size.


etienne-lms:

ruchi393:
Tag applied

PR #252, opened 2021-01-22
libckteec: serialize: Fix crash with NULL pValue in attribute

ruchi393:
An attribute template (type, pValue, ulValueLen) provided with
C_GetAttributeValue() may have pValue field as NULL pointer with
a zero/non-zero ulValueLen field. In these cases TEE should modify
the ulValueLen to hold the exact length of the specified attribute
for the object.

In the current code, for non-zero ulValueLen with pValue as NULL
pointer, an attempt was being made to read the ulValueLen length
data from NULL pointer resulting in segmentation fault. To fix this,
in such scenarios, pass ulValueLen as 0 to TEE.

### DIFF
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -195,7 +195,11 @@ static CK_RV serialize_ck_attribute(struct serializer *obj, CK_ATTRIBUTE *attr)
 			pkcs11_size = sizeof(uint32_t);
 		} else {
 			pkcs11_pdata = attr->pValue;
-			pkcs11_size = attr->ulValueLen;
+			/* Support NULL data pointer with non-zero size */
+			if (!pkcs11_pdata)
+				pkcs11_size = 0;
+			else
+				pkcs11_size = attr->ulValueLen;
 		}
 		break;
 	}


libckteec/src/serialize_ck.c
```diff
@@ -164,7 +164,10 @@ static CK_RV serialize_ck_attribute(struct serializer *obj, CK_ATTRIBUTE *attr)
 			pkcs11_size = sizeof(uint32_t);
 		} else {
 			pkcs11_pdata = attr->pValue;
-			pkcs11_size = attr->ulValueLen;
+			if (!pkcs11_pdata)
+				pkcs11_size = 0;
```
etienne-lms, line 168: Is it a valid scenario to have pValue NULL and ulValueLen from caller client?
Shouldn't this return with CKR_TEMPLATE_INCONSISTENT?

etienne-lms: For example, in `case CKA_ALLOWED_MECHANISMS:`, few lines above,
we access `*attr->pValue` (see `xxx = type[m];`) without checking `pValue` is not NULL. 
Since `attr->ulValueLen` > 0, we assume `pValue` is valid. If it is not, it is a programming error and we segfault. Should we also return without faulting is this case?

ruchi393: > Is it a valid scenario to have pValue NULL and ulValueLen from caller client?
> Shouldn't this return with CKR_TEMPLATE_INCONSISTENT

Ideally, if someone is passing ulValuelen as non-zero, pValue should not be NULL. PKCS#11 specification is not very clear on this. See C_GetAttributesValue(), point 3. It mentions about pValue NULL pointer case. 
Segfault may not be the right thing here. We should either return error or pass NULL ptr with size 0 to TEE and let it handle it.

The trigger for this fix came after I got a segfault in the SoftHSM testcase
p11test ObjectTests::testSetAttributeValue

See the code at lines 135/1358
https://github.com/opendnssec/SoftHSMv2/blob/develop/src/lib/test/ObjectTests.cpp

We can make a call, 
either we return an error from here or pass a NULL ptr with size 0 to TEE and let it do the needful. For C_GetAttributeValue(), it would return the required size. For other scenarios, my guess is attribute should get added with NULL value.

vesajaaskelainen: > Is it a valid scenario to have pValue NULL and ulValueLen from caller client?
> Shouldn't this return with CKR_TEMPLATE_INCONSISTENT?

One can query value size with pValue == NULL and then allocate memory and then have pValeu != NULL and right size.

vesajaaskelainen: I suppose this change is OK as we do not have NULL_PTR (pointer) passing in the interface as such so that is simulated with size being zero.

etienne-lms: Ok. I agree this change is ok., 
Maybe add a comment: /* Support malformed null sized value with non-null pointer */

ruchi393: Sure, I will add the comment

etienne-lms:
The change looks good to me.
However pkcs11 regression test fails. I tested back master w/ and w/o this P-R applied but I can't reproduce the failure:

> (snip)
  3641:  o pkcs11_1012.4 Get Attribute - value with larger buffer
  3642:  pkcs11_1000.c:2786: Expression "get_attr_template_vl[0].ulValueLen == sizeof(obj_value)" (128 == 5) is false
  3643:    pkcs11_1012.4 FAILED
  3644:  o pkcs11_1012.5 Get Attribute - bool + value with exact size + class + bool
  3645:    pkcs11_1012.5 OK
  3646:  o pkcs11_1012.6 Get Attribute - bool + value with larger buffer + class + bool
  3647:  Segmentation fault

ruchi393:
> The change looks good to me.
> However pkcs11 regression test fails. I tested back master w/ and w/o this P-R applied but I can't reproduce the failure:
> 
> > (snip)
> > 3641:  o pkcs11_1012.4 Get Attribute - value with larger buffer
> > 3642:  pkcs11_1000.c:2786: Expression "get_attr_template_vl[0].ulValueLen == sizeof(obj_value)" (128 == 5) is false
> > 3643:    pkcs11_1012.4 FAILED
> > 3644:  o pkcs11_1012.5 Get Attribute - bool + value with exact size + class + bool
> > 3645:    pkcs11_1012.5 OK
> > 3646:  o pkcs11_1012.6 Get Attribute - bool + value with larger buffer + class + bool
> > 3647:  Segmentation fault

Not reproducible at my end also. I will squash commits and re-push to trigger the IBART again to see if failure still comes up.

ruchi393:
IBART passes now

libckteec/src/serialize_ck.c
vesajaaskelainen: I am trying to understand the text "Support malformed null sized value with non-null pointer".

Somehow that is confusing.

"Support NULL data pointer with non-zero size"

etienne-lms:
it looks like this is stalled but the change is valuable.

ruchi393:
Thanks @etienne-lms . I totally forgot about this PR. I  have no changes pending from my end on this PR.
Are you ok with the PR , any comments ?

libckteec/src/serialize_ck.c
```diff
@@ -195,7 +195,14 @@ static CK_RV serialize_ck_attribute(struct serializer *obj, CK_ATTRIBUTE *attr)
 			pkcs11_size = sizeof(uint32_t);
 		} else {
 			pkcs11_pdata = attr->pValue;
-			pkcs11_size = attr->ulValueLen;
+			/*
+			 * Support malformed null sized value with non-null
+			 * pointer
+			 */
```
etienne-lms, line 201: @vesajaaskelainen  suggested to simplify the sentence:
`/* Support NULL data pointer with non-zero size */`

etienne-lms (commented):
there is a pending comment not addressed.

Once done, 
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>`.

libckteec/src/serialize_ck.c
ruchi393: Done

ruchi393:
Thanks. Comment fixed. Tag applied.

vesajaaskelainen:
`Reviewed-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

ruchi393:
Tags applied.

etienne-lms (approved):

PR #254, opened 2021-02-09
tee-supplicant: daemonize before opening a supplicant device

parport0:
Prior to this change tee-supplicant, when called with -d,
first opened /dev/teeprivX, and called daemon(3) later.
This implied that a file descriptor associated with the
supplicant device transferred from one process to another.

This may conflict with security policies related to tee device
file descriptors. Since TEE client ID is tied to a process,
transferring such a sensitive thing as a tee device handle
between processes might be discouraged.

Reorder the two: first daemonize, and open the tee device later.

---

An example of how such a security policy could be implemented can be seen at https://github.com/parport0/linux/tree/tee-fd.

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -696,6 +696,11 @@ int main(int argc, char *argv[])
 			dev = argv[i];
 	}
 
+	if (daemonize && daemon(0, 0) < 0) {
+		EMSG("daemon(): %s", strerror(errno));
+		exit(EXIT_FAILURE);
+	}
+
 	if (dev) {
 		arg.fd = open_dev(dev, &arg.gen_caps);
 		if (arg.fd < 0) {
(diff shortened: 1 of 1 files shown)


jforissier:
Hi @parport0 ,

Thanks for the patch. This looks good to me, although I initially had some doubts it might introduce a race condition. Let me explain so we can figure out if everything is OK.

- Currently when the command `tee-supplicant -d` returns, the `/dev/teeprivXXX` device has been opened already and therefore the TEE driver considers that `tee-supplicant` is ready to process requests. A client application can therefore be started immediately and invoke a TA (which will typically cause some RPC to `tee-supplicant`).
- With this patch, the TEE device is not yet opened when `tee-supplicant -d` returns.
  - Therefore it could happen (with low probability) that supplicant is started, then a client application starts, invokes a TA, then the TEE core will perform a RPC and the TEE driver will find that no supplicant is attached yet. Luckily RPC requests done on behalf of user space are blocking so things will just stall until the child process in `tee-supplicant` is ready. So this case sounds OK.
  - On the other hand, the kernel client API (`tee_client_open_context()`) is non-blocking by default. See `supp_nowait` in the driver. Could this be a problem? (I don't think so)
  - Finally, could a user space client request non-blocking calls? Not with the upstream driver it seems.

I suppose we could put in place some IPC mechanism between the parent and the child process so that the parent would exit only after the child has opened the device, but I'd rather avoid complications if it's not needed of course.

@jenswi-linaro what do you think?

jenswi-linaro:
Looks good to me. Thanks for sharing your thoughts on this. I agree, shouldn't be a problem.

jforissier:
Thanks Jens, so this is also:

parport0:
Thank you.
I have added the `Reviewed-by:` tags.

PR #255, opened 2021-02-09
tee-supplicant: rpmb: switch to MMC_IOC_MULTI_CMD

ricardosalveti:
On some eMMC devices, the mix of RPMB and normal MMC commands can result
in GENERAL FAILURE, which then forces OP-TEE to mark the RPMB device
as dead (part of the rollback vulnerability fix).

In order to work around this problem, the linux kernel also supports
sending all the required MMC_IOC commands in one single multiple ioctl
command request, which forces the RPMB commands to be serialized,
avoiding the issue.

As MMC_IOC_MULTI_CMD is available since linux 4.3, we should be safe to
use it by default on read / write block operations.

For reference, this issue is also nicely described at the
https://www.mail-archive.com/linux-mmc@vger.kernel.org/msg17589.html.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -134,6 +134,9 @@ static pthread_mutex_t rpmb_mutex = PTHREAD_MUTEX_INITIALIZER;
 /* mmc_ioc_cmd.write_flag */
 #define MMC_CMD23_ARG_REL_WR	(1 << 31) /* CMD23 reliable write */
 
+/* Maximum number of commands used in a multiple ioc command request */
+#define RPMB_MAX_IOC_MULTI_CMDS		3
+
 #ifndef RPMB_EMU
 
 #define IOCTL(fd, request, ...)					   \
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Was focused on the actual hardware that I forgot about the rpmb emulation piece, and that will require updates as well.

jenswi-linaro:
> Was so focused on the actual hardware that I forgot about the rpmb emulation piece, and that will require updates as well.

OK, please add a comment when that's fixed so we don't miss the update.

igoropaniuk:
It turned out that this patch also fixes RPMB access issues on iMX8MM (altought the initial problem was that on iMX6ULL it was causing the eMMC controller to crash completely). Basically operation result for the last rpmb frame  in` tee_rpmb_resp_unpack_verify()`  for all RPMB requests was always `RPMB_RESULT_GENERAL_FAILURE`, 
With this patch all good now.

`Tested-by: Igor Opaniuk <igor.opaniuk@foundries.io>`

ricardosalveti:
Great, got everything needed to use emu locally and will try to get back to this patch tomorrow.

ricardosalveti:
PR is finally now able to be properly reviewed.

PR #263, opened 2021-03-10
libckteec: return CKR_CANT_LOCK if mutex fails at lib initialization

etienne-lms:
Return CKR_CANT_LOCK if mutex fails at lib initialization. This
conforms with the PKCS#11 specification v2.40-errata01. Preserve
the exit cases on mutex issues when unlocking and in C_Finalize().


ruchi393:
`Acked-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

etienne-lms:
tag applied

PR #266, opened 2021-03-26
Allocate page aligned shared memory buffers

jenswi-linaro:
Allocate page aligned shared memory buffer guarantee that each shared
memory buffer doesn't accidentally share a page of memory or they may
become aliased when mapped in secure world. This is normally not a big
problem but may make it a bit harder to track down buffer overruns in
shared memory buffers.

In a post Arm v8.4 architecture with FF-A [1] there's trouble since it's
not permitted to share the same physical page twice.

[1] https://developer.arm.com/documentation/den0077/latest

Needed by https://github.com/OP-TEE/optee_os/pull/4509

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -74,6 +74,16 @@ static void teec_mutex_unlock(pthread_mutex_t *mu)
 	pthread_mutex_unlock(mu);
 }
 
+static void *teec_paged_aligned_alloc(size_t sz)
+{
+	void *p = NULL;
+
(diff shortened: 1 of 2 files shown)


jforissier:

etienne-lms (commented):

alexandruavadanii:
Sorry for commenting on an old PR, but is `sysconf(_SC_PAGESIZE)` the best granularity in `posix_memalign(&p, sysconf(_SC_PAGESIZE), sz)`?
I would have expected the same 4K pagesize used by `optee_os`, just to reduce fragmentation if REE OS pagesize is 64KB. Or am I wrong?

PR #267, opened 2021-04-27
Makefile: install optee_client_config.mk at INCLUDEDIR

ricardosalveti:
optee_client_config.mk should be installed to a more traditional and
common folder such as INCLUDEDIR instead of being installed directly to
DESTDIR (which can end up as / depending on the packaging system used).

This was noticed when packaging optee-client for OE, as DESTDIR gets set
to the deploy folder, which is later used as root for packaging
optee-client.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ distclean: clean
 
 copy_export: build
 	mkdir -p $(DESTDIR)$(SBINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
-	cp config.mk $(DESTDIR)/optee_client_config.mk
+	cp config.mk $(DESTDIR)/$(INCLUDEDIR)/optee_client_config.mk
 	cp -a ${O}/libteec/libteec.so* $(DESTDIR)$(LIBDIR)
 	cp -a ${O}/libteec/libteec.a $(DESTDIR)$(LIBDIR)
 	cp ${O}/tee-supplicant/tee-supplicant $(DESTDIR)$(SBINDIR)


ricardosalveti:
This will require changes at optee-test and optee-examples as well, as the default path used for finding optee_client_config.mk requires a similar change.

ricardosalveti:
Include dir might not necessarily be the most correct folder, but is at least somehow aligned with optee-os. Open for suggestions.

Makefile
```diff
@@ -129,7 +129,7 @@ distclean: clean
 
 copy_export: build
 	mkdir -p $(DESTDIR)$(SBINDIR) $(DESTDIR)$(LIBDIR) $(DESTDIR)$(INCLUDEDIR)
-	cp config.mk $(DESTDIR)/optee_client_config.mk
+	cp config.mk $(DESTDIR)/$(INCLUDEDIR)/optee_client_config.mk
```
jforissier, line 132: Who uses `optee_client_config.mk`? I suggest simply removing this line.

ricardosalveti: It's used by optee-test and examples, such as https://github.com/OP-TEE/optee_test/blob/master/Makefile#L12 (for the tee-supplicant plugin stuff).

jforissier: Oh, I totally missed that :-/

The bad new is, this is broken already because of this: https://github.com/OP-TEE/optee_test/blob/master/host/supp_plugin/CMakeLists.txt#L3. Makefile is not used in the defaut OP-TEE build environment which uses Buildroot and CMake...

That being said, OP-TEE produces `conf.mk` and `conf.cmake` which describe the OS/TA config and puts them in the `host_include` folder of the "TA dev kit". So it would seem consistent to do a similar thing for `optee_client`, i.e.:
1. Use the "include" directory (as this patch does)
2. Also install a new `optee_client_config.cmake` (TBD, I think we should open a bug for that)

Therefore...

ricardosalveti: Yeah, seems this is only handled and available when Makefile is used, will open a bug for the cmake piece.

ricardosalveti:
Added the ack-by tag.

Optee_test related change at https://github.com/OP-TEE/optee_test/pull/512

Makefile
ricardosalveti: It might not really be needed with cmake it seems, mostly because CFG_TEE_PLUGIN_LOAD_PATH is already set to a default/known value in optee_test/host/supp_plugin/CMakeLists.txt.

With Makefile it is getting installed so it can later be consume by optee_test:
```
optee_client/config.mk:CFG_TEE_PLUGIN_LOAD_PATH ?= /usr/lib/tee-supplicant/plugins/
optee_test/Makefile:    $(echo) '  INSTALL ${DESTDIR}/$(CFG_TEE_PLUGIN_LOAD_PATH)'
```

Now with cmake that does not seem to be required:
```
optee_client/tee-supplicant/CMakeLists.txt:set (CFG_TEE_PLUGIN_LOAD_PATH "/usr/lib/tee-supplicant/plugins/" CACHE STRING "tee-supplicant's plugins path")
optee_client/tee-supplicant/CMakeLists.txt:set (CMAKE_INSTALL_RPATH "${CFG_TEE_PLUGIN_LOAD_PATH}")
optee_test/host/supp_plugin/CMakeLists.txt:set (CFG_TEE_PLUGIN_LOAD_PATH "/usr/lib/tee-supplicant/plugins/")
optee_test/host/supp_plugin/CMakeLists.txt:install (TARGETS ${PROJECT_NAME} DESTINATION ${CFG_TEE_PLUGIN_LOAD_PATH})
```

It is a bit weird that we use defaults with cmake and rely on config.mk for Makefile.

jforissier: > Now with cmake that does not seem to be required:

Correct.

> It is a bit weird that we use defaults with cmake and rely on config.mk for Makefile.

Yes, that's inconsistent, but why not.

ricardosalveti: Yeah, it's fine, just took me a few to understand :-)

etienne-lms:
There is no doc on `optee_client_config.mk` config scheme when building with pristine Makefile files.
Likely other Makefile specific ways, not described anywhere.
Whatever, this looks good to me, for 3.13.0 if needed.

jforissier:
@etienne-lms you are right, the doc is the code :wink: I will take this in the release (as well as the associated optee_test PR) since this can be considered a bug fix for OE builds.

PR #268, opened 2021-04-28
tee-supplicant: plugin: remove regular file type check

Emantor:
The readdir(3) man page mentions:
```
  Currently, only some filesystems (among them: Btrfs, ext2, ext3, and
  ext4) have full support for returning the file type in d_type. All
  applications must properly handle a return of DT_UNKNOWN.
```
Which is exactly the case on my nitrogen6x test system based on YOCTO
gatesgarth even using ext4:
```
  getdents64(4</usr/lib/tee-supplicant/plugins>,
  [{d_ino=3269, d_off=375390918, d_reclen=64, d_type=DT_UNKNOWN,
    d_name="f07bfc66-958c-4a15-99c0-260e4e7375dd.plugin"}, …], 32768) = 112
```
Even the regular test plugin for the tee-supplicant is identified as
DT_UNKNOWN, preventing the supplicant from loading the plugin.
Instead concatenate both the path and the filename and call stat(2) to
check if the resulting complete path is a regular file.

Signed-off-by: Rouven Czerwinski <r.czerwinski@pengutronix.de>

### DIFF
--- a/tee-supplicant/src/plugin.c
+++ b/tee-supplicant/src/plugin.c
@@ -122,48 +122,49 @@ TEEC_Result plugin_load_all(void)
 	}
 
 	while ((entry = readdir(dir))) {
-		if (entry->d_type == DT_REG) {
-			struct plugin *p;
+		struct plugin *p;
 
-			p = calloc(1, sizeof(struct plugin));
-			if (!p) {
-				EMSG("allocate mem for plugin <%s> failed",
-				     entry->d_name);
-				closedir(dir);
-				return TEEC_ERROR_OUT_OF_MEMORY;
-			}
+		if (!strcmp(entry->d_name, "..") || !strcmp(entry->d_name, "."))
+			continue;
 
-			res = load_plugin((const char *)entry->d_name, p);
-			switch (res) {
-			case PLUGIN_DL_OPEN_ERR:
-				EMSG("open plugin <%s> failed: %s",
-				     entry->d_name, dlerror());
-				free(p);
-				continue;
-			case PLUGIN_DL_SYM_ERR:
-				EMSG("find 'plugin_method' sym in <%s> failed: %s",
-				     entry->d_name, dlerror());
-				free(p);
-				continue;
-			default:
-				DMSG("loading the <%s> plugin were successful",
-				     p->method->name);
-				break;
-			}
-
-			/* Init the plugin */
-			if (p->method->init) {
-				teec_res = p->method->init();
-				if (teec_res) {
-					EMSG("init the <%s> plugin failed with 0x%x",
-					     p->method->name, teec_res);
-					free(p);
-					continue;
-				}
-			}
-
-			push_plugin(p);
+		p = calloc(1, sizeof(struct plugin));
+		if (!p) {
+			EMSG("allocate mem for plugin <%s> failed",
+			     entry->d_name);
+			closedir(dir);
+			return TEEC_ERROR_OUT_OF_MEMORY;
 		}
+
+		res = load_plugin((const char *)entry->d_name, p);
+		switch (res) {
+		case PLUGIN_DL_OPEN_ERR:
+			EMSG("open plugin <%s> failed: %s",
+			     entry->d_name, dlerror());
+			free(p);
+			continue;
+		case PLUGIN_DL_SYM_ERR:
+			EMSG("find 'plugin_method' sym in <%s> failed: %s",
+			     entry->d_name, dlerror());
+			free(p);
+			continue;
+		default:
+			DMSG("loading the <%s> plugin were successful",
+			     p->method->name);
+			break;
+		}
+
+		/* Init the plugin */
+		if (p->method->init) {
+			teec_res = p->method->init();
+			if (teec_res) {
+				EMSG("init the <%s> plugin failed with 0x%x",
+				     p->method->name, teec_res);
+				free(p);
+				continue;
+			}
+		}
+
+		push_plugin(p);
 	}
 
 	closedir(dir);


tee-supplicant/src/plugin.c
```diff
@@ -112,6 +113,8 @@ TEEC_Result plugin_load_all(void)
 	enum plugin_err res = PLUGIN_OK;
 	TEEC_Result teec_res = TEEC_SUCCESS;
 	struct dirent *entry = NULL;
+	char filename[1024];
```
jenswi-linaro, line 116: Please initialize these.
Use `PATH_MAX` instead of `1024`.

tee-supplicant/src/plugin.c
```diff
@@ -122,7 +125,10 @@ TEEC_Result plugin_load_all(void)
 	}
 
 	while ((entry = readdir(dir))) {
-		if (entry->d_type == DT_REG) {
+		snprintf(filename, 1024, "%s/%s", TEE_PLUGIN_LOAD_PATH,
+			 entry->d_name);
+		stat(filename, &sb);
```
jenswi-linaro, line 130: Check return value.

tee-supplicant/src/plugin.c
```diff
@@ -122,7 +125,10 @@ TEEC_Result plugin_load_all(void)
 	}
 
 	while ((entry = readdir(dir))) {
-		if (entry->d_type == DT_REG) {
+		snprintf(filename, 1024, "%s/%s", TEE_PLUGIN_LOAD_PATH,
```
jenswi-linaro, line 128: `sizeof(filename)`, instead of `1024`.
By the way, I wonder how `load_plugin()` below manages to open `entry->d_name` without the directory name.

tee-supplicant/src/plugin.c
jforissier, line 116: `1024` is arbitrary and might cause issues some day. I think we can easily get rid of it, see below.

tee-supplicant/src/plugin.c
jforissier, line 130: How about:
```C
if (fstatat(dirfd(dir), entry->d_name, 0))
	continue;
```

jforissier (commented):
Hi @Emantor,
Thanks for the patch. I suggest a change below but more fundamentally I don't think this "is a file" check is valid. I think we should call `load_plugin()` unconditionally and simply deal with errors there which we have to do anyways. This is a TOCTOU pattern.

tee-supplicant/src/plugin.c
Emantor: `TEE_PLUGIN_LOAD_PATH` gets added to the tee-supplicants `rpath` which allows `dlopen` to resolve the plugins with just the filename.

Emantor:
> Hi @Emantor,
> Thanks for the patch. I suggest a change below but more fundamentally I don't think this "is a file" check is valid. I think we should call `load_plugin()` unconditionally and simply deal with errors there which we have to do anyways. This is a TOCTOU pattern.

Sounds good, let me test this…

tee-supplicant/src/plugin.c
jenswi-linaro: Aha, thanks.

Emantor:
@jforissier I have pushed a new version implementing your suggestions. However, since I am also removing the `if` clause, the diff is far less readable. Thanks for the review @jenswi-linaro, but I found @jforissier's solution more usable.

tee-supplicant/src/plugin.c
```diff
@@ -122,48 +124,52 @@ TEEC_Result plugin_load_all(void)
...
-			switch (res) {
-			case PLUGIN_DL_OPEN_ERR:
-				EMSG("open plugin <%s> failed: %s",
-				     entry->d_name, dlerror());
-				free(p);
-				continue;
-			case PLUGIN_DL_SYM_ERR:
-				EMSG("find 'plugin_method' sym in <%s> failed: %s",
-				     entry->d_name, dlerror());
+		if (S_ISDIR(sb.st_mode))
```
jenswi-linaro, line 130: Is this test just to exclude `.` and `..`? If so we could check the names directly instead. Thinking of the TOCTOU mentioned by @jforissier before. Other than that I believe that all files found should be plugins so it's OK to complain if we can't load them.

jforissier: Agreed, then we could later extend the matching to something more subtle (`*.plugin` which would allow renaming a plugin to `foo.plugin.disabled` to avoid loading it for instance)

tee-supplicant/src/plugin.c
```diff
@@ -122,48 +124,52 @@ TEEC_Result plugin_load_all(void)
...
-			struct plugin *p;
-
-			p = calloc(1, sizeof(struct plugin));
-			if (!p) {
-				EMSG("allocate mem for plugin <%s> failed",
-				     entry->d_name);
-				closedir(dir);
-				return TEEC_ERROR_OUT_OF_MEMORY;
-			}
+		if (fstatat(dirfd(dir), entry->d_name, &sb, 0))
```
jforissier, line 127: I still don't think this is needed.

tee-supplicant/src/plugin.c
```diff
@@ -122,48 +124,52 @@ TEEC_Result plugin_load_all(void)
...
+			closedir(dir);
+			return TEEC_ERROR_OUT_OF_MEMORY;
+		}
+
+		res = load_plugin((const char *)entry->d_name, p);
+		switch (res) {
+		case PLUGIN_DL_OPEN_ERR:
+			EMSG("open plugin <%s> failed: %s",
+			     entry->d_name, dlerror());
+			free(p);
```
jforissier, line 148: Slightly unrelated -- but keeping `p` allocated for the next round would be smarter. Use `if (!p) p = calloc(...);` and set `p = NULL` on success after `push_plugin(p);` below. Perhaps for another patch ;-)

jforissier:
To be clear here is exactly what I have in mind (compile-tested only...): https://github.com/jforissier/optee_client/commit/5716ff411b079078915a47bba8abd98c7883bda4

tee-supplicant/src/plugin.c
Emantor: Pushed a new version which removes the stat buffer and uses strcmp.

jforissier (approved):

tee-supplicant/src/plugin.c
```diff
@@ -122,48 +122,49 @@ TEEC_Result plugin_load_all(void)
...
-			switch (res) {
-			case PLUGIN_DL_OPEN_ERR:
-				EMSG("open plugin <%s> failed: %s",
-				     entry->d_name, dlerror());
-				free(p);
-				continue;
-			case PLUGIN_DL_SYM_ERR:
-				EMSG("find 'plugin_method' sym in <%s> failed: %s",
-				     entry->d_name, dlerror());
+		struct plugin *p;
```
jenswi-linaro, line 128: Please move this to the beginning of the block.

Emantor: Sure, pushed an update.

jenswi-linaro:

Emantor:
Both tags are now applied.

jforissier:
Thanks @Emantor, I am applying this now since it is a bug fix worthwhile for the release.

PR #269, opened 2021-05-09
Add plugin support for AOSP build

vchong:


Android.mk
```diff
@@ -3,13 +3,22 @@
 ################################################################################
 LOCAL_PATH := $(call my-dir)
 
+# 3 (debug) is too noisy
+CFG_TEE_CLIENT_LOG_LEVEL ?= 2
+CFG_TEE_SUPP_LOG_LEVEL ?= 2
```
jforissier, line 8: Why should Android pick different defaults than other platforms? These default to 1 in `config.mk`.

jenswi-linaro:

Android.mk
```diff
@@ -14,6 +14,11 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib
 TEEC_TEST_LOAD_PATH ?= /data/vendor/tee
 CFG_TEE_FS_PARENT_PATH ?= /data/vendor/tee
 
+# set CFG_TEE_PLUGIN_LOAD_PATH before include config.mk
```
jforissier, line 17: How many time this pattern needs to be repeated? ;-) Please add `Define Android-specific configuration before including config.mk` after the `LOCAL_PATH` line  then drop the existing `# set CFG_XXX before..." lines.

```make
LOCAL_PATH := $(call my-dir)

# Define Android-specific configuration before including config.mk

CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib

TEEC_TEST_LOAD_PATH ?= /data/vendor/tee
CFG_TEE_FS_PARENT_PATH ?= /data/vendor/tee
```

Android.mk
```diff
@@ -14,6 +14,11 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib
...
+# there is no /usr/lib nor /vendor/usr/lib in AOSP
```
jforissier, line 18: remove this line

Android.mk
```diff
@@ -14,6 +14,11 @@ CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib
...
+# use /vendor/lib/tee-supplicant/plugins/ for 32b builds
```
jforissier, line 19: this should really be handled automatically by the build system don't you think?

jforissier (commented):
A few minor things

Android.mk
jforissier: Personally I never change these, and in fact I don't know how to change them (QEMU Buildroot environment, `make CFG_TEE_CLIENT_LOG_LEVEL=3 CFG_TEE_SUPP_LOG_LEVEL=3 run` doesn't appear to work).

If that's more convenient for you I'm fine with this change.

jforissier:

Android.mk
vchong: > Personally I never change these, and in fact I don't know how to change them (QEMU Buildroot environment, `make CFG_TEE_CLIENT_LOG_LEVEL=3 CFG_TEE_SUPP_LOG_LEVEL=3 run` doesn't appear to work).

I don't remember changing it myself either in `build.git` but since we've removed the `optee-client` and `xtest` targets from the Makefile, my guess is we'll have to change them here:
https://github.com/OP-TEE/optee_client/blob/master/libteec/CMakeLists.txt#L18
https://github.com/OP-TEE/optee_client/blob/master/tee-supplicant/CMakeLists.txt#L11

> If that's more convenient for you I'm fine with this change.

Great. Thanks!

vchong:
Tags added.

PR #271, opened 2021-06-02
libckteec: fix spelling mistake

zhupg:
Fix spelling mistake in pkcs11_api.c

Signed-off-by: Pengguang Zhu <zpghao@163.com>


jforissier:
> `Prior this change building libckteec in debug mode failed with an error meesage like:`

Describing the situation in the present tense (before the patch is applied) is in general simpler and more direct: "Building in debug mode fails with the following error:".

zhupg:
added reviewed-by tag.

PR #273, opened 2021-06-27
libckteec: Fix signature size query for C_Sign/C_SignFinal

vesajaaskelainen:
As specified in:
PKCS #11 Cryptographic Token Interface Base Specification
Version 2.40 Plus Errata 01
5.2 Conventions for functions returning output in a variable-length buffer

When querying size of returned variable-length buffer actual value of
*pulBufLen for input can be anything as long as pBuf is NULL and pulBufLen
is not NULL.

Related PRs:
- https://github.com/OP-TEE/optee_test/pull/524


ruchi393:
`Reviewed-by:  Ruchika Gupta <ruchika.gupta@linaro.org>`

vesajaaskelainen:
Thanks for the review. Applied tags.

PR #275, opened 2021-06-30
libckteec: Fix wrapped key output buffer size query for C_WrapKey

vesajaaskelainen:
As specified in:
PKCS #11 Cryptographic Token Interface Base Specification
Version 2.40 Plus Errata 01
5.2 Conventions for functions returning output in a variable-length buffer

When querying size of returned variable-length buffer actual value of
*pulBufLen for input can be anything as long as pBuf is NULL and pulBufLen
is not NULL.

Fixes problem found during reviewing PR:
https://github.com/OP-TEE/optee_os/pull/4469


etienne-lms (commented):

vesajaaskelainen:
Thanks for the review. Applied tags.

PR #280, opened 2021-07-29
libckteec: Miscellaneus fixes

ruchi393:
Few bug fixes in pkcs11 library

### DIFF
--- a/libckteec/src/pkcs11_processing.c
+++ b/libckteec/src/pkcs11_processing.c
@@ -1621,8 +1621,11 @@ CK_RV ck_unwrap_key(CK_SESSION_HANDLE session, CK_MECHANISM_PTR mechanism,
 	size_t out_size = 0;
 
 	if (!handle || !mechanism || (count && !attribs) ||
-	    !unwrapping_key || (wrapped_key_len && !wrapped_key))
+	    (wrapped_key_len && !wrapped_key))
 		return CKR_ARGUMENTS_BAD;
+
+	if (!unwrapping_key)
+		return CKR_UNWRAPPING_KEY_HANDLE_INVALID;
 
 	rv = serialize_ck_mecha_params(&smecha, mechanism);
 	if (rv)
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -298,7 +298,12 @@ static CK_RV deserialize_ck_attribute(struct pkcs11_attribute_head *in,
 			return CKR_ATTRIBUTE_TYPE_INVALID;
 
 		memcpy(&pkcs11_data32, data, sizeof(uint32_t));
-		ck_ulong = pkcs11_data32;
+		if (out->type == CKA_KEY_GEN_MECHANISM &&
+		    pkcs11_data32 == PKCS11_CK_UNAVAILABLE_INFORMATION)
+			ck_ulong = CK_UNAVAILABLE_INFORMATION;
+		else
+			ck_ulong = pkcs11_data32;
+
 		memcpy(out->pValue, &ck_ulong, sizeof(CK_ULONG));
 		out->ulValueLen = sizeof(CK_ULONG);
 		return CKR_OK;


libckteec/src/serialize_ck.c
```diff
@@ -298,7 +298,12 @@ static CK_RV deserialize_ck_attribute(struct pkcs11_attribute_head *in,
 			return CKR_ATTRIBUTE_TYPE_INVALID;
 
 		memcpy(&pkcs11_data32, data, sizeof(uint32_t));
-		ck_ulong = pkcs11_data32;
+		if (out->type == CKA_KEY_GEN_MECHANISM &&
```
vesajaaskelainen, line 301: I had this commit in my stack some time ago -- but as I forgot what if fixed I dropped it:
```patch
commit 74edd9c31a01e10b156e69d2c5e46d7ac5c2a8cc
Author: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>
Date:   Sat Jan 9 12:29:14 2021 +0200

    libckteec: Add 64 bit sign extension support for deserialization
    
    Adds support to sign extend 32bit values to 64 bit values as otherwise
    matching with the defines with values ~0UL would not work.
    
    Note: this expects that ALL values will fit in 32 bits as TA interface has
    32 bits defined for transport for CK_ULONG and ~0UL for
    CK_UNAVAILABLE_INFORMATION is only special case that needs to be handled.
    
diff --git a/libckteec/src/serialize_ck.c b/libckteec/src/serialize_ck.c
index 2984608..481582d 100644
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -307,6 +307,8 @@ static CK_RV deserialize_ck_attribute(struct pkcs11_attribute_head *in,
 
                memcpy(&pkcs11_data32, data, sizeof(uint32_t));
                ck_ulong = pkcs11_data32;
+               if (ck_ulong == (uint32_t)~0UL)
+                       ck_ulong = (CK_ULONG)~0UL;
                memcpy(out->pValue, &ck_ulong, sizeof(CK_ULONG));
                out->ulValueLen = sizeof(CK_ULONG);
                return CKR_OK;
```

What is good with that it fixes all cases -- but only problem is that if someone wants to store value ~0UL and not extended one then that could be a problem -- but could not spot a case for that either.

ruchi393: I would prefer explicit check just for readability sake. As you said it gets difficult to remember over time if the change is not readable.

vesajaaskelainen: Wondering if we have more cases like this -- we have often used switch() structure for handling special cases so should do that here as well?

libckteec/src/pkcs11_processing.c
```diff
@@ -1621,9 +1621,12 @@ CK_RV ck_unwrap_key(CK_SESSION_HANDLE session, CK_MECHANISM_PTR mechanism,
 	size_t out_size = 0;
 
 	if (!handle || !mechanism || (count && !attribs) ||
-	    !unwrapping_key || (wrapped_key_len && !wrapped_key))
+	    (wrapped_key_len && !wrapped_key))
 		return CKR_ARGUMENTS_BAD;
 
+	if (!unwrapping_key)
```
vesajaaskelainen, line 1627: Handling errors in here is a bit dangerous as there are priorities how they should be handled -- so I am wondering is it just better to let it pass to TA side and let it handle the error cases.

Above change is OK to relax it from `CKR_ARGUMENTS_BAD` handling.

ruchi393: This error gets handled by TA also if the NULL unwrapping key gets passed to TA. I just thought it would be good to handle it here also additionally rather than passing it all the way to TA. 

Can you explain what you mean by priorities ?

vesajaaskelainen: If we look in [5.1 Function return values](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755792) we can see that there are different groups of return values.

And then in [5.1.7 More on relative priorities of Cryptoki errors](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755799):

> In general, when a Cryptoki call is made, error codes from Section 5.1.1 (other than CKR_OK) take precedence over error codes from Section 5.1.2, which take precedence over error codes from Section 5.1.3, which take precedence over error codes from Section 5.1.6.  One minor implication of this is that functions that use a session handle (i.e., most functions!) never return the error code CKR_TOKEN_NOT_PRESENT (they return CKR_SESSION_HANDLE_INVALID instead).  Other than these precedences, if more than one error code applies to the result of a Cryptoki call, any of the applicable error codes may be returned.  Exceptions to this rule will be explicitly mentioned in the descriptions of functions.

Meaning that we should use that priority order specification when determining order of checks being done and what errors is being reported.

If we look at that list we have kind a problem with `CKR_SESSION_HANDLE_INVALID` because we cannot determine that before `CKR_ARGUMENTS_BAD` as we need to do memory management operations based on arguments before making the TA call. But if we ignore that particular priority problem then we can do rest in proper order in the TA side I think.

libckteec/src/serialize_ck.c
etienne-lms: From (my reading of) the spec, CK_UNAVAILABLE_INFORMATION is used only at 2 places:
- in `ulValueLen` field to denote specific info about the attribute (already handled in libckteec)
- in attribute value for attribute CKA_KEY_GEN_MECHANISM (this change).
- in User Interface Objects related attribute (see section 4.3.5.2 of the spec v2.40-e01)

I agree the change proposed here is better than a blind generic one.
We will have to handle User Interface Object attribs when we'll support them.

etienne-lms (commented):

libckteec/src/pkcs11_processing.c
etienne-lms: I think the currently proposed implementation is fine.
There are many other places where we can return `CKR_ARGUMENTS_BAD` before even testing a possibly invalid session handle (`ck_login()`, `ck_generate_random()`, `ck_create_object()` etc...). If to be addressed, let's deal with that in a specific P-R.

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commit "libckteec: Correct error returned for C_UnwrapKey()".

libckteec/src/pkcs11_processing.c
ruchi393: Thanks @vesajaaskelainen  for the reference to specs. I had completely forgotten about that. I will leave it as it is for now and as @etienne-lms suggested and we can deal with it in a separate PR.

libckteec/src/serialize_ck.c
ruchi393: Yes I also couldn't find any other reference apart from the ones stated by @etienne-lms . Perhaps I will leave this as it is and we can later change to switch if we find more usecases where ck_ulong is being treated differently on trusted and non-secure world side.

libckteec/src/pkcs11_processing.c
vesajaaskelainen: OK

libckteec/src/serialize_ck.c
vesajaaskelainen: OK

vesajaaskelainen:
`Reviewed-by: Vesa Jääskeläinen <vesa.jaaskelainen@vaisala.com>`

ruchi393:
All tags applied.

PR #282, opened 2021-08-06
tee-supplicant: add config CFG_TEE_SUPP_PLUGIN to enable or not plugin

etienne-lms:
This change allows optee_client configuration to enable or not supplicant
plugin support. The feature is enabled from switch CFG_TEE_SUPP_PLUGIN.
The configuration is default enable for backward compatibility purpose.

There are OSes where TEE supplicant plugin cannot be supported for
example when the OS does not enable dynamic/shared libraries. Such
systems should still be able to embed OP-TEE assuming the supplicant
plugin support is not embedded. Prior this change was optee_client
failing to build for such systems with a error message like the one
below reported by Buildroot CI build [1]:

/tmp/instance-2/output-1/build/optee-client-3.13.0/tee-supplicant/src/plugin.c:8:10: fatal error: dlfcn.h: No such file or directory
    8 | #include <dlfcn.h>
      |          ^~~~~~~~~

Link: [1] http://autobuild.buildroot.net/results/56d1ec172d0d857d6a968397b7785f961fad1043/


Android.mk
```diff
@@ -11,6 +11,7 @@ CFG_TEE_SUPP_LOG_LEVEL ?= 2
 CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib
 TEEC_TEST_LOAD_PATH ?= /data/vendor/tee
 CFG_TEE_FS_PARENT_PATH ?= /data/vendor/tee
+CFG_TEE_SUPP_PLUGIN ?= y
```
jforissier, line 14: `CFG_TEE_SUPP_PLUGINS` maybe? Without an S it sounds like we're talking about one specific plugin, so either `PLUGINS` to mean all plugins in general or `CFG_TEE_SUPP_PLUGIN_SUPPORT` (which I find a bit too long).

jforissier (commented):
One minor comment, other than that LGTM.

etienne-lms:
updated 
review tag not applied yet, I'm waiting for some more tests before doing so.

etienne-lms:
tag applied

jenswi-linaro:

PR #283, opened 2021-08-14
libckteec: Add certificate object support

vesajaaskelainen:
Synchronize pkcs11 TA interface for certificate support.

Add necessary `pkcs11.h` definitions so that `xtest` can be compiled for pkcs11 certificate tests.

Relates to:
- https://github.com/OP-TEE/optee_os/issues/4283

Related PRs:
- https://github.com/OP-TEE/optee_os/pull/4797
- https://github.com/OP-TEE/optee_test/pull/542

### DIFF
--- a/libckteec/include/pkcs11.h
+++ b/libckteec/include/pkcs11.h
@@ -242,6 +242,27 @@ typedef CK_KEY_TYPE *CK_KEY_TYPE_PTR;
 #define CKK_SHA224_HMAC			0x02e
 
 /*
+ * Certificates
+ */
+typedef CK_ULONG CK_CERTIFICATE_TYPE;
+typedef CK_ULONG CK_CERTIFICATE_CATEGORY;
+
+/*
+ * Valid values for attribute CKA_CERTIFICATE_TYPE
+ */
+#define CKC_X_509			0x00000000UL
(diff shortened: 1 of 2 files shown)


etienne-lms:

libckteec/include/pkcs11.h
```diff
@@ -241,6 +241,27 @@ typedef CK_KEY_TYPE *CK_KEY_TYPE_PTR;
...
+/*
+ * Certificates
+ */
+typedef CK_ULONG CK_CERTIFICATE_TYPE;
+typedef CK_ULONG CK_CERTIFICATE_CATEGORY;
+
+/*
+ * Valid values for attribute CKA_CERTIFICATE_TYPE
+ */
+#define CKC_X_509			0x00000000
```
etienne-lms, line 253: nitpick: `0x0`, `0x1` and `0x2` are enough. Discard my comment if you prefer the 32bit expansion.
my R-b applies even if not addressed.

vesajaaskelainen: This header should be close to what official [pkcs11 headers](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/include/pkcs11-v2.40/pkcs11t.h) are defined to be. 

That being said -- official ones has UL at the end so I will do that also here.

PR #288, opened 2021-10-13
Assorted fixes

parport0:
We have discovered a couple of places where the return values of functions weren't checked. Add the missing checks, and add `tags` and `TAGS` to .gitignore while at it.

These issues were found with SVACE static analysis tool.

### DIFF
--- a/libckteec/src/pkcs11_processing.c
+++ b/libckteec/src/pkcs11_processing.c
@@ -460,6 +460,10 @@ CK_RV ck_digest_update(CK_SESSION_HANDLE session, CK_BYTE_PTR in,
 	 * Shm io0: (in/out) ctrl = [session-handle] / [status]
 	 */
 	ctrl = ckteec_alloc_shm(sizeof(session_handle), CKTEEC_SHM_INOUT);
+	if (!ctrl) {
+		rv = CKR_HOST_MEMORY;
+		goto bail;
+	}
 	buf = ctrl->buffer;
 
 	memcpy(buf, &session_handle, sizeof(session_handle));


tee-supplicant/src/teec_ta_load.c
```diff
@@ -138,7 +138,11 @@ static int try_load_secure_module(const char* prefix,
 	}
 
 	s = ftell(file);
-	if (s > *ta_size || !ta) {
+	if (s == (size_t) -1) {
```
jforissier, line 141: `if (s < 0)` would be better

jforissier:
For commit: ".gitignore: add tags and TAGS":

tee-supplicant/src/teec_ta_load.c
parport0: s is size_t. Also, ftell always returns -1 on error.

jenswi-linaro: Hmm, wouldn't it be better to use the correct type when taking the return value from `ftell()`?

tee-supplicant/src/teec_ta_load.c
```diff
@@ -138,7 +138,11 @@ static int try_load_secure_module(const char* prefix,
...
+		printf("failed to ftell the TA file\n");
```
jenswi-linaro, line 142: We're printing the name of the file above on error, please do so here too.

tee-supplicant/src/teec_ta_load.c
parport0: It's probably more of a matter of taste; alright, I'll add a new `long` variable.

tee-supplicant/src/teec_ta_load.c
```diff
@@ -137,7 +138,14 @@ static int try_load_secure_module(const char* prefix,
...
 
-	s = ftell(file);
+	l = ftell(file);
+	if (l == -1) {
+		DMSG("failed to ftell the ta %s TA-file", fname);
+		fclose(file);
+		return TA_BINARY_NOT_FOUND;
+	}
+
+	s = (size_t) l;
```
jenswi-linaro, line 148: Cast not needed?

tee-supplicant/src/teec_ta_load.c
```diff
@@ -137,7 +138,14 @@ static int try_load_secure_module(const char* prefix,
 		return TA_BINARY_NOT_FOUND;
 	}
 
-	s = ftell(file);
+	l = ftell(file);
+	if (l == -1) {
```
jenswi-linaro, line 142: We should perhaps rule out all negative values here since we're going to put this in an unsigned variable further down.

tee-supplicant/src/teec_ta_load.c
```diff
@@ -157,7 +161,7 @@ static int try_load_secure_module(const char* prefix,
 	}
 
 	if (s != fread(ta, 1, s, file)) {
-		printf("error fread TA file\n");
+		DMSG("failed to fread the ta %s TA-file", fname);
```
jforissier, line 164: Changing `printf()` to `DMSG()` looks like a good idea but it should be a separate commit IMO (note there is another instance in `try_load_secure_module()`)

jforissier (commented):
With my comment below addressed:

PR #290, opened 2021-11-05
libseteec: Secure Element control

ldts:
libseteec: Secure Element control
    
    A secure element might be configured by OP-TEE; if the SE enabled
    SCP03, the encryption keys should have been derived from the HUK and
    never known to the normal world. In such circumstances, APDU frames
    will need to be routed by the secure world as it is the only one with
    access to the SCP03 keys.
    
    By default, Secure Elements might be shipped with predefined keys
    stored in persistent memory and documented in the device data
    sheets.
    
    This library provides an interface to enable SCP03 using those
    non-secure keys. It also provides an interface to rotate these default
    keys and derive board unique new ones.
    
    A working scenario currently in OP-TEE is the NXP SE050 device.

Reference: https://github.com/OP-TEE/optee_os/pull/4935

In this diagram, please notice that APDU frames sent to the SE050 will be returned with an APDU response to be processed by the client.

simwtop-Page-2


libapduteec/src/apdu.c
```diff
@@ -0,0 +1,201 @@
...
+static void free_shm(TEEC_SharedMemory *shm)
+{
+	if (!shm)
+		return;
+
+	TEEC_ReleaseSharedMemory(shm);
+	free(shm);
+}
+
+__attribute__((constructor)) static int apdu_init(void)
```
jforissier, line 61: Why not provide a public API function for this? Then the caller could deal with errors.

libapduteec/src/apdu.c
```diff
@@ -0,0 +1,201 @@
...
+		goto out;
+	}
+
+	ta_ctx.initiated = true;
+out:
+	e = pthread_mutex_unlock(&ta_ctx.init_mutex);
+	if (e) {
+		EMSG("pthread_mutex_unlock: %s", strerror(e));
+		EMSG("terminating...");
+		exit(EXIT_FAILURE);
```
jforissier, line 109: Killing the whole process just because one library could not initialize is quite drastic!

libapduteec/src/apdu.c
```diff
@@ -0,0 +1,201 @@
...
+{
+	uint32_t login_method = TEEC_LOGIN_PUBLIC;
+	TEEC_UUID uuid = PTA_APDU_UUID;
+	TEEC_Result res = TEEC_SUCCESS;
+	void *login_data = NULL;
+	uint32_t origin = 0;
+	int ret = 0;
+	int e = 0;
+
+	e = pthread_mutex_lock(&ta_ctx.init_mutex);
```
jforissier, line 71: What is this protecting against?

jforissier (commented):
I think you should drop the pthread stuff (let the caller deal with thread safety if needed) and add initialization and finalization functions. That would make things much simpler IMO. Or do you have specific scenarios in mind for this library?

libapduteec/src/apdu.c
ldts: yes I agree. But I'd rather let the client handle it before main executes - ie, as early as possible. Will use a signal for that.

libapduteec/src/apdu.c
ldts: right, pkcs#11 is doing the same thing...I am not sure is a good idea either. In any case, since the constructor is establishing the session I have no need for this lock so will remove.

libapduteec/src/apdu.c
ldts: ie: https://github.com/ldts/apdu/blob/master/main.c#L143

ldts:
um, IBART is failing with: @jbech-linaro 

Resolving deltas: 100% (283/283), completed with 91 local objects.
    17:  
    18:  Your identity is: Joakim Bech <joakim.bech@linaro.org>
    19:  If you want to change this, please re-run 'repo init' with --config-name
    20:  
    21:  repo has been initialized in /home/optee/devel/hikey
    22:  0
    23:  IBART $ repo sync -j4;echo $?
    24:  
    25:  ... A new version of repo (2.17) is available.
    26:  ... You should upgrade soon:
    27:      cp /home/optee/devel/hikey/.repo/repo/repo /home/optee/bin/repo

jforissier:
@ldts it is just a warning, not supposed to be fatal, I suspect a timeout. I am restarting the job and we'll see.

libapduteec/include/apdu_ta.h
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright (C) 2021, Foundries Limited
+ */
+
+#ifndef APDU_TA_H
+#define APDU_TA_H
+
+#define PTA_APDU_UUID { 0x3f3eb880, 0x3639, 0x11ec, \
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+/*
+ * ADPU based communication with the SE05X secure element
```
etienne-lms, line 12: Again, is this library targeting only SE05x devices?

ldts: no it is supposed to target any device...I guess I imported an old file. 
I'll reword.

etienne-lms (commented):
> during initialization, the library will raise the SIGABRT signal
therefore giving the client the chance to start its main program in a
different mode.

A better way IMO is to have initialization and termination API functions for this library. It's more flexible for a client to call some `C_APDU_Initialize()` and deal with result rather that trapping a signal.

libapduteec/include/apdu_ta.h
```diff
@@ -0,0 +1,30 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
```
jenswi-linaro, line 1: This library is about abstracting this so perhaps this file should go into `libapduteec/src`.

libapduteec/src/apdu.c
```diff
@@ -0,0 +1,155 @@
...
+static void free_shm(TEEC_SharedMemory *shm)
+{
+	if (!shm)
+		return;
+
+	TEEC_ReleaseSharedMemory(shm);
+	free(shm);
+}
+
+__attribute__((constructor(1000))) static void apdu_init(void)
```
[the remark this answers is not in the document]
jenswi-linaro: I agree, this sort of thing seems to cause more troubles than it solves.

libapduteec/include/apdu.h
```diff
@@ -0,0 +1,39 @@
...
+/*
+ * Copyright (c) 2021, Foundries.io
+ */
+
+#ifndef APDU_H
+#define APDU_H
+
+#include <unistd.h>
+
+/* The library throws SIGABRT before the main is executed - the main program can
```
jenswi-linaro, line 11: How is `main()` supposed to catch that?

libapduteec/src/apdu.c
```diff
@@ -0,0 +1,155 @@
...
+		EMSG("APDU invalid input parameters");
+		ret = EINVAL;
+		goto out;
+	}
+
+	memcpy(shm_hdr->buffer, hdr, hdr_len);
+	memcpy(shm_src->buffer, src, src_len);
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_VALUE_INPUT,
+					 TEEC_MEMREF_WHOLE,
```
jenswi-linaro, line 124: How about using `TEEC_MEMREF_*` instead and let `TEEC_InvokeCommand()` do the buffer handling?

libapduteec/src/apdu.c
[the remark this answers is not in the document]
ldts: ok, we can do that. @etienne-lms 
ah, you mean we dont need to issue the TEEC_CloseSession before the termination of the process (on destruction)?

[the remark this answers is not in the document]
etienne-lms: yes. libteec and tee/optee drivers release all OP-TEE resources related to the task when the task ends.

libapduteec/include/apdu.h
ldts: @jenswi-linaro just using a higher priority constructor...but it doesnt matter, will remove the constructor thingy.

libapduteec/src/apdu.c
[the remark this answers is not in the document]
ldts: ah, great. ok.

libapduteec/src/apdu.c
ldts: @jenswi-linaro I dont understand, could you give me some more info/reference please?

jenswi-linaro: For example:
https://github.com/OP-TEE/optee_test/blob/13c4de15fa0370bfd37de476f02c103fd47ccc47/host/xtest/regression_1000.c#L272-L275

Skimming through the TEE Client API specification should give some more background.

ldts: @jenswi-linaro ah you mean for param#0?

etienne-lms: Instead of allocating share dmemory, you can pass stright your buffer reference:

```c
	op.paramTypes = TEEC_PARAM_TYPES(TEEC_VALUE_INPUT,
					 TEEC_MEMREF_TEMP_INPUT,
					 TEEC_MEMREF_TEMP_INPUT,
					 TEEC_MEMREF_TEMP_OUTPUT);
	op.params[0].value.a = type;
	op.params[1].tmpref.buffer = hdr;
	op.params[1].tmpref.size = hdr_len;
	op.params[2].tmpref.buffer = src;
	op.params[2].tmpref.size = src_len;
	op.params[3].tmpref.buffer = dst;
	op.params[3].tmpref.size = dst_len;

Comment (jenswi-linaro) [reply libapduteec/src/apdu.c:124]: Thanks, spot on Etienne.

Comment (ldts) [reply libapduteec/src/apdu.c:124]: ah ok. so the same would apply to the the pkcs11 library then?

File: libapduteec/include/apdu.h [hunk 9]
```diff
@@ -0,0 +1,39 @@
...
+	APDU_TYPE_0,
+	APDU_TYPE_1,
+	APDU_TYPE_2,
+	APDU_TYPE_3,
+	APDU_TYPE_4,
+	APDU_TYPE_5,
+	APDU_TYPE_6,
+};
+
+int C_APDU_request(enum apdu_type type,
```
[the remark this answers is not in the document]
ldts: many thanks for this

libapduteec/include/apdu.h
```diff
@@ -0,0 +1,39 @@
...
+
+/* The library throws SIGABRT before the main is executed - the main program can
+ * therefore decide on how best to proceed with its boot sequence.
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum apdu_type {
```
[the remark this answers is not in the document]
ldts: @etienne-lms what do you mean? why is this not enough?

libapduteec/src/apdu.c
etienne-lms: Could be yes. However, looking back into the code, I didn't find places where such scheme would apply. Maybe I missed them. Feel free to propose a P-R if there are places where such sequence would simply libckteec implementation.

libapduteec/include/apdu.h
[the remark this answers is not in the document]
etienne-lms: please, provide a definition of what is an APDU type and which value is expected for this argument.

libapduteec/include/apdu.h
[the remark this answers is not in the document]
etienne-lms: btw, don't forget `@return 0     Success`

libapduteec/include/apdu.h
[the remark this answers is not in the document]
ldts: Would something like this work? 

I don't know what other drivers might need but the NXPSE050 uses three types to indicate if the response must be dropped and or if the raw frame is in the[ extended format.](https://docs.oracle.com/en/java/javacard/3.1/guide/extended-apdu-format.html#GUID-0F452782-9402-43E9-B39C-337772C200DF)

These types are driver specific and I expect -as in the case of the SE05X, that the driver itself will be outside of the OP-TEE repository in an external library.

```

/**
 * Device specific  requestflags to help the driver process the frames.
 *
 * Can be used to indicate if the APDU frame is of an extended type or if the
 * response frame to a particular request should be dropped.
 */
```

libapduteec/include/apdu.h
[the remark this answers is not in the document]
ldts: argh ok

libseteec/include/se.h
```diff
@@ -0,0 +1,86 @@
...
+ * @return -EINVAL	Invalid arguments regarding an APDU message
+ * @return -ENODEV	OP-TEE device not available
+ * @return -EPERM	Access permission failure
+ * @return -EAGAIN	Initialization mutex not available
+ * @return -EACCESS	Secure element processing error
+ *
+ * Returns 0 on success
+ */
+
+int C_SE_apdu_request(enum se_apdu_type type,
```
jenswi-linaro, line 53: Are these function names following any standard? If not how about s/C_SE_/se_/?

libseteec/src/se.c
```diff
@@ -0,0 +1,174 @@
...
+			       login_method, login_data, NULL, &origin);
+	if (res) {
+		EMSG("TEEC open session failed, -EPERM");
+		ret = -EPERM;
+		TEEC_FinalizeContext(&ctx->context);
+		goto out;
+	}
+
+	ctx->initiated = true;
+	return 0;
```
ldts, line 74: sorry, fixing this

libseteec/include/se.h
```diff
@@ -0,0 +1,86 @@
...
+#define SE_H
+
+#include <unistd.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Device specific request flags to help the driver process the frames.
```
etienne-lms, line 17: So application should know which device is used?

libseteec/include/se.h
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * Can be used to indicate if the APDU frame is of an extended type or if the
+ * response frame to a particular request should be dropped.
```
etienne-lms, line 20: Can you give a pointer explaining which type relates to which extension/response/...?

libseteec/src/scp03_ta.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef SCP03_TA_H
+#define SCP03_TA_H
+
+#define TA_SCP03_UUID { 0xbe0e5821, 0xe718, 0x4f77, \
+			{ 0xab, 0x3e, 0x8e, 0x6c, 0x73, 0xa9, 0xc7, 0x35 } }
+
+/*
+ * Enable SCP03 support on the SE
+ *
+ * [in]     value[0].a    extra option (0 do not rotate keys, 1 rotate keys)
```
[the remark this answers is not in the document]
ldts: @etienne-lms, please do not worry - your questions are always challenging - and welcome!- and more often than not uncover important issues

To clarify, the rotate request corresponds to the PUTKEY command.

Notice that it must be captured in GP specs because to enable SCP03,  SCP03 **must be running already** with some default keys.  So, it would be strange if the user is not allowed by design - and captured in the specification- to modify those default keys (this is key rotation).

Hence why this is a generic interface and not NXP SE050 specific.

libseteec/src/apdu_ta.h
```diff
@@ -0,0 +1,33 @@
...
+ * [out] memref[3].buffer     response data (raw apdu response frame data).
+ * [out] memref[3].size       response len (raw apdu respose length).
+ *
+ * Result:
+ * TEE_SUCCESS - Invoke command success
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_NOT_IMPLEMENTED - Invoke command not implemented
+ * TEE_ERROR_GENERIC - Invoke command failure
+ */
+#define TA_CMD_TXRX_APDU_RAW_FRAME		0
```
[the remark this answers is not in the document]
ldts: I'd rather keep the APDU_RAW_FRAME: the type is the type of raw frame

libseteec/src/se.c
```diff
@@ -0,0 +1,174 @@
...
+	uint32_t origin = 0;
+	int ret = 0;
+
+	ret = se_init(&uuid, &apdu_ta_ctx);
+	if (ret)
+		return ret;
+
+	if (!src || !dst || (hdr && !hdr_len)) {
+		EMSG("APDU invalid input parameters, -EINVAL");
+		return -EINVAL;
```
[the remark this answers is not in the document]
ldts: @etienne-lms why can we leave it open for another request? the user will be sending many APDUs over a period of time

libseteec/src/se.c
```diff
@@ -0,0 +1,174 @@
...
+		EMSG("TEEC invoke command failed, -EINVAL");
+		*dst_len = 0;
+		return -EINVAL;
+	}
+
+	*dst_len = op.params[3].tmpref.size;
+	if (*dst_len)
+		return 0;
+
+	EMSG("Processing in the Secure Element failed, -EIO");
```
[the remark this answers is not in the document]
ldts: EACCESS

libseteec/include/se.h
```diff
@@ -0,0 +1,86 @@
...
+ * @return -EINVAL	On error
+ *
+ * Returns 0 on success
+ */
+int C_SE_scp03_enable(void);
+
+/**
+ * C_SE_scp03_rotate_keys()  - Rotate the default SCP03 keys
+ *
+ * Generates the board specific SCP03 keys and writes them in the Secure Element
```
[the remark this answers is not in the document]
ldts: C_se_scp03_generate_keys() ?

libseteec/include/se.h
```diff
@@ -0,0 +1,86 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2021, Foundries.io
+ * Jorge Ramirez-Ortiz <jorge@foundries.io>
+ */
+
+#ifndef SE_H
+#define SE_H
```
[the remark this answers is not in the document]
ldts: se_tee.h

libseteec/include/se.h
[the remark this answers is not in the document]
jenswi-linaro: Why should we have a `C_` prefix?

libseteec/include/se.h
ldts: finally found the reference in the spec. fixing it.

libseteec/include/se.h
ldts: yes, will be in the header

libseteec/include/se.h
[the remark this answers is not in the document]
ldts: right, @etienne-lms I was wondering the same thing, could you clarify please?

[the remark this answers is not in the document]
etienne-lms: drop the `C_`, I agree we don't need it.  My concern was rather on `rotate_keys` that should rather be `generate_keys`.

libseteec/src/apdu_ta.h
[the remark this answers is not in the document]
etienne-lms: For consistency among PTA APIs, prefer command ID macros as `PTA_<pta-short-name>_XXX` as almost all PTA command IDs are defined.
Here: `PTA_APDU_CMD_TXRX_RAW_FRAME` or `PTA_APDU_TXRX_RAW_FRAME`.

Side note: raw frame would rather expect raw data, hence no TYPE hint value. If a type hint is passed by the command then `PTA_APDU_CMD_TXRX_TYPED_FRAME` would be more explicit IMO. Yet, not a strong opinion.

libseteec/include/se.h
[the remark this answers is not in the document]
ldts: but it wont be just generating keys, it will be rotating the keys - ie, replacing the running keys in the device - because they must be in execution while this is happening- with new keys. there is a distinction...

ldts:
@etienne-lms @jenswi-linaro I think all comments have now been addressed.

jenswi-linaro:

libseteec/src/apdu_ta.h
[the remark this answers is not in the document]
ldts: @etienne-lms um...ok, I thought you had asked before to remove the PTA_ prefix for TA_...so I did. 
But ok, will put it back to how it was - and how it is in OP-TEE....maybe I misunderstood.

The frames can be RAW or SCP03 ENCRYPTED.; this is the reason for the PTA...RAW frames that dont need to be SCP03 enrypted could be placed on the bus any time without requirement OP-TEE; it is only when SCP03 has been enabled - and the keys are in the secure world - that the interface makes sense. So the interface will only work for RAW frames hence why I wanted to be explicit and clear about it. The type is the actual type of APDU frame...not whether the frame is encrypted or not.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+ * @param src		Pointer to APDU message payload
+ * @param src_len	Byte length of message payload @src
+ * @param dst		Pointer to APDU message reponse buffer
+ * @param dst_len	Byte length of reponse buffer @dst
+ *
+ * @return -EINVAL	Invalid arguments regarding an APDU message
+ * @return -ENODEV	OP-TEE device not available
+ * @return -EPERM	Access permission failure
+ * @return -EAGAIN	Initialization mutex not available
+ * @return -EACCESS	Secure element processing error
```
etienne-lms, line 43: `EACCES Permission denied (POSIX.1-2001).`
You should use another errno for that error.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+ * @param hdr_len	Byte length of message header @hdr
+ * @param src		Pointer to APDU message payload
+ * @param src_len	Byte length of message payload @src
+ * @param dst		Pointer to APDU message reponse buffer
+ * @param dst_len	Byte length of reponse buffer @dst
+ *
+ * @return -EINVAL	Invalid arguments regarding an APDU message
+ * @return -ENODEV	OP-TEE device not available
+ * @return -EPERM	Access permission failure
+ * @return -EAGAIN	Initialization mutex not available
```
etienne-lms, line 42: ` * @return -EAGAIN	Mutex consumer count overflow`

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+		    unsigned char *src, size_t src_len,
+		    unsigned char *dst, size_t *dst_len);
+
+/**
+ * se_scp03_enable()  - Enable the SCP03 protocol
+ *
+ * Configures the Secure Element to enable the SCP03 protocol using the keys
+ * stored in its non volatile memory
+ *
+ * @return -EINVAL	On error
```
etienne-lms, line 58: This function can also return `-EAGAIN`, `-ENODEV` and `-EPERM`.
Ditto for `se_scp03_rotate_keys()`.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+ * @param dst_len	Byte length of reponse buffer @dst
+ *
+ * @return -EINVAL	Invalid arguments regarding an APDU message
+ * @return -ENODEV	OP-TEE device not available
+ * @return -EPERM	Access permission failure
+ * @return -EAGAIN	Initialization mutex not available
+ * @return -EACCESS	Secure element processing error
+ * @return 0		On success
+ */
+
```
etienne-lms, line 46: remove the empty line

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @return -EINVAL	On error
+ * @return 0		On success
+ */
+int se_scp03_enable(void);
+
+/**
+ * se_scp03_rotate_keys()  - Rotate the default SCP03 keys
+ *
+ * Generates the board specific SCP03 keys and writes them in the Secure Element
```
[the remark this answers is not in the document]
ldts: um ok, I can do that but this IMO is common knowledge...

libseteec/src/se.c
```diff
@@ -0,0 +1,171 @@
...
+
+	if (TEEC_InvokeCommand(&apdu_ta_ctx.session,
+			       TA_CMD_TXRX_APDU_RAW_FRAME, &op, &origin)) {
+		EMSG("TEEC invoke command failed, -EINVAL");
+		*dst_len = 0;
+		return -EINVAL;
+	}
+
+	*dst_len = op.params[3].tmpref.size;
+	if (*dst_len)
```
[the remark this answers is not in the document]
ldts: why do you need the logic here? this is not a driver...the actual driver can determine if there is an error and set dst_len to zero. I thought it was cleaner this way. But I can change it if you insist.

libseteec/src/se.c
```diff
@@ -0,0 +1,171 @@
...
+	TEEC_Operation op = { 0 };
+	uint32_t origin = 0;
+	int ret = 0;
+
+	ret = se_init(&uuid, &scp03_ta_ctx);
+	if (ret)
+		return ret;
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_VALUE_INPUT, 0, 0, 0);
+	op.params[0].value.a = 1;
```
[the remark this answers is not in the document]
ldts: right, but the proper way for doing this would be to commit the change to OP-TEE first and then update the pta_scp03.h...but sure I can pick this up.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+/* As described in Smart Card Standard ISO7816-4 5.3 ADPU message structure */
+enum se_apdu_type {
+	SE_APDU_CASE_1 = 0x0,
+	SE_APDU_CASE_2,
+	SE_APDU_CASE_2E,
+	SE_APDU_CASE_3,
+	SE_APDU_CASE_3E,
+	SE_APDU_CASE_4,
+	SE_APDU_CASE_4E,
+	SE_APDU_CASE_INVALID = 0xFF,
```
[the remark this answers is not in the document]
ldts: nah it is useful - maybe this is not the best name possible but passing an _invalid case_ will cause the driver to try to identify the case from the frame itself.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <unistd.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* As described in Smart Card Standard ISO7816-4 5.3 ADPU message structure */
+enum se_apdu_type {
+	SE_APDU_CASE_1 = 0x0,
```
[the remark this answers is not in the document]
ldts: right, tried to be very explict here but ok

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+#ifndef SE_TEE_H
+#define SE_TEE_H
+
+#include <unistd.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* As described in Smart Card Standard ISO7816-4 5.3 ADPU message structure */
```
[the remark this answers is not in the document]
ldts: ok

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+	SE_APDU_CASE_2E,
+	SE_APDU_CASE_3,
+	SE_APDU_CASE_3E,
+	SE_APDU_CASE_4,
+	SE_APDU_CASE_4E,
+	SE_APDU_CASE_INVALID = 0xFF,
+};
+
+/**
+ * se_apdu_request() - Send an APDU message and get response
```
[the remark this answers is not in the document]
ldts: right, INVALID was doing the no_hint. no_hint a much better name. will change that

libseteec/include/se_tee.h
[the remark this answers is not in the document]
ldts: actually no, the description is good but let me add some more info.
```
/**
 * se_scp03_rotate_keys()  - Rotate the default SCP03 keys
 *
 * Generates the SCP03 unique keys for the board and writes them in the Secure
 * Element non volatile memory.
 *
 * The SCP03 keys must persist also in the secure world across reboots so the
 * keys written in the Secure Element must have been either derived from  some
 * secret or stored in secure storage
 * /
```

libseteec/include/se_tee.h
ldts: we dont know why the SE will be failing to process - could perfectly be EACCES if the session is rejected.

libseteec/include/se_tee.h
ldts: yes but I chose to use this error as it is the most common one - instead of describing all the possible mutex problems...gets tricky

ldts: The pthread_mutex_lock() and pthread_mutex_trylock() functions shall fail if:

       EAGAIN The mutex could not be acquired because the maximum number of recursive locks for mutex has been exceeded.

       EINVAL The mutex was created with the protocol attribute having the value PTHREAD_PRIO_PROTECT and the calling thread's  priority
              is higher than the mutex's current priority ceiling.

       ENOTRECOVERABLE
              The state protected by the mutex is not recoverable.

       EOWNERDEAD
              The mutex is a robust mutex and the process containing the previous owning thread terminated while holding the mutex lock.
              The mutex lock shall be acquired by the calling thread and it is up to the new owner to make the state consistent.

ldts: do we need all of these unlikely errors in the header? seems excessive...

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @param type		Type of the APDU command
+ * @param hdr		Pointer to APDU message header
+ * @param hdr_len	Byte length of message header @hdr
+ * @param src		Pointer to APDU message payload
+ * @param src_len	Byte length of message payload @src
+ * @param dst		Pointer to APDU message reponse buffer
+ * @param dst_len	Byte length of reponse buffer @dst
+ *
+ * @return -EINVAL	Invalid arguments regarding an APDU message
```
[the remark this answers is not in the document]
ldts: but on error, TEE_IOC_INVOKE will return 

static TEEC_Result ioctl_errno_to_res(int err)
{
	switch (err) {
	case ENOMEM:
		return TEEC_ERROR_OUT_OF_MEMORY;
	case EINVAL:
	default:
		return TEEC_ERROR_GENERIC;
	}
}

libseteec/include/se_tee.h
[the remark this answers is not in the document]
ldts: what about:
```
 * se_scp03_rotate_keys_and_enable() - Rotate the default SCP03 keys and Enable
 * the SCP03 session
 *
 * Generates the SCP03 unique keys for the board and writes them in the Secure
 * Element non volatile memory.
 *
 * The SCP03 session with the Secure Element is kept enabled.
 *
 * The SCP03 keys must persist also in the secure world across reboots so the
 * keys written in the Secure Element must have been either derived from  some
 * secret or stored in secure storage.
```

libseteec/include/se_tee.h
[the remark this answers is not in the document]
ldts: but will use the name you proposed further down in the review. much better choice.

libseteec/include/se_tee.h
etienne-lms: Agree. You legitimately prepared the mutex so this only reason `pthread_mutex_lock()` could fail is with EAGAIN.

libseteec/include/se_tee.h
etienne-lms: If there's no SE device or PTA fails to connect to the SE for some _TEE or Secure element processing error_, the failure is not due to an access permission issue.

libseteec/include/se_tee.h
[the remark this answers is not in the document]
etienne-lms: sorry to insist. I am a newby to SCP03 and I don't know what is key rotation.
From the description, i don't understand why the function is labeled `..._rotate_keys()`. Looking a SCP03 docs I can find any reference to key rotation features. Wwwebing for secure element and key rotation information does not bring much help. 

The last section in this description is a nice informative notice, but I don't think it explains what this function does.

Do you mean the below?
```
/*
 * se_scp03_rotate_keys_and_enable() - Rotate the default SCP03 keys and Enable
 * the SCP03 session
 *
 * Generates the SCP03 unique keys for the board and writes them in the Secure
 * Element non volatile memory if not already stored. Rotate the key used for SCP03
 * session among generated unique keys.
 *
 * The SCP03 session with the Secure Element is kept enabled, based on the key
 * selected by the key rotation operation.
 */
```
Note the above description is still confusing to me but i'll try to make the effort.
Rotating means you select the next key in the series of the SCP03 keys that have been provisioned, am I right?

libseteec/src/se.c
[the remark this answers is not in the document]
etienne-lms: another way, which looks fine to me is to not test `*dst_len` and let caller decide upon reponse buffer size and content whether the transmission was successful or not.

libseteec/src/se.c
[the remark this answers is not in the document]
etienne-lms: thanks

libseteec/include/se_tee.h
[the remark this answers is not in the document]
ldts: sorry didnt notice this comment. yes so rotation is to do with symmetric cryptography: basically the user replaces an active key with another one(SCP03 uses symmetric crypto btw)
https://cloud.google.com/kms/docs/key-rotation

The actual implementation varies though; for the SE050, since we didnt want to store any keys on the platform, the rotation call just derives the SCP03 keys from the platform HUK and the SE050 unique id. Ideally the user should only do this once the platform has been secured and the HUK is the final secret one ...if the HUK was to change, the SE050 would be bricked.
Other implementations might choose to use secure storage and then allow to rotate at any time - we also have support for that but we didnt push it upstream as for us it wasnt worth the risk of losing the keys.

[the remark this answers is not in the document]
etienne-lms: What you describe here is a key generation scheme, being based on derivation, rotation or whatever, this remain generating the key to be used for some service. Again I think the description here should explain what rotation is: re-generating a new session key based on some derivation schemes known from client (TEE) and service (SE device).

ldts:
@etienne-lms I opted for simplifying the error messages returned by these APIs - I saw no clear benefit on obscuring the code when debugging issues will most likely require rebuilding it anyways.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,84 @@
...
+ */
+enum se_apdu_type {
+	SE_APDU_CASE_1,
+	SE_APDU_CASE_2,
+	SE_APDU_CASE_2E,
+	SE_APDU_CASE_3,
+	SE_APDU_CASE_3E,
+	SE_APDU_CASE_4,
+	SE_APDU_CASE_4E,
+	SE_APDU_NO_HINT,
```
etienne-lms, line 30: NO_HINT should be in 1st position IMO.
It would make more sense if the spec evolves, and a CASE_5 in introduced.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,84 @@
...
+ * se_scp03_rotate_keys_and_enable() - Attempt to replace the active SCP03 keys
+ * and enable the SCP03 session.
+ *
+ * Generates secure keys for the board and writes them in the Secure Element non
+ * volatile memory. Then re-enables the session.
+ *
+ * Depending on the actual implementation, this call might only generate new
+ * keys once. This could be the case when the board is placed in a secure
+ * state and the keys are derived from the HUK instead of read from secure
+ * storage.
```
etienne-lms, line 73: this description goes far beyond what this API function is expected to describe. Don't describe a device specific behavior in this header file exported by the library.

libseteec/src/se.c
```diff
@@ -0,0 +1,119 @@
...
+		if (!res) {
+			res = TEEC_OpenSession(&ctx->context, &ctx->session,
+					       &ctx->uuid, TEEC_LOGIN_PUBLIC,
+					       NULL, NULL, NULL);
+			if (!res)
+				ctx->open = true;
+		}
+	}
+
+	return (pthread_mutex_unlock(&ctx->lock) || res) ? false : true;
```
etienne-lms, line 60: `return !pthread_mutex_unlock(&ctx->lock) && !res;`

libseteec/src/se.c
```diff
@@ -0,0 +1,119 @@
...
+
+	if (!open_session(&scp03_ta_ctx))
+		return -EPERM;
+
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_VALUE_INPUT, 0, 0, 0);
+	op.params[0].value.a = cmd;
+
+	if (TEEC_InvokeCommand(&scp03_ta_ctx.session,
+			       PTA_CMD_ENABLE_SCP03, &op, NULL))
+		return -EPERM;
```
etienne-lms, line 75: Why always `-EPERM`? It should report a real permission error, not a communication issue or the fact the companion OP-TEE does not embedded the expected PTA.
For example, if mutex_lock returns EAGAIN, at line 68 will return  EPERM. A test enviroment or a real application may misbehave if getting the wrong error code.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @param type		Type of the APDU command.
+ * @param hdr		Pointer to APDU message header.
+ * @param hdr_len	Byte length of message header @hdr.
+ * @param src		Pointer to APDU message payload.
+ * @param src_len	Byte length of message payload @src.
+ * @param dst		Pointer to APDU message reponse buffer.
+ * @param dst_len	Byte length of reponse buffer @dst.
+ *
+ * @return -EPERM	Operation not permitted.
```
jenswi-linaro, line 44: Error codes in POSIX Pthreads are returned as plain errno values without an explicit negation in front.
In kernels it's common to return like you're doing here, but this is user space so we should probably try follow the established convention here instead.
Or do you have some specific reason why you'd like to return negative values?

libseteec/src/se.c
ldts: If it misbehaves it is because it is not reading this API. Misbehaving assumes the handling of the error code is wrong. Eperm is correct to indicate that the call failed. To me the finer granularity is not worth the effort in time nor maintenance

libseteec/include/se_tee.h
ldts: You are right. Will fix

etienne-lms: i think it is me that pointed @ldts forgot the minus signs.
my bad :some-emoji-telling-i-m-sorry:

libseteec/include/se_tee.h
ldts: this is a very dangerous operation that can brick boards...I think the extra information might be beneficial

libseteec/src/se.c
ldts: but yes EPERM is probably misleading...will use new SE error codes instead.

libseteec/include/se_tee.h
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * se_scp03_rotate_keys_and_enable() - Attempt to replace the active SCP03 keys
+ * and enable the SCP03 session.
+ *
+ * Generates secure keys for the board and writes them in the Secure Element non
+ * volatile memory. Then re-enables the session.
+ *
+ * Depending on the actual implementation, this call might only generate new
+ * keys once. This could be the case when the board is placed in a secure
+ * state and the keys are derived from the HUK instead of read from secure
```
etienne-lms, line 82: "... instead of **being** read ..."

etienne-lms: If keys are derived from the HUK, as far as HUK does not change, derived keys will be the same. The fact HUK may change is out of scope I think. Default generation of the RPMB key use the same scheme. If HUK changes, RPMB will be lost its accesses bricked. How different SCP03 keys are?

libseteec/src/apdu_pta.h
```diff
@@ -0,0 +1,33 @@
...
+			{ 0x9b, 0x9d, 0x0f, 0x3f, 0xc9, 0x46, 0x8f, 0x50 } }
+/*
+ * ADPU based communication with a secure element
+ *
+ * Background Info: https://en.wikipedia.org/wiki/ISO/IEC_7816
+ *
+ * [in]  value[0].a           APDU request type.
+ * [in]  memref[1].buffer     APDU header.
+ * [in]  memref[1].size       APDU header len.
+ * [in]  memref[2].buffer     source data (raw APDU request frame data).
```
etienne-lms, line 20: for consistency:
s/source/Source/g
s/response/Response/g

libseteec/include/se_tee.h
ldts: > "... instead of **being** read ..."

how sure are you about that correction?

ldts: > If keys are derived from the HUK, as far as HUK does not change, derived keys will be the same. The fact HUK may change is out of scope I think. Default generation of the RPMB key use the same scheme. If HUK changes, RPMB will be lost its accesses bricked. How different SCP03 keys are?

is exactly the same case...however RPMB is a better known use case

libseteec/src/apdu_pta.h
ldts: ok. is this the last set of changes or should I hold back?

ldts:
@etienne-lms as with  https://github.com/OP-TEE/optee_os/pull/4935 I'll wait for a few days for all comments at this point so we can wrap it up. Please let me know when you are done with the review so I push one last time.

libseteec/include/se_tee.h
etienne-lms: > is exactly the same case...however RPMB is a better known use case

Known from you maybe, not from me (the fact a huk may change).
I meant there is no such warning inline description in OP-TEE libs/suppliant when accessing RPMB. So I think the Secure Element interface API from optee_client should not specifically detail how maybe a platform handles SE keys, known that each platform may have specific constraints and pitfalls on how keys are provisioned, derived, revoked, ... . Optee_docs is by far a better place to explain specifc constraints, IMO.

libseteec/src/apdu_pta.h
etienne-lms: I can't tell by now. 
I do appreciate time spent by reviewers on my changes to have them merged in the best possible state. I hope you do also.

ldts: @etienne-lms absolutely!!

ldts: hi @etienne-lms I added a diagram to the PR (https://github.com/OP-TEE/optee_client/pull/290 ) to bring in some clarity to what we are doing

ldts:
a week with no comments. can we merge please?

ldts:
if the interface should expand - I added an additional commit as an example -  should we keep separate simple PTAs that interface to crypto_se.h or a single one with an expanding set of functions.

I dont think this interface will grow much since _everything_ can be done using the APDU superset but I thought I'd ask.

ldts:
rebased

libseteec/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+set(MAJOR_VERSION 0)
+set(MINOR_VERSION 1)
+set(PATCH_VERSION 0)
+
+set(PROJECT_VERSION "${MAJOR_VERSION}.${MINOR_VERSION}.${PATCH_VERSION}")
+
+################################################################################
+# Packages
+################################################################################
+find_package(Threads REQUIRED)
```
jforissier, line 12: Not needed anymore

libseteec/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+	PUBLIC include
+	PRIVATE src
+)
+
+target_include_directories(teec
+	PUBLIC include
+)
+
+target_link_libraries (seteec
+	PRIVATE pthread
```
jforissier, line 61: Not needed anymore

libseteec/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+	PRIVATE teec
+	PRIVATE m
```
jforissier, line 63: Is `libm` required?

jforissier (commented):
With my comments below addressed:

libseteec/CMakeLists.txt
ldts: pthread is not needed? why not?

libseteec/CMakeLists.txt
ldts: could you explain why please?

libseteec/CMakeLists.txt
jforissier: Oh sorry, I thought you had removed all pthread calls but I see you're still using a mutex to ensure single initialization. That's fine.
`pthread_once()` would be slightly better for this purpose but the difference is not massive ;-)

libseteec/CMakeLists.txt
ldts: no, I dont think so, I will remove

libseteec/Makefile
```diff
@@ -0,0 +1,71 @@
...
+
+LIBSETEEC_SRCS	= se.c
+
+LIBSETEEC_INCLUDES	= ${CURDIR}/include
+LIBSETEEC_INCLUDES 	+= ${CURDIR}/../public
+
+LIBSETEEC_CFLAGS	:= $(addprefix -I, $(LIBSETEEC_INCLUDES)) \
+			$(CFLAGS) -D_GNU_SOURCE -fPIC
+
+LIBSETEEC_LFLAGS	:= $(LDFLAGS) -L$(OUT_DIR)/../libteec -lteec
```
jforissier, line 32: `-lpthread` seems to be missing here, but there is a more fundamental issue with existing code. In theory any code using pthreads (here `libteec`, `tee-supplicant`, and now `libseteec`) should have `-pthreads` in C flags and link flags. Note that when `-pthread` is used at link time, `-lpthread` is not needed. This is because `-pthreads` defines some macros such as `_REENTRANT` etc.
I suggest adding `-lpthread` here for consistency and we will deal with the theoretical `-pthreads` issue later (since we have not yet seen any issue with the lack of this flag).

ldts: funny, was just looking into the exact same thing...and I was wondering how pkcs11 can work since I just borrowed the makefile from there. sure will add!

jforissier:
@etienne-lms do you want to provide an ack/r-b? since you commented

etienne-lms:
ok, i'll make another review round.

Note there is something that bothers me. I don't want to block this P-R with that, so i stated it here just to get some feedback. The APIs defined for this library do not identify the target secure element to connect to. What if there are several SEs under OP-TEE core control? Should there be a notion a "session" to the secure element provided by the API?

ldts:
right, I think the only use case where that would make sense is on a platform that needs a number of asym key-pairs that exceeds the memory in the SE (~64KB in the SE051).

Notice that for performance reasons, only ECC/RSA sign/verify (perhaps RNG) are usually configured to be provided by the NXP SE driver: the cost of transferring data on the i2c/spi bus plus the SCP03 encryption/decryption is not neglect-able. Other crypto operations can go to CAAM (on iMX or tomcrypt). That is also why I opted for not pushing more symmetric ciphers, HMAC and other supported operations (although I could if there is demand for it).

So I think what you say makes sense; I also believe that SE vendors are choosing the NVM memory in their devices for the average user....so while I do agree with extending support for any number of concurrent secure elements, maybe that can be abstracted once the first one is merged? I could try adding support for two SE050s at a later time (also if there is demand for it...maybe SE vendors increase the NVM as time goes by)

if you think what I tried to explain above is enough for adding a session field to the APDU interface then I can do it of course.

libseteec/src/apdu_pta.h
```diff
@@ -0,0 +1,33 @@
...
+ * [out] memref[3].size       Response length (raw APDU respose length).
+ *
+ * Result:
+ * TEE_SUCCESS - Invoke command success
+ * TEE_ERROR_BAD_PARAMETERS - Incorrect input param
+ * TEE_ERROR_NOT_IMPLEMENTED - Invoke command not implemented
+ * TEE_ERROR_GENERIC - Invoke command failure
+ */
+#define PTA_CMD_TXRX_APDU_RAW_FRAME		0
+
```
etienne-lms, line 32: type hints missing.
Update this file from pta_adpu.h proposed in https://github.com/OP-TEE/optee_os/pull/4935.

etienne-lms (changes requested):
Should rename _adpu_ta.h_ to _pta_apdu.h_ and update its content.
It should be enough (IMO) to add an inline comment stating type IDs exposed by the PTA and by the lib are the same. Otherwise, eitehr assert at build time, or add a (somewhat useless) conversion as done in optee_os between PTA API and crypto_se API.

etienne-lms:
> if you think what I tried to explain above is enough for adding a session field to the APDU interface then I can do it of course.

No need to add a session field. We would rather need a clean API for session management. Let's see that when there's a need and the lib API will likely evolve with new API functions.

libseteec/src/se.c
```diff
@@ -0,0 +1,149 @@
...
+SE_RV se_apdu_request(enum se_apdu_type apdu_type,
+		      unsigned char *hdr, size_t hdr_len,
+		      unsigned char *src, size_t src_len,
+		      unsigned char *dst, size_t *dst_len)
+{
+	uint32_t type = PTA_APDU_TXRX_CASE_NO_HINT;
+	TEEC_Operation op = { 0 };
+
+	if (!open_session(&apdu_ta_ctx))
+		return SER_CANT_OPEN_SESSION;
```
etienne-lms, line 99: open session only after type sanitation.

etienne-lms (commented):

ldts:
@etienne-lms awesome. thanks

libseteec/include/se_tee.h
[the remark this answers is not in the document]
ldts: well no, key rotation is key rotation...key generation is a different concept.

ldts:
@jforissier I think this PR is now reviewed

jforissier:
@ldts yes, thanks.

PR #292, opened 2021-11-30
tee-supplicant: recv_with_out_flags(): check EINTR

jenswi-linaro:
Adds special treatment if recvmsg() returns error with EINTR, in that
case the was the syscall interrupted by delivery of a signal and we
should just try again with the same arguments.

### DIFF
--- a/tee-supplicant/src/tee_socket.c
+++ b/tee-supplicant/src/tee_socket.c
@@ -496,13 +496,26 @@ static TEEC_Result tee_socket_send(size_t num_params,
 static ssize_t recv_with_out_flags(int fd, void *buf, size_t len, int inflags,
 				   int *out_flags)
 {
-	struct iovec iov = { .iov_base = buf, .iov_len = len, };
-	struct msghdr msg = { .msg_iov = &iov, .msg_iovlen = 1, };
-	ssize_t r = recvmsg(fd, &msg, inflags);
+	ssize_t r = 0;
 
-	if (r >= 0)
+	while (true) {
+		struct iovec iov = { .iov_base = buf, .iov_len = len, };
+		struct msghdr msg = { .msg_iov = &iov, .msg_iovlen = 1, };
+
+		r = recvmsg(fd, &msg, inflags);
+		if (r < 0) {
+			/*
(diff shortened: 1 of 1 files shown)


jforissier (approved):

jenswi-linaro:
Tag applied

jenswi-linaro:
I'm reproducing the error on my Hikey board now.

jenswi-linaro:
Tag applied to "tee-supplicant: recv_with_out_flags(): check EINTR", note that the check for EINTR at end of read_with_timeout() was removed. I've also added one more commit.

This together with https://github.com/OP-TEE/optee_test/pull/571 should take care of the recent sporadic failures in test case 2004.

jenswi-linaro:
Updated the commit message in "tee-supplicant: read_with_timeout(): fix non-blocking peeking"

tee-supplicant/src/tee_socket.c
```diff
@@ -588,8 +588,17 @@ static TEEC_Result read_with_timeout(int fd, void *buf, size_t *blen,
 	return TEEC_SUCCESS;
 
 err:
-	if (e == EAGAIN || e == EWOULDBLOCK)
+	if (e == EAGAIN || e == EWOULDBLOCK) {
+		/*
+		 * If *blen is supplied as 0 then we're not supposed wait
+		 * for data so the call to poll has been skipped. In case there
+		 * is no data available recvfrom() will return an error with
```
etienne-lms, line 595: The implementation uses `recvmsg()`, not `recvfrom()`. Prefer refer to the former.

etienne-lms:
> (...) note that the check for EINTR at end of read_with_timeout()

+1

For commit "tee-supplicant: read_with_timeout(): fix non-blocking peeking':
Minor comment + typo in commit log : "If there is none avail**able** it just also return**s** success.".
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` once addressed.

I would also rephrase, in commit log:
"... check how much data is available **~~and return~~ , set** how much in *blen and return success."

jenswi-linaro:
Updated as requested and applied tag.

PR #297, opened 2022-01-11
libteec: copy out partial shadow buffer

jenswi-linaro:
Without dynamic shared memory enabled (CFG_CORE_DYN_SHM=n) temporary
memrefs uses a shadow buffer to transport data to and from the kernel.

Prior to this patch in case the out length of such a buffer is larger
than the initial size then no copy out is performed even though some
relevant data may be available. Fix this by copying out as much data as
the shadow buffer can hold.

Fixes: https://github.com/OP-TEE/optee_os/issues/5101

### DIFF
--- a/libteec/src/tee_client_api.c
+++ b/libteec/src/tee_client_api.c
@@ -48,6 +48,8 @@
 
 #include "teec_benchmark.h"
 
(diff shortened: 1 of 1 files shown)


jforissier:
`Tested-by: Jerome Forissier <jerome@forissier.org>`

PR #299, opened 2022-01-31
More env support

kubiko:
Add support to define tee supplicant paths at the run time using env variables.
`OP_TEE_FS_PARENT_PATH`, `OP_TEE_LIB`, and `OP_TEE_PLUGIN_LOAD_PATH`

When trying to run tee supplicant process in confinement, some critical directories needs to be defined dynamically during runtime, rather than at the build time.
This PR enables this run time scenario.

PR #300, opened 2022-02-21
tee-supplicant: fix plugin loading logic

kubiko:
List of the plugins to be loaded by tee-supplicant is constructed based on the content of the `TEE_PLUGIN_LOAD_PATH` directory. However when plugins are loaded only file name is passed to the `dlopen()` function.
This means that actual plugin is loaded at first hit based on `LD_LIBRARY_PATH` search path.
This could result in loading of completely different file than actually intended.

Solution is to pass full file path to the `dlopen()`, that is `TEE_PLUGIN_LOAD_PATH` + plugin filename.

### DIFF
--- a/tee-supplicant/src/plugin.c
+++ b/tee-supplicant/src/plugin.c
@@ -112,6 +112,7 @@ TEEC_Result plugin_load_all(void)
 	enum plugin_err res = PLUGIN_OK;
 	TEEC_Result teec_res = TEEC_SUCCESS;
 	struct dirent *entry = NULL;
+	char path[PATH_MAX] = { 0 };
 
 	dir = opendir(supplicant_params.plugin_load_path);
 
@@ -124,6 +125,7 @@ TEEC_Result plugin_load_all(void)
 
 	while ((entry = readdir(dir))) {
 		struct plugin *p;
+		int r;
 
 		if (!strcmp(entry->d_name, "..") || !strcmp(entry->d_name, "."))
 			continue;
@@ -135,8 +137,15 @@ TEEC_Result plugin_load_all(void)
 			closedir(dir);
 			return TEEC_ERROR_OUT_OF_MEMORY;
 		}
-
-		res = load_plugin((const char *)entry->d_name, p);
+		r = snprintf(path, sizeof(path), "%s/%s",
+		             supplicant_params.plugin_load_path, entry->d_name);
+		if (r < 0 || r >= (int)sizeof(path)) {
+			EMSG("assemble of full path for plugin <%s> failed",
+			     entry->d_name);
+			closedir(dir);
+			return TEEC_ERROR_GENERIC ;
+		}
+		res = load_plugin(path, p);
 		switch (res) {
 		case PLUGIN_DL_OPEN_ERR:
 			EMSG("open plugin <%s> failed: %s",


kubiko:
Rebasing on maser after #299 was merged in

anisyanka:
```
Reviewed-by:  Aleksandr Anisimov <anisimov.alexander.s@gmail.com>
```

tee-supplicant/src/plugin.c
```diff
@@ -136,7 +137,8 @@ TEEC_Result plugin_load_all(void)
 			return TEEC_ERROR_OUT_OF_MEMORY;
 		}
 
-		res = load_plugin((const char *)entry->d_name, p);
+		snprintf(path, sizeof(path), "%s/%s", supplicant_params.plugin_load_path, (const char *)entry->d_name);
```
jforissier, line 140: The cast is not needed I think.

jforissier (commented):
One minor comment (please check), in any case:

tee-supplicant/src/plugin.c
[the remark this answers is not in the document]
kubiko: error handling added

tee-supplicant/src/plugin.c
```diff
@@ -135,8 +137,15 @@ TEEC_Result plugin_load_all(void)
 			closedir(dir);
 			return TEEC_ERROR_OUT_OF_MEMORY;
 		}
-
-		res = load_plugin((const char *)entry->d_name, p);
+		r = snprintf(path, sizeof(path), "%s/%s",
+		             supplicant_params.plugin_load_path, entry->d_name);
+		if (r < 0 || r == sizeof(path)) {
```
jforissier, line 142: `r >= sizeof(path)`

jenswi-linaro:
With @jforissier's comment addressed:

kubiko:
> With @jforissier's comment addressed:
> `Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>`

added

tee-supplicant/src/plugin.c
kubiko: fixed

jforissier:
```
 19252:  src/plugin.c:142:18: error: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Werror=sign-compare]
 19253:    142 |   if (r < 0 || r >= sizeof(path)) {
 19254:        |                  ^~
 19255:  cc1: all warnings being treated as errors
```
This should be `if (r < 0 || r >= (int)sizeof(path))` I suppose.

kubiko:
> ```
>  19252:  src/plugin.c:142:18: error: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Werror=sign-compare]
>  19253:    142 |   if (r < 0 || r >= sizeof(path)) {
>  19254:        |                  ^~
>  19255:  cc1: all warnings being treated as errors
> ```
> 
> 
>     
>  
>   
> 
> This should be `if (r < 0 || r >= (int)sizeof(path))` I suppose.

thank you for that.
I run test build here and that seems to pass

jforissier:
>  I run test build here and that seems to pass

Not surprising, CI builds with many warnings enabled and with `-Werror`. Thanks for the update.

PR #302, opened 2022-03-17
tee-supplicant: add --rpmb-cid command line option

jforissier:
In OP-TEE OS, the RPMB device used for secure storage is selected at
compile time via an integer identifier (CFG_RPMB_FS_DEV_ID). As
mentioned in [1], this ID is assigned by the Linux kernel and is used
when tee-supplicant opens the device on behalf of OP-TEE. There are a
couple of issues with that:

1. U-Boot and Linux may assign a different number to the same RPMB
device. Therefore, the TEE supplicant components in U-Boot and Linux
cannot both trust the ID given by OP-TEE.

2. If a system has several RPMB devices, and even if we ignore removable
ones, there is no guarantee that the devices will always be enumerated
in the same order by the kernel on boot. This results in different
device numbers. I observed this behavior on a Hikey620 board with an
external eMMC module plugged into the micro SD slot. Sometimes the
on-board RPMB (which I don’t use for testing) is /dev/mmcblk0rpmb and
the external one is /dev/mmcblk1rpmb; sometimes it is the other way
around.

In order to remove any ambiguity, introduce a new command line argument
to tee-supplicant: --rpmb-cid <CID>. <CID> is the device identification
register of the eMMC device that OP-TEE should use for RPMB. It is
unique for every flash device. When --rpmb-cid is given, the device
number given by OP-TEE is ignored and the specified device is used
instead. <CID> can be found in sysfs, for example:
```
 # Read the CID of MMC device 0. Its RPMB partition is /dev/mmcblk0rpmb.
 $ cat /sys/class/mmc_host/mmc0/mmc0\:0001/cid
 11010030303847453000e0a18ceb13df
 $
```
Link: https://github.com/OP-TEE/optee_os/blob/3.16.0/mk/config.mk#L159-L162

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -201,6 +201,35 @@ static void close_mmc_fd(int fd)
 	close(fd);
 }
 
+/*
+ * Read @n bytes from @fd, takes care of short reads and EINTR.
+ * Adapted from “Advanced Programming In the UNIX Environment” by W. Richard
+ * Stevens and Stephen A. Rago, 2013, 3rd Edition, Addison-Wesley
+ * (EINTR handling was added)
+ */
+static ssize_t readn(int fd, void *ptr, size_t n)
+{
+	size_t nleft = n;
+	ssize_t nread = 0;
+	uint8_t *p = ptr;
+
+	while (nleft > 0) {
+		if ((nread = read(fd, p, nleft)) < 0) {
+			if (errno == EINTR)
+				continue;
+			if (nleft == n)
+				return -1; /* error, nothing read, return -1 */
+			else
+				break; /* error, return amount read so far */
+		} else if (nread == 0) {
+			break; /* EOF */
+		}
+		nleft -= nread;
+		p += nread;
+	}
+	return n - nleft; /* return >= 0 */
+}
+
 /* Device Identification (CID) register is 16 bytes. It is read from sysfs. */
 static uint32_t read_cid(uint16_t dev_id, uint8_t *cid)
 {
@@ -220,9 +249,11 @@ static uint32_t read_cid(uint16_t dev_id, uint8_t *cid)
 	}
 
 	for (i = 0; i < 16; i++) {
-		st = read(fd, hex, 2);
-		if (st < 0) {
-			EMSG("Read CID error (%s)", strerror(errno));
+		st = readn(fd, hex, 2);
+		if (st != 2) {
+			EMSG("Read CID error");
+			if (errno)
+				EMSG("%s", strerror(errno));
 			res = TEEC_ERROR_NO_DATA;
 			goto err;
 		}
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -25,6 +25,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <dirent.h>
 #include <fcntl.h>
 #include <linux/types.h>
 #include <linux/mmc/ioctl.h>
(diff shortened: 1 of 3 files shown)


tee-supplicant/src/rpmb.c
```diff
@@ -795,6 +796,55 @@ static uint32_t rpmb_get_dev_info(uint16_t dev_id, struct rpmb_dev_info *info)
...
+	for (id = 0; id < 255; id++) {
+		snprintf(path, sizeof(path),
+			 "/sys/class/mmc_host/mmc%u/mmc%u:0001/cid", id, id);
+
+		fd = open(path, O_RDONLY);
+		if (fd < 0)
+			continue;
+again:
+		len = read(fd, cid, cid_len);
+		if (len == -1 && errno == EINTR)
```
apalos, line 833: I think there's more you need to check here. -1 indicates an error but the read() has no guarantee it's going to read all cid_len in one go.

jenswi-linaro: It sounds like you're asking for something `readn()` by Stevens, Advanced Programming in the UNIX Environment and other books. It might be good to use such a function instead.

jforissier: @jenswi-linaro yes that would be the safest. My hope was that given we're reading from sysfs, we may have some guarantee that the data is returned all at once (and perhaps even that EINTR cannot happen). If I can't find such a guarantee clearly documented somewhere, I will add a `readn()` function.

tee-supplicant/src/rpmb.c
```diff
@@ -276,6 +277,50 @@ static TEEC_Result read_cid(uint16_t dev_id, uint8_t *cid)
...
+ * Maximum number of MMC devices expected on the system. If more are present,
+ * the one specified by --rpmb-cid might not be found.
+ */
+#define NUM_MMC_DEVICES 15
+
+/*
+ * - If --rpmb-cid is given, find the eMMC RPMB device number with the specified
+ * CID, cache the number, copy it to @nedv_id and return true. If not found
+ * return false.
+ * - If --rpmb-cid is not given, @copy dev_id to @ndev_id and return true.
```
etienne-lms, line 290: s/@copy dev_id to/copy @dev_id to/

tee-supplicant/src/rpmb.c
```diff
@@ -201,37 +202,123 @@ static void close_mmc_fd(int fd)
...
 		return TEEC_ERROR_ITEM_NOT_FOUND;
+	st = readn(fd, cid, 32);
+	if (st != 32) {
+		EMSG("Read CID error");
+		if (errno)
+			EMSG("%s", strerror(errno));
+		res = TEEC_ERROR_NO_DATA;
+		goto err;
 	}
+err:
```
etienne-lms, line 255: Add `res = TEEC_SUCCESS;` before branch label `err` and rename this one to `out`.

jforissier: Oops! Will fix.

tee-supplicant/src/rpmb.c
```diff
@@ -201,37 +202,124 @@ static void close_mmc_fd(int fd)
...
+{
+	size_t nleft = n;
+	ssize_t nread = 0;
+	uint8_t *p = ptr;
+
+	while (nleft > 0) {
+again:
+		if ((nread = read(fd, p, nleft)) < 0) {
+			if (errno == EINTR)
+				goto again;
```
etienne-lms, line 221: nitpicking: `continue` would be fine.

etienne-lms (commented):

tee-supplicant/src/rpmb.c
```diff
@@ -201,37 +202,124 @@ static void close_mmc_fd(int fd)
...
+			return TEEC_ERROR_NO_DATA;
+	}
+	return TEEC_SUCCESS;
+}
+
+/*
+ * Maximum number of MMC devices expected on the system. If more are present,
+ * the one specified by --rpmb-cid might not be found.
+ */
+#define NUM_MMC_DEVICES 15
```
jenswi-linaro, line 285: I agree that 15 seems to be a very large number in this context, but still annoying to have such limits.
If you can find a way to avoid this limit it would be welcome, but no big deal if it's too hard to fix at the moment.

tee-supplicant/src/rpmb.c
```diff
@@ -201,37 +202,124 @@ static void close_mmc_fd(int fd)
...
+		} else if (nread == 0) {
+			break; /* EOF */
+		}
+		nleft -= nread;
+		p += nread;
+	}
+	return n - nleft; /* return >= 0 */
+}
+
+static TEEC_Result read_cid_str(uint16_t dev_id, char cid[33])
```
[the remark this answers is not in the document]
jforissier: I will introduce a define.

tee-supplicant/src/rpmb.c
jforissier: The other way I can think about is to enumerate the sysfs entries with `readdir()`. That will be more complicated of course but let me give it a try.

jforissier:
PR reworked to enumerate sysfs entries with `readdir()`. I squashed the review fixups as well as a couple of bug fixes I found when testing again with a real eMMC on HiKey.

tee-supplicant/src/rpmb.c
```diff
@@ -304,6 +305,70 @@ static TEEC_Result read_cid(uint16_t dev_id, uint8_t *cid)
...
+		fprintf(stderr, "RPMB device %s is at /dev/mmcblk%urpmb\n",
+			cid, id);
+		found = true;
+		break;
+	}
+
+	closedir(dir);
+
+	if (!found)
+		err = true;
```
muvarov, line 365: err is not used later in function code.

jenswi-linaro:

jforissier:
@etienne-lms note I have not kept your R-b in the last two commits since the code has changed quite a bit.

tee-supplicant/src/rpmb.c
```diff
@@ -66,7 +66,8 @@ struct rpmb_req {
 
 /* Response to device info request */
 struct rpmb_dev_info {
-	uint8_t cid[16];
+#define RPMB_CID_SZ 16
```
etienne-lms, line 70: prefer have this macro definition outside `struct rpmb_dev_info` definition.

tee-supplicant/src/rpmb.c
```diff
@@ -304,6 +305,70 @@ static TEEC_Result read_cid(uint16_t dev_id, uint8_t *cid)
...
+ * return false.
+ * - If --rpmb-cid is not given, copy @dev_id to @ndev_id and return true.
+ */
+static bool remap_rpmb_dev_id(uint16_t dev_id, uint16_t *ndev_id)
+{
+	TEEC_Result res = TEEC_ERROR_GENERIC;
+	static bool found = false;
+	static bool err = false;
+	static uint16_t id = 0;
+	char cid[33] = { 0 };
```
etienne-lms, line 320: s/33/CID_STR_SZ + 1/

tee-supplicant/src/rpmb.c
```diff
@@ -304,6 +305,70 @@ static TEEC_Result read_cid(uint16_t dev_id, uint8_t *cid)
...
+			err = true;
+			break;
+		}
+		id = (uint16_t)num;
+		res = read_cid_str(id, cid);
+		if (res)
+			continue;
+		if (strcmp(cid, supplicant_params.rpmb_cid))
+			continue;
+		fprintf(stderr, "RPMB device %s is at /dev/mmcblk%urpmb\n",
```
etienne-lms, line 356: s/stderr/stdout/ or use `IMSG()` or `DMSG()`?

etienne-lms (commented):
`Reviewed-by: Etienne Carriere <etienne.carriere@linaro.org>` for commits
"tee-supplicant: rpmb: read CID in one go"
"tee-supplicant: add --rpmb-cid command line option"
with minor comments addressed.

tee-supplicant/src/rpmb.c
jforissier: Right, forgot that one

PR #306, opened 2022-04-18
tee-supplicant: Support multiple ta load paths

chaohub:
CFG_TEE_CLIENT_LOAD_PATH can have multiple paths separated by :


tee-supplicant/src/teec_ta_load.c
```diff
@@ -172,19 +172,48 @@ static int try_load_secure_module(const char* prefix,
...
+	int res = 0;
+
+	multi_path = strdup(multi_prefix);
+	if (!multi_path) {
+		EMSG("out of memory");
+		return -1;
+	}
+
+	tofree = multi_path;
+	while ((single_path = strsep(&multi_path, ":")) != NULL) {
```
jenswi-linaro, line 193: This parsing could be done once at startup in order to avoid doing this `strdup()`, `strsep()`, `free()` sequence each time a TA is to be loaded.

tee-supplicant/src/teec_ta_load.c
```diff
@@ -176,15 +176,13 @@ int TEECI_LoadSecureModule(const char* dev_path,
...
-				     dev_path, destination, ta, ta_size);
-	if (res != TA_BINARY_NOT_FOUND)
-		return res;
-#endif
-
-	return try_load_secure_module(TEEC_LOAD_PATH,
-				      dev_path, destination, ta, ta_size);
+	for (int i = 0; i < num_ta_prefix; i++) {
+		res = try_load_secure_module(ta_prefix[i],
+				dev_path, destination, ta, ta_size);
```
etienne-lms, line 183: fix indentation
```c
		res = try_load_secure_module(ta_prefix[i], dev_path,
					     destination, ta, ta_size);
```

tee-supplicant/src/teec_ta_load.h
```diff
@@ -31,6 +31,11 @@
 #define TA_BINARY_FOUND 0
 #define TA_BINARY_NOT_FOUND -1
 
+/* support multiple TA load path */
```
etienne-lms, line 34: nitpicking: s/support/Support/

tee-supplicant/src/teec_ta_load.h
```diff
@@ -31,6 +31,11 @@
...
+#define MAX_TA_PREFIX 8
```
etienne-lms, line 35: It's likely there will be less than 8 alternate dir paths, however it would be more flexible to remove that limitation and allocated `ta_prefix[]` dynamically.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -67,6 +67,9 @@
 #define RPC_BUF_SIZE	(sizeof(struct tee_iocl_supp_send_arg) + \
 			 RPC_NUM_PARAMS * sizeof(struct tee_ioctl_param))
 
+char *ta_prefix[MAX_TA_PREFIX] = {0};
+int num_ta_prefix = 0;
```
etienne-lms, line 71: You can (should) remove the initialization value when: global variables are default zero-initialized.

etienne-lms (changes requested):
Fix commit message to also mention `TEEC_LOAD_PATH` and add your `Signed-off-by` tag.

tee-supplicant/src/tee_supplicant.c
chaohub: I will remove the values.

tee-supplicant/src/teec_ta_load.h
chaohub: Pre-allocate enough entries is efficient if the number of entries is not large. I doubt the practicality for cases with more than 8 paths. Adding code and runt time dynamic allocation for virtually non-exist case, it's kind of over kill.

etienne-lms: It would be better to not have built-in limitation if not strictly needed (IMHO).
I'll rely on maintainers point of view.

chaohub: That's fair. Let the maintainers to pick which way to go.

jenswi-linaro: Please pick a dynamic solution.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -762,30 +769,81 @@ int main(int argc, char *argv[])
...
+	ta_prefix = (char **)malloc(sizeof(char *) * num_ta_prefix);
+	if (!ta_prefix) {
+		EMSG("out of memory");
+		goto exit;
+	}
+
+	i = 0;
+#ifdef TEEC_TEST_LOAD_PATH
+	test_ta_prefix_multipath = strdup(TEEC_TEST_LOAD_PATH);
+	if (!test_ta_prefix) {
```
jforissier, line 802: `test_ta_prefix_multipath`

tee-supplicant/src/tee_supplicant.c
```diff
@@ -762,30 +769,81 @@ int main(int argc, char *argv[])
...
+		EMSG("out of memory");
+		goto exit;
+	}
+	while ((temp = strsep(&test_ta_prefix_multipath, ":")) != NULL) {
+		ta_prefix[i++] = temp;
+	}
```
jforissier, line 808: `ta_prefix[i]` may be an empty string if multiple colons are found in a row.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -762,30 +769,81 @@ int main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	/* Support multiple ta load paths */
```
jforissier, line 772: I'd rather have a helper function to parse the paths. There is duplication.

config.mk
```diff
@@ -31,6 +31,8 @@ CFG_TEE_CLIENT_LOG_FILE ?= $(CFG_TEE_FS_PARENT_PATH)/teec.log
 
 # CFG_TEE_CLIENT_LOAD_PATH
 #   The location of the client library file.
```
jforissier, line 33: This line is just wrong, it should be removed. The same error is present in `CMakeLists.mk`.

tee-supplicant/src/teec_ta_load.c
```diff
@@ -176,15 +176,13 @@ int TEECI_LoadSecureModule(const char* dev_path,
...
-	res = try_load_secure_module(TEEC_TEST_LOAD_PATH,
-				     dev_path, destination, ta, ta_size);
-	if (res != TA_BINARY_NOT_FOUND)
-		return res;
-#endif
-
-	return try_load_secure_module(TEEC_LOAD_PATH,
-				      dev_path, destination, ta, ta_size);
+	for (int i = 0; i < num_ta_prefix; i++) {
+		res = try_load_secure_module(ta_prefix[i],
```
jforissier, line 182: Note, the documentation for `try_load_secure_module()` lacks the description of its first two arguments.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -762,30 +769,81 @@ int main(int argc, char *argv[])
...
 
 	while (!arg.abort) {
 		if (!process_one_request(&arg))
 			arg.abort = true;
 	}
 
+exit:
+	free(test_ta_prefix_multipath);
+	free(ta_prefix_multipath);
+	free(ta_prefix);
```
jforissier, line 846: I'm not convinced this kind of cleanup is desirable immediately before exiting.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -701,6 +704,10 @@ int main(int argc, char *argv[])
 	int e = 0;
 	int long_index = 0;
 	int opt = 0;
+	char *test_ta_prefix_multipath = NULL;
+	char *ta_prefix_multipath = NULL;
```
jforissier, line 708: The purpose of the variables is a bit hard to figure out from their names IMO.

jforissier (commented):
Many comments from me below :wink: I propose the following instead: #311.

PR #309, opened 2022-05-20
tee-supplicant: close shm fd before freeing memory

jenswi-linaro:
The resources of a shm is released in process_free(), this includes the
file descriptor and the memory buffer which was registered. Closing
the file descriptor unregisters the memory buffer.

The memory buffer was, prior to this patch, freed before the file
descriptor was closed. This could lead to another thread reusing this
memory buffer before it has been unregistered. This is normally not a
problem since the buffer will not be read or modified after it has been
freed. However, FF-A mandates that a physical memory isn't registered
already when registering. Son in the case we can have occasional failures.

Fixes: 075c56eebdc9 ("tee_supplicant: add register memory feature")


jforissier (commented):

jenswi-linaro:
Tag applied.

PR #311, opened 2022-06-24
tee-supplicant: support multiple TA load paths

jforissier:
Parse CFG_TEE_CLIENT_LOAD_PATH (TEEC_LOAD_PATH) and TEEC_TEST_LOAD_PATH
as a colon-separated list of paths. This commit also updates the
documentation of CFG_TEE_CLIENT_LOAD_PATH and try_load_secure_module()
which were either lacking or incorrect.

Suggested-by: Chao Chen <ccha@amazon.com>

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -67,6 +67,9 @@
 #define RPC_BUF_SIZE	(sizeof(struct tee_iocl_supp_send_arg) + \
 			 RPC_NUM_PARAMS * sizeof(struct tee_ioctl_param))
 
+char **ta_path;
+char *ta_path_str;
+
 union tee_rpc_invoke {
 	uint64_t buf[(RPC_BUF_SIZE - 1) / sizeof(uint64_t) + 1];
 	struct tee_iocl_supp_recv_arg recv;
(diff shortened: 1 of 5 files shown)


etienne-lms (commented):
LGTM with a minor comment. @jforissier, my R-b tag unless you change the implementation as per @jenswi-linaro suggestion.

etienne-lms (commented):

tee-supplicant/src/tee_supplicant.c
```diff
@@ -693,6 +696,46 @@ static void *thread_main(void *a)
...
+		goto err;
+
+	p = ta_path_str;
+	while (strtok_r(p, ":", &saveptr)) {
+		p = NULL;
+		n++;
+	}
+	n++; /* NULL terminator */
+
+	ta_path = malloc(n * sizeof(char *));
```
jenswi-linaro, line 723: The last element in the array should be zeroed, perhaps `calloc()` here instead would be a better choice.

jenswi-linaro:

tee-supplicant/src/tee_supplicant.c
etienne-lms: I also though it should but since line 730 breaks when `ta_path[n++]` is NULL, last element is properly inited to NULL, IIUC.

PR #315, opened 2022-08-11
tee-supplicant: -d: return after TEE device is opened

jforissier:
This commit addresses a race condition when a kernel module using OP-TEE
is loaded immediately after tee-supplicant is started. To understand the
problem, consider that with a shell background task there is no guarantee
that the service is available to the kernel when the command returns.
So the following:

  tee-supplicant &
  modprobe <some_module>

...may fail when <some_module> invokes the kernel TEE client API (note
that kernel users do NOT wait for the supplicant to be available [1],
contrary to user space clients [1]).

This scenario was addressed by commit 66cdd5db37db ("tee-supplicant: add
daemon mode (-d)"), although the commit description did not explicitly
mention it. With '-d ' the supplicant command would open the device
before returning. Unfortunately, this was inadvertently broken by commit
523d50bdede6 ("tee-supplicant: daemonize before opening a supplicant
device").

Restore the previous behavior while still keeping the open() call in the
child process, by introducing some synchronization between the parent
and the child. A pipe is created and the parent issues a blocking read.
After successfully opening the device the child writes data to the pipe,
thus releasing the parent. If the child crashes or exits before writing
the parent is released with 0 bytes read and exits with an error status.

The daemon() call is replaced by make_daemon() which is copied from
FreeBSD [3] and modified as described above.

Link: [1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/tee/tee_core.c?h=v5.19#n1128
Link: [2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/tee/tee_core.c?h=v5.19#n117
Link: [3] https://github.com/freebsd/freebsd-src/blob/release/13.1.0/lib/libc/gen/daemon.c

### DIFF
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -491,8 +491,8 @@ static int usage(int status)
 {
 	fprintf(stderr, "Usage: tee-supplicant [options] [<device-name>]\n");
 	fprintf(stderr, "\t-h, --help: this help\n");
-	fprintf(stderr, "\t-d, --daemonize: run as a daemon (fork after successful "
-			"initialization)\n");
+	fprintf(stderr, "\t-d, --daemonize: run as a daemon (fork and return "
+			"after child has opened the TEE device or on error)\n");
 	fprintf(stderr, "\t-f, --fs-parent-path: secure fs parent path [%s]\n",
 			supplicant_params.fs_parent_path);
 	fprintf(stderr, "\t-t, --ta-dir: TAs dirname under %s [%s]\n", TEEC_LOAD_PATH,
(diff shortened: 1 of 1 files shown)


tee-supplicant/src/tee_supplicant.c
```diff
@@ -736,9 +736,61 @@ static void set_ta_path(void)
 	exit(EXIT_FAILURE);
 }
 
+/*
+ * Similar to the standard libc function daemon(0, 0) but the parent process
+ * issues a blocking read on pipefd[0] before exiting.
+ * Returns 0 on success, <0 on error.
+ */
+static int make_daemon(int pipefd[2])
```
jenswi-linaro, line 744: Perhaps better to have this in a separate file to keep it simple with license and copyright.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -821,6 +880,15 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if (daemonize) {
```
jenswi-linaro, line 883: We should perhaps call `plugin_load_all()` before this so we can return that error to the parent too.

tee-supplicant/src/tee_supplicant.c
jforissier: TBH I'm not sure it makes much sense, because the code here is actually quite different from the FreeBSD source I mentioned in the commit. In fact the Linux man page for `daemon(3)` pretty much gives the algorithm, which I followed and only checked the FreeBSD code to make sure I did not forget anything. When doing so I ignored the unnecessary code such as the signal handling stuff, not needed because the parent cannot be a session leader so it's exit won't cause the child to receive a SIGHUP. From the man page:

>       The daemon() function is for programs wishing to detach themselves from
>       the controlling terminal and run in the background as system daemons.

`fork();`, `setsid();`

>       If nochdir is zero, daemon() changes the process's current working  di‐
>       rectory to the root directory ("/"); 

`chdir("/");`

>       If noclose is zero, daemon() redirects standard input, standard  output
>       and  standard  error  to  /dev/null; 

`fd = open("/dev/null", O_RDWR);`, `dup2(fd, 0);` etc.

>    This  function forks, and if the fork(2) succeeds, the parent calls _exit(2),

`_exit(0);` (I wrongly used `exit()`, I'll fix)

The rest of the code is the pipe IPC.

So I'd rather remove any reference to FreeBSD from the commit description.

tee-supplicant/src/tee_supplicant.c
jforissier: OK

tee-supplicant/src/tee_supplicant.c
jenswi-linaro: OK

jenswi-linaro (commented):

PR #316, opened 2022-08-12
Simplify CMake library version declarations and configure include_directories for BUILD_INTERFACE and INSTALL_INTERFACE

eaaltonen:
Two CMake fix commits originally from the start of https://github.com/OP-TEE/optee_client/pull/259/ (stale).

* The version handling simplifies the CMake build a bit.
* BUILD_INTERFACE and INSTALL_INTERFACE are related to configuring paths differently for the build tree and install tree (_used if/when a CMake config-file package is added_).

These don't depend on `libuuid`, so can be brought in before those changes.

Refreshed with fixup commits for changes since last year.


eaaltonen:
@etienne-lms: FYI I squashed the fixup (_refresh_) commits, but kept your `Reviewed-by` line.

jforissier:
For commit "Remove setting include_directory for teec": do you mean `libseteec` rather than `teec`?

eaaltonen:
> For commit "Remove setting include_directory for teec": do you mean `libseteec` rather than `teec`?

Re-phrased to
`Don't set teec include_directory from libseteec`
Hopefully that's more clear.

And also added Acked-by.

jforissier:
Would you mind adding a `cmake:` prefix to the last two commits? Thanks.

eaaltonen:
> Would you mind adding a `cmake:` prefix to the last two commits? Thanks.

Not at all; and done. As long as you don't mind the slightly longer summary line.

PR #319, opened 2022-08-26
Add libckteeaclc helper library

eaaltonen:
Add helper librarary `libckteeaclc` containing functions that can be used to generate the hashed UUID of the user or group. These can then be configured to ckteec provided tokens for _Access Control List_ (ACL) based access.

These commits are the cherry-picked and slightly polished commit series from https://github.com/OP-TEE/optee_client/pull/259 (not merged).

This PR depends on https://github.com/OP-TEE/build/pull/583, which makes `libuuid` available in the target file system, and `uuid.pc` available at build time.

- [x] squash fixup commits before merging.


etienne-lms (commented):
This mandates libuuid support is embedded. Should we enforce that at build time? There are environments/platforms that don't care about pkcs11 support and would not expect constraints from an unused feature when building optee_client. Could there be a config switch to build/embed or nor libckteeaclc?

Few minor comments.

eaaltonen:
> Could there be a config switch to build/embed or nor libckteeaclc?

Certainly. However `libuuid` is 21844 bytes on the qemu image so my first proposal was to simply take it in.

If you care to suggest a name for the config switch and think we need one, I can add it.

Makefile
```diff
@@ -41,10 +42,14 @@ build-libckteec: build-libteec
...
 	@$(MAKE) --directory=libckteec --no-print-directory --no-builtin-variables
 
-build-libckteeaclc:
+build-libckteeaclc: check-libuuid
 	@echo "Building libckteeaclc.so"
 	@$(MAKE) --directory=libckteeaclc --no-print-directory --no-builtin-variables
 
+check-libuuid:
+	@echo "Finding uuid.pc"
+	pkg-config --atleast-version=2.34 uuid
```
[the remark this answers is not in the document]
eaaltonen: I've yet to come across a Linux distro without `pkg-config`. Do you know an example?

libckteeaclc/src/cktee_uuid.c
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Vaisala Oyj.
+ */
+
+#include <ckteeaclc.h>
+#include <stdio.h>
+#include <string.h>
+
```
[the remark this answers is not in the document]
eaaltonen: Hmm. I actually got a compile error with this from Makefile build. Since
```
$ pkg-config --cflags uuid
-I/usr/include/uuid
```
It looks like `#include <uuid.h>` is more reliable.

I didn't find any recommendation about order of includes, so just used alphabetical.

eaaltonen:
> Few minor comments.

Address minor comments.

libckteeaclc/include/ckteeaclc.h
```diff
@@ -13,12 +13,58 @@
...
+ * making kernel_namespace_uuid const would probably require initializing with
+ * a magic byte array. The non const version can be initialized with
+ * `uuid_parse`.
+ */
+
+/**
+ * ckteeaclc_lib_init() - Initialize the ckteeaclc library.
+ * Must be called before use.
+ */
+int ckteeaclc_lib_init(void);
```
jforissier, line 41: Not needed if the global `kernel_namespace_uuid` goes (see below)

libckteeaclc/src/cktee_uuid.c
```diff
@@ -0,0 +1,37 @@
...
+	uuid_t g_uuid = { 0 };
+	char gid_buf[CKTEEACLC_L_UUID] = { 0 };
+	int rv = snprintf(gid_buf, CKTEEACLC_L_UUID, "gid=%x", group);
+
+	if (rv < 0)
+		return rv;
+	if (rv >= CKTEEACLC_L_UUID)
+		return -1;
+
+	size_t gid_buf_len = strnlen(gid_buf, CKTEEACLC_L_UUID);
```
jforissier, line 29: Declare all variables a the beginning of the function

libckteeaclc/src/cktee_uuid.c
```diff
@@ -0,0 +1,37 @@
...
+{
+	uuid_t g_uuid = { 0 };
+	char gid_buf[CKTEEACLC_L_UUID] = { 0 };
+	int rv = snprintf(gid_buf, CKTEEACLC_L_UUID, "gid=%x", group);
+
+	if (rv < 0)
+		return rv;
+	if (rv >= CKTEEACLC_L_UUID)
+		return -1;
+
```
jforissier, line 28: Add
```C
	rv = uuid_parse(KERNEL_NAMESPACE, k_uuid);
	if (rv < 0)
		return rv;
```

libckteeaclc/src/cktee_uuid.c
```diff
@@ -0,0 +1,37 @@
...
+uuid_t kernel_namespace_uuid;
+
+int ckteeaclc_lib_init(void)
+{
+	return uuid_parse(KERNEL_NAMESPACE, kernel_namespace_uuid);
+}
+
+int ckteeaclc_group_acl_uuid(char uuid_buf[CKTEEACLC_L_UUID], gid_t group)
+{
+	uuid_t g_uuid = { 0 };
```
jforissier, line 20: Add `uuid_t k_uuid = { 0 };`

libckteeaclc/include/ckteeaclc.h
```diff
@@ -13,12 +13,58 @@
...
 
 #define CKTEEACLC_NO_GROUP ((gid_t)0xFFFFFFFFFFFFFFFF)
+#define KERNEL_NAMESPACE "58ac9ca0-2086-4683-a1b8-ec4bc08e01b6"
+
+/**
+ * TEE Client UUID name space identifier in Linux kernel.
+ *
+ * Remember to call ckteeaclc_lib_init before use.
+ */
+extern uuid_t kernel_namespace_uuid;
```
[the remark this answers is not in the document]
eaaltonen: Not strictly necessary. I was thinking:
* it might be nice to initialize this only once since it's a constant
* we could expose it for other uses (_but we don't have to_)

jenswi-linaro:
Out of curiosity, what is `libckteeaclc` short for?

eaaltonen:
> Out of curiosity, what is libckteeaclc short for?

PKCS#11 names itself Cryptoki from _Cryptographic Token Interface_. The PKCS#11 functions start with `CK_`.
ACL is from _Access Control List_.
I presume `libckteec` is Cryptoki _Trusted Execution Environment_ (TEE) client library.

Hence `libckteeaclc` = _Cryptographic Token Interface Trusted Execution Environment Access Control List client library_ (quite a mouthful).

libckteeaclc/include/ckteeaclc.h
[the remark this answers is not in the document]
eaaltonen: 2nd fixup commit for `Add function to encode a group login based uuid` removes this and the related changes.

libckteeaclc/src/group.c
```diff
@@ -0,0 +1,66 @@
...
+	int grouplistsize = 8; /* initial guess */
+	gid_t *groups = (gid_t *)calloc(grouplistsize, sizeof(gid_t));
+	if (groups == NULL)
+		return E_MEMORY;
+	int ret = getgrouplist(username, group, groups, &grouplistsize);
+
+	if (ret == -1) {
+		gid_t *p_groups = groups;
+
+		if ((groups = reallocarray(groups, grouplistsize, sizeof(gid_t))) == NULL) {
```
jenswi-linaro, line 45: Prefer not having assignments as a side effect of the if check.

libckteeaclc/include/ckteeaclc.h
```diff
@@ -0,0 +1,94 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) Vaisala Oyj.
```
jenswi-linaro, line 3: A year would be nice.

libckteeaclc/src/cktee_uuid.c
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: BSD-2-Clause
+/*
+ * Copyright (c) Vaisala Oyj.
```
jenswi-linaro, line 3: A year would be nice.

jenswi-linaro:
> Hence `libckteeaclc` = _Cryptographic Token Interface Trusted Execution Environment Access Control List client library_ (quite a mouthful).

Thanks for the explanation.

Does this have to be specific to PKCS#11 or could it be thought of as more generic?

eaaltonen:
> Does this have to be specific to PKCS#11 or could it be thought of as more generic?

I suppose it could also be thought as more generic.

It currently provides:
1. helper methods for group membership checks
2. user space calculation of the same `TEE_UUID` as what `drivers/tee/tee_core.c` does

In our use case we use it to:
* initialize/configure the login method to the specific UUID (_for a ckteec token_)
* later login with ACL based method.

If you foresee this used more generically, I suppose `teeaclc` would be more fitting. Shall I make the change?

jforissier:
> I suppose `teeaclc` would be more fitting.

+1 with decoupling from PKCS#11 but I would personally drop the last `c` (client is meaningless in this context).

libckteeaclc/include/ckteeaclc.h
```diff
@@ -38,6 +38,17 @@ extern "C" {
...
+/**
+ * ckteeaclc_try_resolve_group - Try to resolve gid_t for a given `group_name`.
+ *
+ * @param gid_out Ptr to gid result. After the call the value will be either
+ * - Group id or
+ * - CKTEEACLC_NO_GROUP
+ * @param group_name Name of group to resolve.
+ * @return Zero on success, errno otherwise.
+ */
+int ckteeaclc_try_resolve_group(gid_t *gid_out, const char *group_name);
```
jforissier, line 50: What is the added value of this function compared to calling `getgrnam()` or `getgrnam_r()` directly? (note that this function is not reentrant).

eaaltonen: This is called in two places in [teeckcli.cpp](https://github.com/OP-TEE/optee_client/pull/259/files#diff-20c23c25b0e9371d44ed52fadc532984f981051fb3dc60911aad33ee65a9e03a)

for `SO group` and `User group`. For me the cli code was easier to read by using this helper function. I concede it is a very simple one.

eaaltonen:
@jenswi-linaro: added fixup commits for minor comments.

libckteeaclc/src/group.c
```diff
@@ -0,0 +1,69 @@
...
+
+#include <ckteeaclc.h>
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int ckteeaclc_try_resolve_group(gid_t *gid_out, const char *group_name)
+{
+	struct group *group_s = getgrnam(group_name);
```
jforissier, line 14: Unless the library has other reasons for not being re-entrant/thread-safe, I'd rather use `getgrnam_r()` here.

libckteeaclc/src/group.c
```diff
@@ -0,0 +1,64 @@
...
+	cuserid(username);
+	return ckteeaclc_user_is_member_of(username, group);
+}
+
+enum rv_groupmember ckteeaclc_user_is_member_of(const char *username, gid_t group)
+{
+	enum rv_groupmember result = E_MEMORY;
+	int i = 0;
+	int grouplistsize = 8; /* initial guess */
+	gid_t *groups = (gid_t *)calloc(grouplistsize, sizeof(gid_t));
```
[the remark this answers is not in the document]
jenswi-linaro: The cast is still there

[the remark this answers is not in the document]
eaaltonen: Removed in latest push

libckteeaclc/include/ckteeaclc.h
eaaltonen: Mainly since we attempt to map a user provided string (_unvalitated input_) to a group defined in the system. The case that the string does not map to any defined group is conceptually normal. Indeed `getgrnam_r` will return zero in that case.

eaaltonen: The function could also be named:
* `PREFIX_resolve_group`
* `PREFIX_gid_from_name`

for example

libckteeaclc/src/group.c
eaaltonen: Changed to use `getgrnam_r()`

libckteeaclc/include/ckteeaclc.h
jforissier: +1 for `gid_from_name()`

libckteeaclc/include/ckteeaclc.h
```diff
@@ -41,6 +41,10 @@ extern "C" {
 /**
  * ckteeaclc_try_resolve_group - Try to resolve gid_t for a given `group_name`.
  *
+ * Note that zero is returned also when specified the `group_name` does not map
+ * to any specified group. This case must always be checked by comparing if
+ * `git_out` equals CKTEEACLC_NO_GROUP.
+ *
```
jforissier, line 47: That sounds convoluted. Why not return `-ENOENT` instead?

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+static long ckteeaclc_getgr_r_size_max(void)
+{
+	long getgr_r_size_max = sysconf(_SC_GETGR_R_SIZE_MAX);
```
jforissier, line 16: Call this `s` please...

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
...
+	return getgr_r_size_max;
+};
 
 int ckteeaclc_try_resolve_group(gid_t *gid_out, const char *group_name)
 {
-	struct group *group_s = getgrnam(group_name);
+	struct group grp = { 0 };
+	char *buffer = NULL;
+	struct group *grpResult = NULL;
+	size_t getgr_r_size_max = 0;
```
jforissier, line 28: `s` or `size`

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
...
+	int rv = 0;
+	getgr_r_size_max = ckteeaclc_getgr_r_size_max();
+	buffer = (char *)(malloc(getgr_r_size_max));
```
jforissier, line 31: `calloc()`

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
...
+	if (buffer == NULL)
```
jforissier, line 32: `if (!buffer)`

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
...
+		return errno ? errno : -1;
```
jforissier, line 33: `return -ENOMEM;`

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
...
+	memset(buffer, 0, getgr_r_size_max);
+
+	rv = getgrnam_r(group_name, &grp, buffer, getgr_r_size_max, &grpResult);
 
-	if (!group_s) {
+	free(buffer);
```
jforissier, line 38: Wrong, `grpResult` points into `buffer`

libckteeaclc/src/group.c
```diff
@@ -8,17 +8,41 @@
...
+	}
+	return getgr_r_size_max;
+};
 
 int ckteeaclc_try_resolve_group(gid_t *gid_out, const char *group_name)
 {
-	struct group *group_s = getgrnam(group_name);
+	struct group grp = { 0 };
+	char *buffer = NULL;
+	struct group *grpResult = NULL;
```
jforissier, line 27: No camelCase please. Use same name as in man page: `result`.

jforissier (commented):
This is becoming difficult to review due to the many small fixups, I suggest creating an new PR with the new name and all fixups squashed. A few more comments still.

libckteeaclc/src/group.c
eaaltonen: It does, but we are not using any of the pointers, only `gr_gid`.

libckteeaclc/src/group.c
eaaltonen: Just to double check, is the intent to support only architectures where the address `NULL` is zero?

jforissier: It is a coding style thing. AFAIK they're guaranteed to be equivalent by the C standard, even when `NULL` is non-zero. Have you ever met such an architecture?

libckteeaclc/src/group.c
jforissier: Right, my mistake.

eaaltonen:
> This is becoming difficult to review due to the many small fixups, I suggest creating an new PR with the new name and all fixups squashed.

I opted to squash the fixups pending from 2022-08-31 in the latest push. Here are the changes:

#### Applied fixups and changes

* autosquash
* Rename ckteeaclc to teeacl
* Rename teeacl_try_resolve_group to teeacl_gid_from_name

#### New fixup commits

* getgr_r_size_max→s
* getgr_r_size_max→b_size
* Use calloc
* if (!buffer)
* return -ENOMEM
* grpResult→result
* That sounds convoluted. Why not return -ENOENT instead?

I'll sort the new fixups to their target commits a bit later, if you are satisfied with the changes.

libckteeaclc/include/ckteeaclc.h
eaaltonen: Thanks, I applied this suggestion, see commit `Return -ENOENT if no group is found`.

libckteeaclc/src/group.c
eaaltonen: Applied. I'd actually like to use `alloca` here, but the label states it's machine- and compiler-dependent.

jforissier (commented):
Please do create a new PR.

eaaltonen:
Replaced by https://github.com/OP-TEE/optee_client/pull/323

PR #320, opened 2022-08-31
tee-supplicant: android: make RPMB_EMU a conditional assignment

massonju:
At the moment the RPMB_EMU variable in the Makefile uses a simple assignment and unconditionally sets the variable.

Move it to a conditional assignment and allow users to override it.


jforissier (approved):

Thanks!

massonju:
Tag applied, thanks for the review

PR #323, opened 2022-09-06
Add libteeacl helper library

eaaltonen:
Add helper library libteeacl containing functions that can be used to generate the hashed UUID of the user or group. These can then be configured to ckteec provided tokens for Access Control List (ACL) based access.

These commits are the cherry-picked and slightly polished commit series from #259 (not merged).

This PR depends on OP-TEE/build#583, which makes libuuid available in the target file system, and uuid.pc available at build time.


libckteeaclc/src/cktee_uuid.c
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: BSD-2-Clause
```
jenswi-linaro, line 1: Was this file added by mistake?

eaaltonen:
@jforissier any comments on this newly opened PR?

libteeacl/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+################################################################################
+# Built library
+################################################################################
+add_library(teeacl ${SRC})
+
+set_target_properties(teeacl PROPERTIES
+	VERSION ${PROJECT_VERSION}
+	SOVERSION ${PROJECT_VERSION_MAJOR}
+	C_STANDARD 11
+	C_STANDARD_REQUIRED ON
```
jforissier, line 25: Are you really using C11 features? I'm afraid this could cause troubles with people using older toolchains perhaps.

libteeacl/src/group.c
```diff
@@ -0,0 +1,57 @@
...
+}
+
+enum rv_groupmember teeacl_user_is_member_of(const char *user, gid_t group)
+{
+	enum rv_groupmember result = E_MEMORY;
+	int ret = 0;
+	int i = 0;
+	int grouplistsize = 8; /* initial guess */
+	gid_t *p_groups = NULL;
+	gid_t *groups = calloc(grouplistsize, sizeof(gid_t));
```
jforissier, line 68: Separate declarations from code with a blank line

libteeacl/src/group.c
```diff
@@ -0,0 +1,57 @@
...
+	if (groups == NULL)
```
jforissier, line 69: `if (!groups)`

libteeacl/src/group.c
```diff
@@ -0,0 +1,98 @@
...
+	buffer = calloc(1, b_size);
+	if (!buffer) {
+		*gid_out = TEEACL_NO_GROUP;
+		return -ENOMEM;
+	}
+
+	rv = getgrnam_r(group_name, &grp, buffer, b_size, &result);
+
+	free(buffer);
+	if (result == NULL) {
```
jforissier, line 41: `if (!result)`

libteeacl/src/tee_uuid.c
```diff
@@ -0,0 +1,35 @@
...
+	if (rv >= TEEACL_L_UUID)
+		return -1;
+	rv = uuid_parse(KERNEL_NAMESPACE, k_uuid);
+	if (rv < 0)
+		return rv;
+
+	gid_buf_len = strnlen(gid_buf, TEEACL_L_UUID);
+
+	uuid_generate_sha1(g_uuid, k_uuid, gid_buf, gid_buf_len);
+	uint len = 6;
```
jforissier, line 30: Move up with other declarations

jforissier (commented):
Comments for "Add libteeacl library"

Makefile
```diff
@@ -41,6 +42,14 @@ build-libckteec: build-libteec
 	@echo "Building libckteec.so"
 	@$(MAKE) --directory=libckteec --no-print-directory --no-builtin-variables
 
+build-libteeacl: check-libuuid
```
jforissier, line 45: This target already exists below

Makefile
```diff
@@ -41,6 +42,14 @@ build-libckteec: build-libteec
...
+	@echo "Building libteeacl.so"
+	@$(MAKE) --directory=libteeacl --no-print-directory --no-builtin-variables
+
+check-libuuid:
+	@echo "Finding uuid.pc"
+	pkg-config --atleast-version=2.34 uuid
```
jforissier, line 51: This check belongs in `libteeacl/Makefile`

libteeacl/Makefile
```diff
@@ -21,14 +21,16 @@ LIBTEEACL_AR_LIBRARY	:= $(LIB_NAME).a
 
 LIBTEEACL_SRC_DIR	:= src
 
-LIBTEEACL_SRCS	= group.c
+libteeacl_SRCS	= tee_uuid.c
```
jforissier, line 24: Why change to lowercase? Either is fine I guess (although uppercase seems better since used for other libraries), but please chose in the first commit.

jforissier (commented):
Comments on "Add function to encode a group login based uuid"

libteeacl/CMakeLists.txt
eaaltonen: I suppose not.

stdbool.h and `//` comments are used, so I suppose C99 would be appropriate.

I think it might make sense to declare
```
set(CMAKE_C_STANDARD C99)
set(CMAKE_C_STANDARD_REQUIRED TRUE)
```
at the top level `CMakeLists.txt`, unless you prefer these to be dropped entirely?

libteeacl/include/teeacl.h
```diff
@@ -0,0 +1,55 @@
...
+enum rv_groupmember teeacl_current_user_is_member_of(gid_t group);
+
+/**
+ * teeacl_user_is_member_of() - Check if `user` is a member in `group`.
+ *
+ * @param user Username string.
+ * @param group Group id to check membership of.
+ * @return enum rv_groupmember form result.
+ */
+enum rv_groupmember teeacl_user_is_member_of(const char *user, gid_t group);
```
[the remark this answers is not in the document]
eaaltonen: * `getgrouplist` requires user to be specified as `const char *`
* we already have a function to resolve `gid_t` from group name
* I find it preferable to use `gid_t` mostly in other code due to
    * type information
    * size
    * moves lookup earlier, simplifying error handling.

libteeacl/include/teeacl.h
```diff
@@ -38,6 +38,26 @@ extern "C" {
...
+ * If memory allocation fails, -ENOMEM is returned.
+ * For other failures, errno is returned and `gid_out` will be set to
+ * TEEACL_NO_GROUP.
+ *
+ * @param gid_out Ptr to gid result. After the call the value will be either
+ * - Group id or
+ * - TEEACL_NO_GROUP
+ * @param group_name Name of group to resolve.
+ * @return 0 if matching group is found, see function description for other
+ * cases.
```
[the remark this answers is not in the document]
eaaltonen: I'm trying to refer to the main part after the brief line.

libteeacl/include/teeacl.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef TEEACL_H
+#define TEEACL_H
+
+#include <grp.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define TEEACL_NO_GROUP ((gid_t)0xFFFFFFFFFFFFFFFF)
```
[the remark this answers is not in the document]
eaaltonen: > This assumes that gid_t is 64 bits, which is certainly not always the case

This assumes `gid_t` is _no more_ than 64 bits. It truncates quite well to 32 bits also, which is what happens in practice.

> Besides, you don't need this special value do you?

I suppose it could now be dropped after the `-ENOENT` return code addition.

libteeacl/Makefile
eaaltonen: Accidental change. I agree either works, but will put back to upper case for consistency.

Makefile
eaaltonen: This idea does seem attractive at first, but based on my experience so far I think the top level directory is the correct place to specify any library dependencies.

Library dependencies generally form graphs. On a good day, these are _directed graphs_. They are trees in only very specific cases.

I've often seen some common libraries, like boost or json serialization libraries, used in several places on a projects build directory tree. This has led, in practice, to the dependency _and it's version_ specified multiple times - and this is undesirable.

libteeacl/include/teeacl.h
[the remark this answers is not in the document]
eaaltonen: Doxygen uses the term "detailed description" so rephrased to use that.

Makefile
jforissier: Fair enough. Let's keep it here then.

libteeacl/include/teeacl.h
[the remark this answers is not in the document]
jforissier: I agree `gid_t` is preferable. Then the question becomes, should `uid_t` be used for the user? If you think the API is easier to use with a string, fine with me.

jforissier (commented):
With `set_target_properties()` updated:

libteeacl/include/teeacl.h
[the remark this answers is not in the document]
eaaltonen: Now I remember why I wanted to have this. It allowed to always write some defined value to `gid_out` in `teeacl_gid_from_name`.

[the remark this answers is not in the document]
jforissier: > Now I remember why I wanted to have this. It allowed to always write some defined value to `gid_out` in `teeacl_gid_from_name`.

...and that's helpful because...?

libteeacl/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+	C_EXTENSIONS ON
```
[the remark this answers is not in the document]
eaaltonen: Removed

libteeacl/include/teeacl.h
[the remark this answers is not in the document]
eaaltonen: The value might stand out so much, that even someone who did not read the function description would realize it's a special value.
But this API works too.

libteeacl/include/teeacl.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+#define TEEACL_L_UUID 48
+
+/**
+ * teeacl_gid_from_name - Try to resolve gid_t for a given `group_name`.
+ *
+ * If a matching group is found, zero is returned and `gid_out` will be set to
+ * the found value.
+ * If no group is found, -ENOENT is returned and `gid_out` will be set to
+ * TEEACL_NO_GROUP.
```
jforissier, line 46: TEEACL_NO_GROUP does not exists anymore at this point. Please do not introduce it in the first commit.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>` with @jforissier comment addressed + adding a scope prefix to the commit log header lines, something like:
`libteeacl: New ACL helper library`
`libteeacl: Add function to encode a group login based uuid`
`libteeacl: Add function to resolve name to gid`

libteeacl/include/teeacl.h
eaaltonen: * scrubbed TEEACL_NO_GROUP
* sorted src files alphabetically
* removed unnecessary curly braces
* removed PKCS#11 references from teeacl.h
* scope prefix and `acked-by` to commit messages

libteeacl/include/teeacl.h
```diff
@@ -12,11 +12,39 @@
 #define TEEACL_H
 
 #include <grp.h>
+#include <uuid/uuid.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#define KERNEL_NAMESPACE "58ac9ca0-2086-4683-a1b8-ec4bc08e01b6"
```
jenswi-linaro, line 21: It looks like this is the same value as in `tee_client_uuid_ns` at:
https://github.com/torvalds/linux/blob/521a547ced6477c54b4b0cc206000406c221b4d6/drivers/tee/tee_core.c#L26-L35

could be nice with a reference in case the connection isn't clear.

jenswi-linaro (commented):
With or without my comment addressed:

libteeacl/include/teeacl.h
eaaltonen: Added
```
+// TEE Client UUID name space identifier (UUIDv4)
+// same as `tee_client_uuid_ns` in linux kernel drivers/tee/tee_core.c
 #define KERNEL_NAMESPACE "58ac9ca0-2086-4683-a1b8-ec4bc08e01b6"
```

libteeacl/include/teeacl.h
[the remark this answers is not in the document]
eaaltonen: > Then the question becomes, should uid_t be used for the user?

To answer this, I started sketching a `enum rv_groupmember teeacl_uid_is_member_of(uid_t uid, gid_t group)`, which would first get the required username and then call `teeacl_user_is_member_of`.

As a result, I'm not keen on adding such a function. It would need to be able to report errors from both user db and group db lookups - so even more extensive and complicated return codes.

If a user needs/wants to use `uid_t`, I suggest we provide (if needed) a helper to get the username string, more or less the below code.
```
long teeacl_getpw_r_size_max(void)
{
	long s = sysconf(_SC_GETPW_R_SIZE_MAX);

	if (s == -1)
		return 1024;
	return s;
};

int teeacl_username_from_uid(char username[L_cuserid], uid_t uid) {
	struct passwd pwd = { 0 };
	char *buffer = NULL;
	size_t b_size = 0;
	struct passwd *result = NULL;
	int ret = 0; // return value from this function
	int rv = 0;

	b_size = teeacl_getpw_r_size_max();
	buffer = calloc(1, b_size);
	if (!buffer) {
		return -ENOMEM;
	}

	rv = getpwuid_r(uid, &pwd, buffer, b_size, &result);
	if (!result) {
		if (rv == 0)
			ret = -ENOENT;
		else
			ret = rv;
	} else {
		assert(strnlen(pwd.pw_name, L_cuserid) < L_cuserid);
		strncpy(username, pwd.pw_name, L_cuserid);
		ret = 0;
	}
	free(buffer);
	return ret;
}
```

@jforissier Would this PR be ready for merging? I can add the above suggestion on a separate PR if you think we should include it.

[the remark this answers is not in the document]
jforissier: @eaaltonen good for now

jforissier:
Thanks @eaaltonen. I have merged this PR manually because I have slightly edited the commit subjects and descirptions (for example s/uuid/UUID/, s/Acl/ACL/, and added the text you gave in introduction to this PR which I think was very good to understand this is supposed to be used with the PKCS#11 client library).

PR #328, opened 2022-10-24
tee-supplicant: fs: no need to use strlen() to check for empty string

jforissier:
A string is empty when its first character is zero. Therefore, we can avoid calling strlen() on each FS request.


jenswi-linaro (commented):

PR #331, opened 2022-11-08
fix `ld: cannot find -luuid` for cross building

BaochengSu:
For cross building, the cross version <CROSS_COMPILE>pkg-config should
be used instead of the native pkg-config.

This is because for the dynamic linking dependency such as
uuid-dev/libuuid, the build machine architecture version should not be
used, instead, the host machine version should be used. Or in other
words, install uuid-dev:arm64 for cross building the aarch64 target on
Debian/Ubuntu.

However, the native pkg-config could not find the correct cross
compiling dependencies if only the target arch uuid-dev is installed.
Even worse, it produces the incorrect LFLAGS without the required -L, if
the native uuid-dev is installed.

By adding multiarch, on Debian/Ubuntu, a <CROSS_COMPILE>pkg-config is
created to correctly handle such problem.

This fixes below issues while cross building.

- If only the target uuid-dev is installed:

  make: *** [Makefile:55: check-libuuid] Error 1

- If both the native and the target uuid-dev are installed:

  ld: cannot find -luuid

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ LIBDIR ?= /usr/lib
 INCLUDEDIR ?= /usr/include
 
 .PHONY: all build build-libteec build-libckteec build-libseteec \
-	build-libteeacl check-libuuid install copy_export clean cscope \
+	build-libteeacl install copy_export clean cscope \
 	clean-cscope \
 	checkpatch-pre-req checkpatch-modified-patch checkpatch-modified-file \
 	checkpatch-last-commit-patch checkpatch-last-commit-file \
@@ -46,13 +46,9 @@ build-libseteec: build-libteec
 	@echo "Building libseteec.so"
 	@$(MAKE) --directory=libseteec --no-print-directory --no-builtin-variables
 
-build-libteeacl: check-libuuid
+build-libteeacl:
 	@echo "Building libteeacl.so"
 	@$(MAKE) --directory=libteeacl --no-print-directory --no-builtin-variables
-
-check-libuuid:
-	@echo "Finding uuid.pc"
-	$(PKG_CONFIG) --atleast-version=2.34 uuid
 
 install: copy_export
 
--- a/libteeacl/Makefile
+++ b/libteeacl/Makefile
@@ -3,7 +3,7 @@ include ../config.mk
 
 OUT_DIR := $(OO)/libteeacl
 
-.PHONY: all libteeacl clean
+.PHONY: all libteeacl check-libuuid clean
 
 all: libteeacl
 install: libteeacl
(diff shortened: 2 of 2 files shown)


jenswi-linaro:
Looks good, please apply:

etienne-lms (commented):

etienne-lms (commented):
Change already addressed by https://github.com/OP-TEE/optee_client/pull/335.

BaochengSu:
> Change already addressed by #335.

Not completely, see my comments there.

etienne-lms:
@BaochengSu, can you update this P-R with required changes on top of current master?

BaochengSu:
@etienne-lms Done, plz check.

etienne-lms:
Hi @BaochengSu, sorry I missed to update the review.
Patch is ok but commit message: could you move the `Fixes:` tag next to the other tags (S-o-b, A-b). It can also be shorten using only the Git commit sha1 without github URL.
```patch
 libteeacl: Move the uuid-dev checking inside
 
-Fixes: https://github.com/OP-TEE/optee_client/commit/d448e8bb2e526ba132311a61927f0a9177ae8c4d ("Allow cross-compile pkg-config")
-
 For uuid dependency checking, cross complier version pkg-config should
 be used. However top level makefile does not hold the information of the
 cross compiler. To solve that, move the checking from top level makefile
 to the library level.

+Fixes: d448e8bb2e52 ("Allow cross-compile pkg-config")
 Signed-off-by: ...
 Acked-by: ...

**Comment by BaochengSu** [comment]
Hi @etienne-lms , refined the commit log to have the `Fixes:` tag relocated.

> It can also be shorten using only the Git commit sha1 without github URL.

Actually I am using the only the git commit-ish in the `git log`, I guessing github is just too smart.

```
commit 7a875f24a553bc4892ea2c14fae15cd1ad5642dd (HEAD -> su/fix-libuuid, fork/su/fix-libuuid)
Author: Baocheng Su <baocheng.su@siemens.com>
Date:   Mon Nov 7 20:19:29 2022 +0800

    libteeacl: Move the uuid-dev checking inside
    
    For uuid dependency checking, cross complier version pkg-config should
    be used. However top level makefile does not hold the information of the
    cross compiler. To solve that, move the checking from top level makefile
    to the library level.
    
    Fixes: d448e8bb2e52 ("Allow cross-compile pkg-config")
    Signed-off-by: Su Bao Cheng <baocheng.su@siemens.com>
    Acked-by: Etienne Carriere <etienne.carriere@linaro.org>
```

**Review by etienne-lms — COMMENTED** [review]
commit header:
```patch
-libteeacl: Move the uuid-dev checking inside
+libteeacl: Move uuid-dev checking in libteec makefile
```

s/complier/compiler/

BaochengSu:
> commit header:
> 
> ```diff
> -libteeacl: Move the uuid-dev checking inside
> +libteeacl: Move uuid-dev checking in libteec makefile
> ```
> 
> s/complier/compiler/

Done, plz check

etienne-lms (approved):
all good. thanks.

PR #335, opened 2022-11-21
Allow cross-compile pkg-config

grandpaul:
optee-client fails to cross build from source. This commit apply CROSS_COMPILE on pkg-config to let it be able to build by cross-compiler.

Signed-off-by: Helmut Grohne <helmut@subdivi.de>
Signed-off-by: Ying-Chun Liu (PaulLiu) <paul.liu@linaro.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ build-libteeacl: check-libuuid
 
 check-libuuid:
 	@echo "Finding uuid.pc"
-	pkg-config --atleast-version=2.34 uuid
+	$(PKG_CONFIG) --atleast-version=2.34 uuid
 
 install: copy_export
 
--- a/flags.mk
+++ b/flags.mk
@@ -5,6 +5,7 @@
 CROSS_COMPILE   ?= arm-linux-gnueabihf-
 CC              ?= $(CROSS_COMPILE)gcc
 AR		?= $(CROSS_COMPILE)ar
+PKG_CONFIG	?= $(CROSS_COMPILE)pkg-config
 
 override CFLAGS += -Wall -Wbad-function-cast -Wcast-align \
 		   -Werror-implicit-function-declaration -Wextra \
--- a/libteeacl/Makefile
+++ b/libteeacl/Makefile
@@ -27,10 +27,10 @@ LIBTEEACL_SRCS	+= tee_uuid.c
 LIBTEEACL_INCLUDES	= ${CURDIR}/include
 
 LIBTEEACL_CFLAGS	:= $(addprefix -I, $(LIBTEEACL_INCLUDES)) \
-			$(shell pkg-config --cflags uuid) \
+			$(shell $(PKG_CONFIG) --cflags uuid) \
 			$(CFLAGS) -D_GNU_SOURCE -fPIC
 
-LIBTEEACL_LFLAGS	:= $(LDFLAGS) $(shell pkg-config --libs uuid)
+LIBTEEACL_LFLAGS	:= $(LDFLAGS) $(shell $(PKG_CONFIG) --libs uuid)
 
 LIBTEEACL_OBJ_DIR	:= $(OUT_DIR)
 LIBTEEACL_OBJS	:= $(patsubst %.c,$(LIBTEEACL_OBJ_DIR)/%.o, $(LIBTEEACL_SRCS))


jforissier:
Thanks for the patch. Please note: the first `Signed-off-by:` has to be the same as the commit author.

This looks reasonable, but can you explain how you would typically build on e.g., a x86_64 machine? What package do you need installed in addition to the cross-compiler? Then what is the build command? The reason I am asking is, I realize the OP-TEE development environment does not use the makefiles in `optee_client` any more (it uses CMake). Therefore we are unlikely to detect breakage early. I would very much like to add a CI job to this repository to avoid this situation.

In any case:

grandpaul:
Hi Jerome,

This is actually a bug report comes from Debian.
https://bugs.debian.org/1023233

On Debian they test cross-build regularly. http://crossqa.debian.net/src/optee-client
So receiving these kind of bugs is not very rare.
I packaged optee-client to Debian just for easier testing on SystemReady IR for imx8m. 
But actually I didn't do cross-builds on Debian package of optee-client myself.

jforissier:
@grandpaul very well, would you mind adding the links to the commit description?

```
Link: https://bugs.debian.org/1023233
Link: http://crossqa.debian.net/src/optee-client
```

jforissier:
Merged, thanks! (I deleted the empty line after the Link: tags, not needed)

BaochengSu:
This commit is not completed for the corss compiling, see below errors on the latest master:

```bash
builder@2274f4f5266d:~/optee_client$ make -j
Building libteec.so
Finding uuid.pc
atleast-version=2.34 uuid
/bin/sh: 1: atleast-version=2.34: not found
make: [Makefile:55: check-libuuid] Error 127 (ignored)
```

This is due to the `flags.mk` is not included into the `Makefile`, hence the `$(PKG_CONFIG)` is empty while executing the rule `check-libuuid`.

This is addressed in 590ee8706e87d046de9798567e6fdbe1c5330bb0 of my other PR #331. Since check-uuid is only specific to libteeacl, better to move this rule to the makefile of that lib, to avoid include the `flags.mk` in the main Makefile.

BaochengSu:
@jforissier 

> This looks reasonable, but can you explain how you would typically build on e.g., a x86_64 machine? What package do you need installed in addition to the cross-compiler? Then what is the build command? The reason I am asking is, I realize the OP-TEE development environment does not use the makefiles in `optee_client` any more (it uses CMake). Therefore we are unlikely to detect breakage early. I would very much like to add a CI job to this repository to avoid this situation.

I've built a docker image for cross building the aarch64 optee-client on amd64. See below:

```Dockerfile
# Dockerfile for cross building aarch64
# 
# Use below command to build:
#    docker build -t optee-client-builder:cross-aarch64 -f ./docker/Dockerfile.cross-aarch64 .
#
# Use below command to use the image:
#    docker run -it --rm -u $(id -u):$(id -g) -v ${PWD}:/home/builder/optee_client optee-client-builder:cross-aarch64
# Then inside the container:
#    make -j
#
FROM debian:bullseye-slim

# Add arm64 arch
RUN dpkg --add-architecture arm64

# Config OS
RUN apt-get update && \
    apt-get install -y locales ca-certificates && \
    localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
ENV LANG=en_US.utf8

# Install dep tools
ENV DEP_PACKAGES='crossbuild-essential-arm64 uuid-dev:arm64 pkg-config'
RUN apt-get update && \
    apt-get install --no-install-recommends -y ${DEP_PACKAGES} && \
    apt-get clean

ENV CROSS_COMPILE=aarch64-linux-gnu-

# Add new user
RUN useradd --create-home --shell /bin/bash builder
USER builder
WORKDIR /home/builder/optee_client
```

Or check my forked branch https://github.com/BaochengSu/optee_client/tree/su/cross-build-docker

jforissier:
@BaochengSu thank you, that's helpful. I will create a PR to introduce CI based on this Docker configuration.

jforissier:
@BaochengSu any idea how it could work on Ubuntu instead of Debian? I can't figure out which packages are needed for `aarch64-linux-gnu-pkg-config uuid-dev`.

jforissier:
See https://github.com/OP-TEE/optee_client/pull/338. We need to figure out what to do in the code and in the CI so that all 8 builds pass. Thanks.

eaaltonen:
@jforissier: I was interested in this as well. I configured the `arm64` ports repositories and installed `libuuid1:arm64 uuid-dev:arm64`.
This seemed to compile, although I saw a
```
Finding uuid.pc
atleast-version=2.34 uuid
/bin/sh: 1: atleast-version=2.34: not found
make: [Makefile:55: check-libuuid] Error 127 (ignored)
```
so top Makefile should also `include ./flags.mk`.

Issue #348, opened 2023-04-20
Type definitions in tee_client_api.h violate the GlobalPlatform TEE Client API spec


tchebb:
### Background

The [GlobalPlatform TEE Client API Specification v1.0][spec], which OP-TEE's libteec adheres to, defines API objects as structures visible to and allocated by the user rather than as opaque pointers.

[spec]: https://higherlogicdownload.s3.amazonaws.com/GLOBALPLATFORM/transferred-from-WS5/TEE_Client_API_Specification-V1.0_c.pdf

While some of those objects—like `TEEC_UUID`—are fully specified by the spec, others—specifically `TEEC_Context`, `TEEC_Session`, `TEEC_SharedMemory`, and `TEEC_Operation`—end with an `imp` field of a type defined by the implementation. For example, `TEEC_SharedMemory` is specified as follows:
```c
typedef struct
{
	void* buffer;
	size_t size;
	uint32_t flags;
	<Implementation-Defined Type> imp;
} TEEC_SharedMemory;
```

And this is what the spec has to say about the `imp` field (emphasis mine):
> Implementations are allowed to extend some of the data structures defined in this specification to include a single field of *implementation-defined* type, named `imp`. **Implementations MUST NOT add new fields outside of `imp`**.

### The problem

But here's how we currently define our version of `TEEC_SharedMemory`:
https://github.com/OP-TEE/optee_client/blob/8533e0e6329840ee96cf81b6453f257204227e6c/public/tee_client_api.h#L291-L306

We don't have a field named `imp` at all! We instead put multiple implementation-specific fields right in the struct, **in direct violation of the spec.** I imagine we did this because it seemed an immaterial distinction, but I've found a case where it makes a difference for me as a user of libteec:

### Why it matters

I'm currently working on making Apache's [Teaclave TrustZone SDK][teaclave-tz], a Rust abstraction over libteec, work with more implementations than just OP-TEE. Part of that work is making its [FFI structure definitions][rust-defs], which currently hardcode OP-TEE's implementation-specific fields, generic over multiple implementations. The obvious way to do that is to use Rust generics, adding a type parameter that defines the concrete type of the `imp` field. But that requires me to split the OP-TEE implementation-specific fields into their own structure, resulting in this concrete type:

```c
typedef struct
{
	void* buffer;
	size_t size;
	uint32_t flags;
	struct {
	 	int id;
 	 	size_t alloced_size;
  		void *shadow_buffer;
 	 	int registered_fd;
[...]

PR #349, opened 2023-07-14
libteec: Move OP-TEE defined fields into an imp struct

JLLinaro:
GlobalPlatform TEE Client API Specification v1.0 specifies that the structs TEEC_Context, TEEC_Session, TEEC_SharedMemory, and TEEC_Operation shall have a user defined struct named imp. In OP-TEE the struct is not there and instead the user defined fields are declared directly in the top structs.
This commit introduces the imp struct to better support using different implementations. The imp fields now represent the implementation defined parts of the structs that was previously declared directly in the top struct. All previously available parameters are preserved in the imp struct.

Link: https://github.com/OP-TEE/optee_client/issues/348
Reported-by: Tom Hebb <tommyhebb@gmail.com>


JLLinaro:
The automatic tests fail but that has been solved in the following pull request https://github.com/OP-TEE/optee_test/pull/684

jforissier:
This LGTM, but as mentioned in https://github.com/OP-TEE/optee_client/issues/348, it changes the layout of `TEEC_SharedMemory` and therefore is an ABI breakage. So the `libteec` version should be bumped from 1.0.0 to 2.0.0.

JLLinaro:
@jenswi-linaro version is now updated to 2.0.0

jforissier:
We need `MAJOR_VERSION := 2` in `libteec/Makefile`, too.

public/tee_client_api.h
```diff
@@ -293,16 +295,18 @@ typedef struct {
...
+	struct {
+		int id;
+		size_t alloced_size;
+		void *shadow_buffer;
+		int registered_fd;
+		union {
+			bool dummy;
+			uint8_t flags;
+		} internal;
+    } imp;
```
etienne-lms, line 309: fix indentation

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commit 
"libteec: Move OP-TEE defined fields into an imp struct" 
(with indentation issue fixed)

jforissier:
Please squash the second commit into the first one (and keep subject "libteec: Move OP-TEE defined fields into an imp struct").
With that:

jforissier:
There is a conflict, would you please rebase onto master, and also add the Acked-by: tags from Etienne and me?

libteec/include/tee_client_api.h
```diff
@@ -384,8 +388,10 @@ typedef union {
  */
 typedef struct {
 	/* Implementation defined */
-	TEEC_Context *ctx;
-	uint32_t session_id;
+	struct{
```
jenswi-linaro, line 391: Please add a space (` `), between `struct` and `{`.

jforissier:
Ping @JLLinaro

libteec/include/tee_client_api.h
```diff
@@ -293,16 +295,18 @@ typedef struct {
...
-	} internal;
+	struct {
+		int id;
+		size_t alloced_size;
+		void *shadow_buffer;
+		int registered_fd;
+		union {
+			bool dummy;
+			uint8_t flags;
+		} internal;
```
vincent-mailhol, line 308: Is the `dummy` field still needed following this change? `dummy` was introduced in 4f3d4cbb7824bab8f5c0a445f301528a53142f6b not to break the ABI. Here we are breaking it anyway, so while at it, I suggest some clean up:
```suggestion
		uint8_t flags;
```
*N.B.:* If there is a wish to keep more flag bits for future use, then just declare `flags` as `uint32_t`.

jforissier: I agree. `uint32_t flags;` LGTM.

jforissier:
@JLLinaro any chance this could make it into the next release? (-rc1 is in 3 days).

On the other hand, I am now wondering if this is a good idea. Who cares about this? Sure it is in the spec, but what difference does it make? Apps are not supposed to manipulate the `imp` field anyways. And changing it now will have impacts on downstream users (version change).

jenswi-linaro:
The answer to your question is in https://github.com/OP-TEE/optee_client/issues/348

jforissier:
> The answer to your question is in #348

Ah, yes. It would be nice to mention that the `imp` struct makes it easier to create a biding for Rust.

jenswi-linaro:
I think we need to explain to @JLLinaro that it's going to cost him a beer each time github reminds us that it's not done yet. :-)

libteec/include/tee_client_api.h
jforissier: @JLLinaro comment not addressed

libteec/include/tee_client_api.h
jforissier: @JLLinaro comment not addressed

libteec/include/tee_client_api.h
```diff
@@ -293,16 +295,17 @@ typedef struct {
...
-		uint8_t flags;
-	} internal;
+	struct {
+		int id;
+		size_t alloced_size;
+		void *shadow_buffer;
+		int registered_fd;
+		union {
+			uint32_t flags;
+		} internal;
```
jforissier, line 307: The union should not be needed.

jforissier:
Merged manually (with commit message slightly edited).

PR #350, opened 2023-08-10
CMake build structure, target and style fixes

eaaltonen:
Resolve some inconsistencies originating from commit e0a12e23 "cmake: Initial CMake support".

I think these would be good to fix both to:
* improve the consistency
* prepare for adding a CMake config-file package to the install

Also it would be good to do a look ahead check for the used target names. If/when a config-file package is added, it will expose targets to the developers, typically prefixed.

The prefix would likely be one of:
* `optee`
* `Optee`
* `OPTEE`

according to the project's preference, and the targets will then look something like:

* `OPTEE::optee-client-headers`
* `OPTEE::teec`
* `OPTEE::ckteec`
* `OPTEE::teeacl`


eaaltonen:
@jforissier, @etienne-lms, @jenswi-linaro: any comments?

jforissier (commented):
- For "cmake: remove whitespace before parenthesis":
  (minor: it would be best to have this commit first in the series, since you are adding new calls in other commits without the space)
- For the other commits: not being a CMake expert I can at least say they seem reasonable to me so:

eaaltonen:
> For "cmake: remove whitespace before parenthesis" it would be best to have this commit first in the series

done.

> Acked-by

Added. However, I noticed that `libteec.pc` wasn't actually installed, so I added two extra commits for that.

eaaltonen:
The main thing I'd like your opinion on is if you're happy with the `optee-client-headers` name. Once a devel package is added, a target rename is an extra inconvenience for the downstream users.

jforissier:
For "pkgconfig file: remove extra comments"

libteec/CMakeLists.txt
```diff
@@ -76,3 +76,7 @@ target_link_libraries(teec
 # FIXME: Should we change this to /usr/local/lib?
 install(TARGETS teec LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
                       ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}")
+
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/libteec.pc"
+	DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig"
+)
```
jforissier, line 82: Close parenthesis on the previous line for consistency?

jforissier:
> The main thing I'd like your opinion on is if you're happy with the `optee-client-headers` name. Once a devel package is added, a target rename is an extra inconvenience for the downstream users.

Actually it is unclear to me if that's a good thing. I am not very familiar with CMake. What's the common name scheme? We don't want to do something specific unless we have to. Why is optee-client-headers needed at all? Why isn't OPTEE::teec sufficient? Other libraries such as ckteec don't have an equivalent -headers package.

Also why do we keep this `public/` directory? IMO the header files in `public/` should be in `libteec/include`, while the current content of `libteec/include` is private and should therefore be moved directly under `libteec/`.

eaaltonen:
>  For "pkgconfig file: remove extra comments"
> Acked-by: Jerome Forissier <jerome.forissier@linaro.org>

added, thanks.

libteec/CMakeLists.txt
eaaltonen: I picked consistency with the examples near the beginning of the file.

As for the choice between the conventions:

* closing parenthesis on a separate line reduces diff noise a bit
* Many IDEs, such as CLion tend to "help" you and place the paranthesis where you suggested - so this might also be a good convention just because the tooling tends that way.

jforissier:
> A PKCS11 library with ckteec extensions for group access should perhaps support dynamic loading (_if that's feasible_), for that case the `optee-client-headers` would be required in order to use the same constants **without** linking directly to `ckteec`.

You mean without linking directly to `teec`? Why would it be a problem (it is the base library that will be needed anyways)? Would [this](https://stackoverflow.com/questions/61270645/how-to-use-headers-from-library-but-do-not-link-to-it-with-cmake) help? (the first answer -- `FOO_INCLUDE_DIR`)

jforissier:
> > You mean without linking directly to teec?
> 
> No. `ckteec` can and should link directly to `teec`, that's fine. `ckteec` itself provides the PKCS#11 functions and with OpenSSL and NSS, the PKCS#11 modules are normally loaded dynamically. It would also be helpful in that the user code could more easily use another PKCS#11 provider on x86_64 architectures and OP-TEE on ARM achitectures.

Ah OK, so optee-client-headers would have the PKCS#11 defines?

eaaltonen:
> Ah OK, so optee-client-headers would have the PKCS#11 defines?

Maybe, or possibly even more fine grained. Dynamic loading code would anyway need to include the PKCS#11 headers - and group login would need some extras, like `TEEC_LOGIN_GROUP`.

libteec/CMakeLists.txt
eaaltonen: If you pick one convention, I can offer to reformat all of the `CMakeLists.txt` files accordingly.

eaaltonen:
@jforissier : if I move the headers under `teec` for now, would that be something you'd be happy to merge?

jforissier:
> @jforissier : if I move the headers under `teec` for now, would that be something you'd be happy to merge?

I think so.

eaaltonen:
> For "pkgconfig file: remove extra comments"
> 

Removed previous `Acked-by` due to heavy editing.

jforissier:
> Done. Can you re-open this PR, or do I need to open a new one?

Please open a new one. I cannot re-open this one; when I hover the mouve on the "Reopen and comment" button it says "The cmake-structure branch was force-pushed or recreated.".

PR #352, opened 2023-08-24
libptateec: manufacturing protection PTA

ldts:
This abstraction provides TEEC access to stable PTAs present in the OP-TEE upstream tree.

The first of these miscellaenous PTAs to be integrated in the library is the iMX Manufacturing Protection [1] for which two functions are provided:
  - Retrieval of the EC Public Key.
  - Signature generation.

In order to access the service provided by the manufacturing protection PTA, OP-TEE needs to be built with this option https://github.com/OP-TEE/optee_os/pull/6274 

[1] AN13676, i.MX RT1170 Manufacturing Protection


ldts:
The build error seems unrelated to the change.

ldts:
any comments?

libptateec/include/pta_tee.h
```diff
@@ -0,0 +1,62 @@
...
+#define PTA_TEE_H
+
+#include <unistd.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef unsigned long PTA_ULONG;
+typedef PTA_ULONG PTA_RV;
```
jenswi-linaro, line 16: What's wrong with TEEC_Result?

ldts: We dont know and can't foresee what the PTAs will choose to report..errors could have different meanings depending the calls I guess. having another indirection adds flexibility.

jenswi-linaro: Since we're only supporting upstream PTAs here we can encourage them to use TEEC_Result error codes, and if needed pick values in the `0x80000000 – 0x8FFFFFFF` range for an occasional extension.

ldts: so shall I get rid of the typedef and just cast in case of needing to extend ?

jenswi-linaro: Yes, please.

ldts:
actioned.

libptateec/src/pta_imx_manufacturing_protection.c
```diff
@@ -0,0 +1,94 @@
...
+
+TEEC_Result pta_imx_mprotect_get_key(char *key, size_t *len)
+{
+	TEEC_Result res = TEEC_SUCCESS;
+	TEEC_Operation op = { 0 };
+
+	if (!key || !len || !*len)
+		return TEEC_ERROR_BAD_PARAMETERS;
+
+	if (!pta_open_session(&manufacturing_protection_ta_ctx))
```
jenswi-linaro, line 26: How is the session closed?

ldts: um I didnt think it matter to keep an open session for all calls. the way I thought the user (a single process) would access the call is by periodically signing randome stuff and maybe even getting the key from time to time. maybe every few minutes.

jenswi-linaro: Leaving a session open wastes memory so there should be a way to close the session without killing the process.

ldts: sure, ok let me fix that.

libptateec/include/pta_tee.h
```diff
@@ -0,0 +1,69 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * pta_imx_mprotect_get_key() - Retrieves the iMX CAAM Manufacturing Protection
+ * EC public key. The components x,y are retrieved in RAW format and should
+ * be converted to DER or PEM as required.
+ *
```
etienne-lms, line 20: could you describe the arguments?

libptateec/src/pta.h
```diff
@@ -0,0 +1,34 @@
...
+#include <pta_tee.h>
+#include <signal.h>
+#include <stdatomic.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <tee_client_api.h>
+#include <teec_trace.h>
+
+struct ta_context {
```
etienne-lms, line 20: Could you also add a description for struct and API functions?

libptateec/src/pta_imx_manufacturing_protection.c
```diff
@@ -0,0 +1,110 @@
...
+	op.paramTypes = TEEC_PARAM_TYPES(TEEC_MEMREF_TEMP_INPUT,
+					 TEEC_MEMREF_TEMP_OUTPUT,
+					 TEEC_MEMREF_TEMP_OUTPUT, TEEC_NONE);
+
+	op.params[0].tmpref.buffer = msg;
+	op.params[0].tmpref.size = msg_len;
+	op.params[1].tmpref.buffer = sig;
+	op.params[1].tmpref.size = *sig_len;
+	op.params[2].tmpref.buffer = mpmr;
+	op.params[2].tmpref.size =  *mpmr_len;
```
etienne-lms, line 82: 2 space chars

libptateec/src/pta_imx_manufacturing_protection.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+#define PTA_IMX_MP_CMD_SIGNATURE_MPMR	0
+
+/*
+ * Get the manufacturing protection public key
+ *
+ * [out]	memref[0].buffer	Public key buffer
+ * [out]	memref[0].size		Public key size
+ *
+ * Return codes:
```
etienne-lms, line 29: I suggest you remove this list. The command can return more codes.

libptateec/src/pta.h
```diff
@@ -0,0 +1,34 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2023, Foundries.io Ltd
+ */
+#ifndef PTA_H
+#define PTA_H
```
etienne-lms, line 6: PTA_H seems too short to me for a header file exposed to other numerous libs.
What about PTATEEC_PTA_H

etienne-lms (commented):
LGTM aside the few comments.

libptateec/src/pta_imx_manufacturing_protection.h
ldts: no I dont think it can. I think we should keep this.

libptateec/src/pta.h
```diff
@@ -0,0 +1,60 @@
...
+ *	pta_open_session();
+ *	[...]
+ *	pta_invoke_cmd();
+ *	[...]
+ * }
+ */
+
+/**
+ * pta_open_session() - Opens a session with the PTA uuid in the ta_context.
+ * If the session is already open it will increment a session counter.
```
etienne-lms, line 45:
```
 * @ctx Session context, already opened ornot
 * Field @ctx->uuid defines the target PTA

File: libptateec/src/pta.h [hunk 9]
```diff
@@ -0,0 +1,60 @@
...
+ */
+TEEC_Result pta_open_session(struct ta_context *ctx);
+
+/**
+ * pta_invoke_cmd() - Invokes a command in the PTA
+ */
+TEEC_Result pta_invoke_cmd(struct ta_context *ctx, uint32_t cmd_id,
```
etienne-lms, line 52:
```
 * @ctx Opened session context
 * @cmd_id Command passed to target PTA
 * @operation TEE operation arguments passed to target PTA
 * @error_origin: Output TEE_ORIGIN_* emitter of the result code

Comment (ldts) [reply libptateec/src/pta.h]: thanks

File: libptateec/src/pta_imx_manufacturing_protection.h (hunk 6 above)
Comment (etienne-lms) [reply libptateec/src/pta_imx_manufacturing_protection.h:29]: It looks the PTA can also return `TEE_ERROR_SHORT_BUFFER`, `TEE_ERROR_OUT_OF_MEMORY`, `TEE_ERROR_NOT_SUPPORTED` or `TEE_ERROR_GENERIC`.

**Comment by ldts** [comment]
@etienne-lms @jenswi-linaro anything else?

File: libptateec/src/pta.c [hunk 10]
```diff
@@ -0,0 +1,73 @@
...
+	pthread_mutex_lock(&ctx->lock);
+	if (!ctx->open) {
+		atomic_store(&ctx->count, 0);
+		pthread_mutex_unlock(&ctx->lock);
+
+		return TEEC_ERROR_COMMUNICATION;
+	}
+	pthread_mutex_unlock(&ctx->lock);
+	ret = TEEC_InvokeCommand(&ctx->session, cmd_id, op, error_origin);
+	atomic_fetch_sub(&ctx->count, 1);
```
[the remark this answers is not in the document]
etienne-lms: Sorry, I think it's a bit weird that a generic PTA invocation API requires each call to `pta_open_session()` be always followed by a single call to `pta_invoke_cmd()`.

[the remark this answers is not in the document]
jenswi-linaro: Yeah, this session handling is a bit more complicated than needed with mutex and reference counter. Couldn't we skip it all and rely on the normal TEE Client primitives?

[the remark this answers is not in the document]
ldts: IMO being capable of rejecting a close request coming from some separate thread because an invoke is about to follow is a nice feature to have (it really only costs the mutex and reference counter and it has no impact on performance).

But sure I can remove it if you think it adds too much maintenance noise.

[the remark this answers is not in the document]
jenswi-linaro: Please remove it then.

jenswi-linaro:
I guess this wasn't needed after all.

PR #354, opened 2023-08-25
tee-supplicant: Allow to control RPMB emulation during runtime

### LINKED ISSUE #353: Make RPMB_EMU a runtime switch
Would it be possible to turn`RPMB_EMU` into a runtime switch for tee-supplicant? We are seeing even distro packages (https://packages.debian.org/bookworm/tee-supplicant) being built with the default `RPMB_EMU=1` which makes them useless for real use. Did anyone look into that already?

jan-kiszka:
This permits to build with RPMB_EMU=1 without breaking real usage. Specifically helpful for distros. Closes #353.

Still testing, thus "draft".

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -26,6 +26,7 @@
  */
 
 #include <dirent.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <linux/types.h>
 #include <linux/mmc/ioctl.h>
@@ -46,8 +47,6 @@
 #ifdef RPMB_EMU
 #include <stdarg.h>
 #include "hmac_sha2.h"
-#else
-#include <errno.h>
 #endif
 
 /*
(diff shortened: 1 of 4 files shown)


tee-supplicant/src/rpmb.c
```diff
@@ -890,7 +887,13 @@ static uint32_t rpmb_data_req(int fd, struct rpmb_data_frame *req_frm,
 		goto out;
 	}
 
-	st = IOCTL(fd, MMC_IOC_MULTI_CMD, mcmd);
+	if (use_rpmb_emu()) {
```
jforissier, line 890: Not sure it is very helpful to modify all the call sites to test for emu enabled. In this particular case we could define a `ioctl` wrapper for example `do_ioctl()` which would call either `ioctl()` or `ioctl_emu()`.

jforissier (commented):
LGTM overall.

The comment in the Makefile should preferably be adjusted because the meaning of RPMB_EMU changes slightlty (from "use RPMB emulation" to "support RPMB emulation and use it when `--rpmb-emu` is given). The wording in `CMakeLists.txt` is fine though.

```diff
-# Emulate RPMB ioctl's by default
+# Support RPMB emulation by default (--rpmb-emu command line argument)
 RPMB_EMU ?= 1
```

tee-supplicant/src/rpmb.c
jan-kiszka: We don't have multiple call sites for those, that's why I left the if-else there and didn't add another level of wrappers.

jforissier: Ah yes, I see what you mean. Please disregard my comment then.

jenswi-linaro:
Please add something about the new --rpmb-emu command line option in the commit message.

jan-kiszka:
Continued by https://github.com/OP-TEE/optee_client/pull/355

PR #355, opened 2023-08-26
tee-supplicant: Enable command line support for RPMB_EMU

lsun100:
The current eMMC RPMB support is enabled with RPMB_EMU=1 by default at compiling time and needs re-compiling to access real eMMC RPMB. It's inconvenient especially when this package is provided by known Linux distributions like Ubuntu. This commit enhances it by adding command line option for RPMB_EMU. It still uses RPMB emulation by default to be backwoard compatible, but could use '--rpmb-emu-disable' to access the actual eMMC RPMB.

### DIFF
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -4,7 +4,6 @@ project (tee-supplicant C)
 # Configuration flags always included
 ################################################################################
 option (CFG_TA_TEST_PATH "Enable tee-supplicant to load from test/debug path" OFF)
-option (RPMB_EMU "Enable tee-supplicant to emulate RPMB" ON)
 option (CFG_TA_GPROF_SUPPORT "Enable tee-supplicant support for TAs instrumented with gprof" ON)
 option (CFG_FTRACE_SUPPORT "Enable tee-supplicant support for TAs instrumented with ftrace" ON)
 option (CFG_TEE_SUPP_PLUGINS "Enable tee-supplicant plugin support" ON)
@@ -72,11 +71,6 @@ endif()
 if (CFG_TA_TEST_PATH)
 	target_compile_definitions (${PROJECT_NAME}
 		PRIVATE -DCFG_TA_TEST_PATH=${CFG_TA_TEST_PATH})
-endif()
-
-if (RPMB_EMU)
-	target_compile_definitions (${PROJECT_NAME}
-		PRIVATE -DRPMB_EMU=1)
 endif()
 
 if (CFG_TA_GPROF_SUPPORT)
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -26,12 +26,14 @@
  */
 
 #include <dirent.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <linux/types.h>
 #include <linux/mmc/ioctl.h>
 #include <netinet/in.h>
 #include <pthread.h>
 #include <rpmb.h>
+#include <stdarg.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
@@ -43,12 +45,7 @@
 #include <tee_supplicant.h>
 #include <unistd.h>
 
-#ifdef RPMB_EMU
-#include <stdarg.h>
 #include "hmac_sha2.h"
-#else
-#include <errno.h>
-#endif
 
 /*
  * Request and response definitions must be in sync with the secure side
@@ -139,17 +136,16 @@ static pthread_mutex_t rpmb_mutex = PTHREAD_MUTEX_INITIALIZER;
 /* Maximum number of commands used in a multiple ioc command request */
 #define RPMB_MAX_IOC_MULTI_CMDS		3
 
-#ifndef RPMB_EMU
+typedef struct {
+	int (*ioctl)(int d, long unsigned int request, ...);
+	int (*get_fd)(uint16_t dev_id);
+	TEEC_Result (*read_cid)(uint16_t dev_id, uint8_t *cid);
+	TEEC_Result (*read_size_mult)(uint16_t dev_id, uint8_t *value);
+	TEEC_Result (*read_rel_wr_sec_c)(uint16_t dev_id, uint8_t *value);
+	bool (*remap_dev_id)(uint16_t dev_id, uint16_t *ndev_id);
+} rpmb_ops_t;
 
-#define IOCTL(fd, request, ...)					   \
-	({							   \
-		int ret;					   \
-		ret = ioctl((fd), (request), ##__VA_ARGS__);	   \
-		if (ret < 0)					   \
-			EMSG("ioctl ret=%d errno=%d", ret, errno); \
-		ret;						   \
-	})
-
+static rpmb_ops_t rpmb_ops;
 
 /* Open and/or return file descriptor to RPMB partition of device dev_id */
 static int mmc_rpmb_fd(uint16_t dev_id)
@@ -389,10 +385,6 @@ out:
 	return found;
 }
 
-#else /* RPMB_EMU */
-
-#define IOCTL(fd, request, ...) ioctl_emu((fd), (request), ##__VA_ARGS__)
-
 /* Emulated rel_wr_sec_c value (reliable write size, *256 bytes) */
 #define EMU_RPMB_REL_WR_SEC_C	1
 /* Emulated rpmb_size_mult value (RPMB size, *128 kB) */
@@ -414,9 +406,7 @@ struct rpmb_emu {
 		uint16_t address;
 	} last_op;
 };
-static struct rpmb_emu rpmb_emu = {
-	.size = EMU_RPMB_SIZE_BYTES
-};
+static struct rpmb_emu *rpmb_emu;
 
 static struct rpmb_emu *mem_for_fd(int fd)
 {
@@ -429,7 +419,7 @@ static struct rpmb_emu *mem_for_fd(int fd)
 		return NULL;
 	}
 
-	return &rpmb_emu;
+	return rpmb_emu;
 }
 
 #if (DEBUGLEVEL >= TRACE_FLOW)
@@ -609,7 +599,7 @@ static void ioctl_emu_read_ctr(struct rpmb_emu *mem,
 	frm->op_result = compute_hmac(mem, frm, 1);
 }
 
-static uint32_t read_cid(uint16_t dev_id, uint8_t *cid)
+static uint32_t read_cid_emu(uint16_t dev_id, uint8_t *cid)
 {
 	/* Taken from an actual eMMC chip */
 	static const uint8_t test_cid[] = {
@@ -747,7 +737,7 @@ static int ioctl_emu(int fd, unsigned long request, ...)
 	return res;
 }
 
-static int mmc_rpmb_fd(uint16_t dev_id)
+static int mmc_rpmb_fd_emu(uint16_t dev_id)
 {
 	(void)dev_id;
 
@@ -755,7 +745,7 @@ static int mmc_rpmb_fd(uint16_t dev_id)
 	return 0;
 }
 
-static TEEC_Result read_size_mult(uint16_t dev_id, uint8_t *value)
+static TEEC_Result read_size_mult_emu(uint16_t dev_id, uint8_t *value)
 {
 	(void)dev_id;
 
@@ -763,7 +753,7 @@ static TEEC_Result read_size_mult(uint16_t dev_id, uint8_t *value)
 	return TEEC_SUCCESS;
 }
 
-static TEEC_Result read_rel_wr_sec_c(uint16_t dev_id, uint8_t *value)
+static TEEC_Result read_rel_wr_sec_c_emu(uint16_t dev_id, uint8_t *value)
 {
 	(void)dev_id;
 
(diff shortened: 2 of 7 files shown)


jenswi-linaro:
This relates to https://github.com/OP-TEE/optee_client/pull/354

lsun100:
> This relates to #354

I see.  Looks like #354 still has conditional compilation. I talked with one Linux distro vendor earlier. They don't want to recompile packages with different enabling options in LTS release. However, it's possible to use a patch to support both functionalities at runtime and keep the default behavior consistent with old versions. Thus this patch is created.

jforissier:
> 

How about we just disable `RPMB_EMU` by default?  Would the distros be willing to pick such a patch for their LTS releases? If so that would make things much simpler. What I propose is:
```diff
diff --git a/tee-supplicant/CMakeLists.txt b/tee-supplicant/CMakeLists.txt
index 57a3326..bb181d5 100644
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -4,7 +4,7 @@ project (tee-supplicant C)
 # Configuration flags always included
 ################################################################################
 option (CFG_TA_TEST_PATH "Enable tee-supplicant to load from test/debug path" OFF)
-option (RPMB_EMU "Enable tee-supplicant to emulate RPMB" ON)
+option (RPMB_EMU "Enable tee-supplicant to emulate RPMB" OFF)
 option (CFG_TA_GPROF_SUPPORT "Enable tee-supplicant support for TAs instrumented with gprof" ON)
 option (CFG_FTRACE_SUPPORT "Enable tee-supplicant support for TAs instrumented with ftrace" ON)
 option (CFG_TEE_SUPP_PLUGINS "Enable tee-supplicant plugin support" ON)
diff --git a/tee-supplicant/Makefile b/tee-supplicant/Makefile
index b862a7e..c193a2f 100644
--- a/tee-supplicant/Makefile
+++ b/tee-supplicant/Makefile
@@ -3,8 +3,8 @@ include ../config.mk
 
 OUT_DIR := $(OO)/tee-supplicant
 
-# Emulate RPMB ioctl's by default
-RPMB_EMU	?= 1
+# Set to 1 to emulate RPMB ioctl's
+RPMB_EMU	?= 0
 
 .PHONY: all tee-supplicant clean
 
diff --git a/tee-supplicant/tee_supplicant_android.mk b/tee-supplicant/tee_supplicant_android.mk
index 49bc388..e3a3bd8 100644
--- a/tee-supplicant/tee_supplicant_android.mk
+++ b/tee-supplicant/tee_supplicant_android.mk
@@ -30,7 +30,7 @@ LOCAL_SRC_FILES += src/tee_socket.c
 LOCAL_CFLAGS += -DCFG_GP_SOCKETS=1
 endif
 
-RPMB_EMU ?= 1
+RPMB_EMU ?= 0
 ifeq ($(RPMB_EMU),1)
 LOCAL_SRC_FILES += src/sha2.c src/hmac_sha2.c
 LOCAL_CFLAGS += -DRPMB_EMU=1
```

jforissier:
@jan-kiszka

tee-supplicant/src/tee_supplicant.c
```diff
@@ -493,6 +493,7 @@ static int usage(int status)
 	fprintf(stderr, "\t-h, --help: this help\n");
 	fprintf(stderr, "\t-d, --daemonize: run as a daemon (fork and return "
 			"after child has opened the TEE device or on error)\n");
+	fprintf(stderr, "\t-e, --rpmb-emu-disable: RPMB emunation disable\n");
```
jan-kiszka, line 496: This seems inverted: Emulation is the corner case, not the common one. So we likely rather want -e having the semantic of --rpmb-emulation.

tee-supplicant/src/rpmb.c
```diff
@@ -987,3 +977,23 @@ uint32_t rpmb_process_request(void *req, size_t req_size, void *rsp,
...
+		rpmb_ops.read_rel_wr_sec_c = read_rel_wr_sec_c_emu;
+		rpmb_ops.remap_dev_id = remap_rpmb_dev_id_emu;
+	} else {
+		rpmb_ops.ioctl = ioctl;
+		rpmb_ops.get_fd = mmc_rpmb_fd;
+		rpmb_ops.read_cid = read_cid;
+		rpmb_ops.read_size_mult = read_size_mult;
+		rpmb_ops.read_rel_wr_sec_c = read_rel_wr_sec_c;
+		rpmb_ops.remap_dev_id = remap_rpmb_dev_id;
+	}
```
jan-kiszka, line 998: IIRC, but I also didn't address this in my version, emulation comes with a rather large static buffer. Maybe that should now be allocated dynamically and only if emulation is actually used.

jan-kiszka:
I'm happy to close #353 if this one is supposed to take over. I'm still not done with my own testing as I ran into side issues and then out of time.

lsun100:
> > 
> 
> How about we just disable `RPMB_EMU` by default? Would the distros be willing to pick such a patch for their LTS releases? If so that would make things much simpler. What I propose is:
> 
> ```diff
> diff --git a/tee-supplicant/CMakeLists.txt b/tee-supplicant/CMakeLists.txt
> index 57a3326..bb181d5 100644
> --- a/tee-supplicant/CMakeLists.txt
> +++ b/tee-supplicant/CMakeLists.txt
> @@ -4,7 +4,7 @@ project (tee-supplicant C)
>  # Configuration flags always included
>  ################################################################################
>  option (CFG_TA_TEST_PATH "Enable tee-supplicant to load from test/debug path" OFF)
> -option (RPMB_EMU "Enable tee-supplicant to emulate RPMB" ON)
> +option (RPMB_EMU "Enable tee-supplicant to emulate RPMB" OFF)
>  option (CFG_TA_GPROF_SUPPORT "Enable tee-supplicant support for TAs instrumented with gprof" ON)
>  option (CFG_FTRACE_SUPPORT "Enable tee-supplicant support for TAs instrumented with ftrace" ON)
>  option (CFG_TEE_SUPP_PLUGINS "Enable tee-supplicant plugin support" ON)
> diff --git a/tee-supplicant/Makefile b/tee-supplicant/Makefile
> index b862a7e..c193a2f 100644
> --- a/tee-supplicant/Makefile
> +++ b/tee-supplicant/Makefile
> @@ -3,8 +3,8 @@ include ../config.mk
>  
>  OUT_DIR := $(OO)/tee-supplicant
>  
> -# Emulate RPMB ioctl's by default
> -RPMB_EMU	?= 1
> +# Set to 1 to emulate RPMB ioctl's
> +RPMB_EMU	?= 0
>  
>  .PHONY: all tee-supplicant clean
>  
> diff --git a/tee-supplicant/tee_supplicant_android.mk b/tee-supplicant/tee_supplicant_android.mk
> index 49bc388..e3a3bd8 100644
> --- a/tee-supplicant/tee_supplicant_android.mk
> +++ b/tee-supplicant/tee_supplicant_android.mk
> @@ -30,7 +30,7 @@ LOCAL_SRC_FILES += src/tee_socket.c
>  LOCAL_CFLAGS += -DCFG_GP_SOCKETS=1
>  endif
>  
> -RPMB_EMU ?= 1
> +RPMB_EMU ?= 0
>  ifeq ($(RPMB_EMU),1)
>  LOCAL_SRC_FILES += src/sha2.c src/hmac_sha2.c
>  LOCAL_CFLAGS += -DRPMB_EMU=1
> ```
Yes, this would be much easier if possible. But unfortunately the distros we're working with doesn't like this idea. Please see the feedback below that from one of the distros. They prefer to pick up fixes/patches instead of upgrading the package. They also requested 'not changing default behavior'.

================
...
Upgrading a package to a new upstream release is generally not permitted in an already-released version of Ubuntu. Please see our policy here:
https://wiki.ubuntu.com/StableReleaseUpdates
...
Hi Liming,
**Changing the mode from emulated to non-emulated would be a regression for those expecting emulated behavior, so that would also violate our SRU policy.**
...
================

tee-supplicant/src/tee_supplicant.c
jan-kiszka: Those distros could be given such a switch to keep the old behavior. In newer major distro releases, this can then be adjusted to something more useful than emulation only or emulation by default.

lsun100:
> I'm happy to close #353 if this one is supposed to take over. I'm still not done with my own testing as I ran into side issues and then out of time.

Thanks for the kindness. I verified this patch on our setup. We could combine it if this patch also works for your case.

tee-supplicant/src/rpmb.c
lsun100: > emulation

The 'emulation' itself seems useful (at least for our case). It's used to test the RPMB related functionalities before switching to real RPMB, which needs to program the one-time permanent RPMB keys.

jan-kiszka: Sure, but you also don't ship debug symbols or allocate debug trace buffers statically when you operate in production mode. I'm just asking for reducing the overhead by a potentially unused mode during runtime.

tee-supplicant/src/tee_supplicant.c
lsun100: > 

Actually we're working with Canonical support to add the real RPMB support (with this patch) into their existing LTS release. They're holding it 'pending the adoption of the pull request'.

jan-kiszka: We have real RPMB support on Debian 12 running, but with custom packages (https://github.com/siemens/meta-iot2050). It is currently quite some effort to do that generically, also because of many missing kernel patches. That's also why I would focus on proper next-release solutions than a specific backport.

tee-supplicant/src/rpmb.c
lsun100: Got it. I misunderstood the top comment. Will update it and post a new version.

lsun100: > 

Updated (in latest 21bad5b6faeb0d1ccaabd93266e91ef5c26ac7e1).

tee-supplicant/src/tee_supplicant.c
lsun100: > 

Ubuntu 22.04 has tee-supplicant package available but compiled with RPMB_EMU. BlueField SoC is shipped with Ubuntu by default and would like to use tee-supplicant with real RPMB. Let me check with distro again how to move forward.

jan-kiszka: Same thing with Debian (likely because upstream has emulation on by default...) - plus that all those extra tweaks to the service file are missing which are needed as long as we have a userspace tee-supplicant in the loop. I'll try to talk to the package maintainer at debconf about how to get it in shape for trixie.

jforissier: @lsun100 @jan-kiszka thank you, your feedback is very much appreciated.

jan-kiszka: Actually, I'm more and more wondering if we shouldn't switch to config file based approach here. Background is that for any such startup policy adjustments, you will not only have to customize the systemd unit file but also the initramfs script that you need in order to use tee-supplicant early (fTPM, efivarfs...).

jan-kiszka: Right, that's (also) why I was advocating for an in-kernel supplicant, at least to make fTPM and StandAloneMM use cases happy.

dannf: Hi - I'm working with @lsun100 to see if there's a way to enable the non-emu version in Ubuntu, as a post-release update. With respect to whether or not `-e` enables or disables emulation, I wonder if we could have both `--rpmb-emu-disable` and `--rpmb-emu-enable` arguments that allow the user to express their intent, with a configurable compile-time default. Then, theoretically, we could add `--rpmb-emu-disable` support leaving the default as "enabled", avoiding a change in behavior to existing users.  The default could later change in a future feature release, but users who are passing one of these two flags could continue to have it honored.

dannf, line 496: Typo - s/emunation/emulation/

lsun100: > 

@dannf , how about @jan-kiszka 's proposal to use "--rpmb-emulation" only. That's to say, use real RPMB by default and specify "--rpmb-emulation" when needed for emulation. In ubuntu 'debian/tee-supplicant.service' file, we could this argument like "ExecStart=/usr/sbin/tee-supplicant --rpmb-emulation"?

dannf: There is a possibility that we have users using the `tee-supplicant` binary directly, or have implemented their own override service. Pushing out an update that requires a new argument to restore existing behavior would break those users.

jan-kiszka: But emulation is a testing toy, not the real purpose of tee-supplicant. I would suggest to give those users the chance to restore the behavior at build-time but avoid bothering the rest.

dannf: @jan-kiszka leaving it as a build-time option may very well be what is best for the upstream project - I'm not qualified to comment on that. My commentary is really just about the impact the chosen implementation would have on the feasibility of including those changes in a post-release update to an existing Ubuntu release. These objectives may end up being incompatible, in which case Ubuntu users will likely need to wait for a future release for a non-emulated support.

tee-supplicant/CMakeLists.txt
```diff
@@ -4,7 +4,6 @@ project (tee-supplicant C)
 # Configuration flags always included
 ################################################################################
 option (CFG_TA_TEST_PATH "Enable tee-supplicant to load from test/debug path" OFF)
-option (RPMB_EMU "Enable tee-supplicant to emulate RPMB" ON)
```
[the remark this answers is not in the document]
jforissier: Would it really help? What would happen if set to OFF? I'd rather simplify things as much as possible.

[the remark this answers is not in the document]
etienne-lms: You mean preserving the config switch but just default it is `OFF`? If so, fine with me.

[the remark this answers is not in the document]
jforissier: No I mean remove this config entirely, because otherwise it would be misleading when set to OFF.

PR #358, opened 2023-09-21
tee-supplicant: Cleanup TEEC_TEST_LOAD_PATH

mikatammi:
Cleanup TEEC_TEST_LOAD_PATH and CFG_TA_TEST_PATH, as the TEEC_TEST_LOAD_PATH was not working properly anyway.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -9,7 +9,6 @@ CFG_TEE_SUPP_LOG_LEVEL ?= 2
 
 # Define Android-specific configuration before including config.mk
 CFG_TEE_CLIENT_LOAD_PATH ?= /vendor/lib
-TEEC_TEST_LOAD_PATH ?= /data/vendor/tee
 CFG_TEE_FS_PARENT_PATH ?= /data/vendor/tee
 CFG_TEE_SUPP_PLUGINS ?= y
 ifneq ($(strip $($(combo_2nd_arch_prefix)TARGET_TOOLS_PREFIX)),)
--- a/config.mk
+++ b/config.mk
@@ -44,12 +44,6 @@ CFG_TEE_SUPP_PLUGINS ?= y
 #   The location of the user plugins
 CFG_TEE_PLUGIN_LOAD_PATH ?= /usr/lib/tee-supplicant/plugins/
 
-# CFG_TA_TEST_PATH
-#   Enable the tee test path.  When enabled, the supplicant will try
-#   loading from a debug path before the regular path.  This allows test
-#   such as 1008.5 that test loading of corrupt TAs.
-CFG_TA_TEST_PATH ?= n
-
 # CFG_GP_SOCKETS
 #   Enable Global Platform Sockets support
 CFG_GP_SOCKETS ?= y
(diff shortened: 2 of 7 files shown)


jenswi-linaro:
In the commit message, please mention that TEEC_TEST_LOAD_PATH and CFG_TA_TEST_PATH are removed and instead, TEEC_LOAD_PATH is supposed to hold all the paths where to search for TAs.

mikatammi:
Done

jforissier:
This most likely breaks `xtest 1008` as mentioned in the removed code. See https://github.com/OP-TEE/build/blob/3.22.0/br-ext/package/optee_client_ext/optee_client_ext.mk#L20. I am fine with removing `CFG_TA_TEST_PATH` but the test environment needs a way to prepend `/tmp` to the load path.

mikatammi:
I'm just thinking after reading the code that it couldn't have worked before. Unless the test puts stuff into directories /foo /bar /baz

mikatammi:
https://optee.mooo.com:5000/logs/OP-TEE/optee_client/358/1524673594/b93b5bbff017bf0f4fbafadbb642800ab39452e5#02-qemu-test

^^^ But doesn't this show the test 1008 as completed?

mikatammi:
So is this fine in its current form, or is more work still needed?

mikatammi:
The other possibility is that I will just fix it to work the way it was intended. But the current state of the mainline, it will always prepend "/foo:/bar::/baz:" to the TEEC_LOAD_PATH, regardless whether user has defined TEEC_TEST_LOAD_PATH or not.

etienne-lms:
Actually, I think `set_ta_path()` function should also be fully removed as well as global variables `ta_path` and `ta_path_str`. `TEECI_LoadSecureModule()` could straight call `try_load_secure_module()` with `"/lib"` as basedir. It's where TAs basedir is expected (aside the buggy leftover _/foo_, _/bar_ and _/baz_ paths cleaned by this P-R).

etienne-lms:
@mikatammi, reading back my last comment I think it is inaccurate. Obviously `set_ta_path()` is needed to split `CFG_TEE_CLIENT_LOAD_PATH` `:`-separated paths  as detailed in the config switch description. 
Your change looks good to me.  `Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`. There is also a ref to in optee_test to be removed.

@jforissier, @jenswi-linaro, I think it's worth having this fix (at least the _/foo:/bar::/baz_ path) in next tag.

jenswi-linaro:

mikatammi:
Good that I didn't have time to fix this :) I will rebase this to remove conflicts

jforissier:
@mikatammi please add Etienne's and Jens' Reviewed-by: tags, thanks!

jforissier:
> I will do!
> 
> > There is also a ref to in optee_test to be removed.
> 
> What does this mean? Should I do something about it?

I don't know :wink:  @etienne-lms?

There is this line however which is now useless: https://github.com/OP-TEE/build/blob/4.0.0/br-ext/package/optee_client_ext/optee_client_ext.mk#L20

etienne-lms:
Yes, I meant if this patch is merged, we should update `optee_client_ext.mk` to remove the pointed line.
Obviously not part of the P-R.

PR #361, opened 2023-10-19
teec: use multiple of page size for page aligned buffers

jenswi-linaro:
When allocating a page aligned buffer, round the size up the next multiple of page size. With this we can guarantee that a part of that page isn't registered already.


jbech-linaro (commented):

jenswi-linaro:
Tag applied.

jforissier:
Bug fix worth taking in 4.0.0 I suppose?

jenswi-linaro:
Yes, please.

jenswi-linaro:
Sadly it doesn't fix https://github.com/OP-TEE/optee_os/issues/6374 but the error looks more or less the same from a normal world point of view so it's nice to remove one source of this error.

PR #369, opened 2024-01-04
tee-supplicant: fix build with kernel < 4.16

ffontaine:
Commit 3ac968ee7c927271e83ea3a4247839649202ab5e moved `linux/tee.h` from `libteec/include` to `libteec/src` resulting in the following build failure with any kernel < 4.16 (i.e before https://github.com/torvalds/linux/commit/033ddf12bcf5326b93bd604f50a7474a434a35f9):

```
/home/buildroot/autobuild/instance-3/output-1/build/optee-client-4.0.0/tee-supplicant/src/tee_supplicant.c: In function 'register_local_shm': /home/buildroot/autobuild/instance-3/output-1/build/optee-client-4.0.0/tee-supplicant/src/tee_supplicant.c:356:44: error: storage size of 'data' isn't known
  356 |         struct tee_ioctl_shm_register_data data;
      |                                            ^~~~
```

To fix this build failure, update `CMakeLists.txt` and `Makefile` of tee-supplicant to add `libteec/src` to the include directories.

Fixes: 3ac968ee7c92 ("Makefile, cmake: move teec related headers")

### DIFF
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -93,7 +93,10 @@ endif()
 ################################################################################
 # Public and private header and library dependencies
 ################################################################################
-target_include_directories(${PROJECT_NAME} PRIVATE src)
+target_include_directories(${PROJECT_NAME}
+	PRIVATE src
+	PRIVATE ../libteec/src
+)
 
 target_link_libraries(${PROJECT_NAME}
 	PRIVATE teec
--- a/tee-supplicant/Makefile
+++ b/tee-supplicant/Makefile
@@ -40,7 +40,7 @@ TEES_OBJ_DIR	:= $(OUT_DIR)
 TEES_OBJS 	:= $(patsubst %.c,$(TEES_OBJ_DIR)/%.o, $(TEES_SRCS))
 TEES_INCLUDES 	:= ${CURDIR}/../libteec/include \
 		   ${CURDIR}/src \
-		   ${CURDIR}/../libteec/include \
+		   ${CURDIR}/../libteec/src \
 
 TEES_CFLAGS	:= $(addprefix -I, $(TEES_INCLUDES)) $(CFLAGS) \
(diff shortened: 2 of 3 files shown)


jenswi-linaro:
Wouldn't that mean exporting that .h file to other packages?

ffontaine:
I'm not an expert in optee-client, I don't even understand why a local copy of `linux/tee.h` is embedded instead of using the one provided by the kernel. An other option on buildroot side would be to add a dependency on kernel 4.16 to optee-client.

jenswi-linaro:
The copy is there to provide early access to new features. It needs to be put somewhere reachable by tee-supplicant too, but it should at the same time not be exported to other packages.

ffontaine:
Then, I would suggest to replace
https://github.com/OP-TEE/optee_client/blob/98dc4f488c2f9e433e3e6574ddbdafd116f523db/tee-supplicant/CMakeLists.txt#L96
by
```
target_include_directories(${PROJECT_NAME}
	PRIVATE src
	PRIVATE ../libteec/src
)
```

Makefile will also have to be updated. Does it seems acceptable?

jenswi-linaro:
Works for me.

ffontaine:
PR updated

jenswi-linaro:
Please add a fixes tag `Fixes: 3ac968ee7c92 ("Makefile, cmake: move teec related headers")`

It would be nice with a short description of what the patch does also apart from all the trouble that the offending commit caused.

ffontaine:
I updated the PR as requested

jforissier (commented):

jenswi-linaro:

ffontaine:
I updated the PR to also fix `tee_supplicant_android.mk`

jforissier:
> I updated the PR to also fix `tee_supplicant_android.mk`

Good. Please add the `Acked-by:` tags, thanks.

PR #380, opened 2024-05-24
tee-supplicant: remove useless use of __ANDROID__ condition

etienne-lms:
Since commit 37975f1ba31e ("rmpb: update AOSP RPMB device path") the RPMB device patch is the same __ANDROID__ directive being set or not hence simplify the implementation by removing use of that directive.


jforissier (commented):
s/patch/path/. With that:

jenswi-linaro:

etienne-lms:
Review tags are applied.

PR #381, opened 2024-05-24
tee-supplicant: add option --unique-rpmb for find unique rpmb device

etienne-lms:
Add tee-supplicant argument --unique-rpmb for when there is only one eMMC device on the platform (hence a single eMMC/RPMB partition) but the mmcblk device index is not known.

### DIFF
--- a/tee-supplicant/src/rpmb.c
+++ b/tee-supplicant/src/rpmb.c
@@ -237,6 +237,15 @@ out:
 	return res;
 }
 
+static bool mmcblk_dev_has_rpmb_partition(uint16_t dev_id)
+{
+	struct stat statbuf = { };
+	char path[32] = { 0 };
+
+	snprintf(path, sizeof(path), "/dev/mmcblk%urpmb", dev_id);
+	return !stat(path, &statbuf) && (statbuf.st_mode & S_IFMT) == S_IFCHR;
+}
+
 static int hexchar2int(char c)
 {
 	if (c >= '0' && c <= '9')
@@ -340,13 +349,16 @@ static bool remap_rpmb_dev_id(uint16_t dev_id, uint16_t *ndev_id)
 	static uint16_t id = 0;
 	char cid[CID_STR_SZ + 1] = { };
 	struct dirent *dent = NULL;
+	uint unique_rpmb_id = 0;
+	uint rpmb_count = 0;
 	DIR *dir = NULL;
 	int num = 0;
 
 	if (err || found)
 		goto out;
 
-	if (!supplicant_params.rpmb_cid) {
+	if (!supplicant_params.rpmb_cid &&
+	    !supplicant_params.find_unique_mmc_rpmb) {
 		id = dev_id;
 		found = true;
 		goto out;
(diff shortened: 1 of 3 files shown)


etienne-lms:
This cmdline option targets system with only 1 eMMC device. As soon as there are 2 or more, we need to differentiate them. This P-R treat case where we don't need that. It's likely there is only 1 eMMC device on many embedded systems, but maybe also one (or more) SDcard slots that can easily be distinguish from the eMMC among the _/dev/mmclk*_ devices.
Your RPMB changes from #6852 will provide a more flexible discovery, finding a matching RPMB key.  This lightweight change aims at being easily portable on system with older kernels (and a single eMMC device).

jenswi-linaro:
I understand the intent. What should happen if there's more than one? If the option is given by mistake, what is the best action? Refuse to start or take the first found device?

jforissier:
IMO the handling of RPMB in OP-TEE is complicated enough, we should refrain from adding more options to tee-supplicant since we know that the proper solution is the in-kernel RPMB support.

etienne-lms:
> If the option is given by mistake, what is the best action? Refuse to start or take the first found device?

If the option is given by mistake, I would assume it's a programming error, as when providing any other invalid cmdline argument.


> IMO the handling of RPMB in OP-TEE is complicated enough, we should refrain from adding more options to tee-supplicant since we know that the proper solution is the in-kernel RPMB support.

The issue I see is that in-kernel RPMB will likely not be ported in old Linux kernel, event LTSes, and maybe not straight forward to backport once merged in a next v6.x.

jenswi-linaro:
The new probe ABI is designed to work with tee-supplicant too. So if we're looking for compatibility for older kernels we can very well implement the ABI in tee-supplicant. If the kernel supports the ABI it will seamlessly take precedence.

etienne-lms:
What about currently existing implementation? With current OP-TEE distribution (manifest.git+build.git) based on Linux kernel v6.6, I face non-systematic boot issues on stm32mp15C-ed1/ev1 boards: these embed a SDcard slot and an eMMC device. Linux v6.6 assign device index 0 or 1 to each (**/dev/mmcblk{0|1}**) which can't be predicted (AFAICT).

How to overcome that? Wait Linux kernel and OP-TEE OS implement the ABI, then backport? I the mean time, these boards are not stable. 

If you don't want to add a tee-supplicant cmdline option that will be useless in the near futur, would a build configuration switch be more acceptable in tee-supplicant?

jenswi-linaro:
Another option is to dig out the CID from sysfs and supply that with `--rpmb-cid`.

jforissier:
> Another option is to dig out the CID from sysfs and supply that with `--rpmb-cid`.

@etienne-lms see https://gitlab.com/Linaro/trustedsubstrate/meta-ledge-secure/-/blob/main/meta-ledge-secure/recipes-security/optee/optee-client/create-tee-supplicant-env

etienne-lms:
Thanks @jforissier, I think this script should nicely suit my needs, preventing me from building a distrib that would be specific to a given eMMC device.

etienne-lms:
Sorry @jenswi-linaro, I see that I misunderstood your https://github.com/OP-TEE/optee_client/pull/381#issuecomment-2166010997 that suggested that. 

I close this P-R and create another one retrieving and passing the eMCC CID.

etienne-lms:
For the record: superseded by https://github.com/OP-TEE/build/pull/760

Issue #382, opened 2024-05-29
tee-supplicant readiness from systemd


embetrix:
I use tee-supplicant to start from systemd unit as follow:

```
[Unit]
Description=TEE Supplicant
Before=myapps.target
DefaultDependencies=no

[Service]
ExecStart=/usr/sbin/tee-supplicant

[Install]
WantedBy=basic.target
```

systemd considers the unit as ready as soon as tee-supplicant starts, which causes problem of the applications  belonging to **myapps.target** as tee-supplicant sometimes is not ready yet and doesn't have access to the optee-storage.

I would like to change the unit to **Type=notify** as follow:

```
[Unit]
Description=TEE Supplicant
Before=myapps.target
DefaultDependencies=no

[Service]
Type=notify
ExecStart=/usr/sbin/tee-supplicant

[Install]
WantedBy=basic.target
```
But to signal readiness to systemd I have to send **sd_notify(0, "READY=1")** call.

what will be the best place in tee-supplicant code to place the call ?  ( I use only PKCS11 TA)

**ExecStart=/usr/sbin/tee-supplicant -d**  does't work here as it will be forking and cannot be used with Type=notify

jforissier:
> systemd considers the unit as ready as soon as tee-supplicant starts, which causes problem of the applications  belonging to **myapps.target** as tee-supplicant sometimes is not ready yet and doesn't have access to the optee-storage.

OP-TEE is supposed to be able to wait for tee-supplicant to be ready, at least when it requests a TA to be loaded. I verified this in the QEMU environment (start the environment, kill `tee-supplicant`, run `xtest`: it waits, hit Ctr-C, run `/usr/bin/tee-supplicant -d`, run `xtest` again: it works). What kind of error do you get?

jforissier:
CC @mcfrisk

embetrix:
> > systemd considers the unit as ready as soon as tee-supplicant starts, which causes problem of the applications  belonging to **myapps.target** as tee-supplicant sometimes is not ready yet and doesn't have access to the optee-storage.
> 
> OP-TEE is supposed to be able to wait for tee-supplicant to be ready, at least when it requests a TA to be loaded. I verified this in the QEMU environment (start the environment, kill `tee-supplicant`, run `xtest`: it waits, hit Ctr-C, run `/usr/bin/tee-supplicant -d`, run `xtest` again: it works). What kind of error do you get?

[...]

PR #383, opened 2024-05-30
tee-supplicant: send READY=1 notification to systemd

embetrix:
This is very useful when tee-supplicant is started from systemd and can used with **Type=notify** to signal readiness

https://github.com/OP-TEE/optee_client/issues/382

### DIFF
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -28,6 +28,7 @@ set(SRC
 	src/hmac_sha2.c
 	src/rpmb.c
 	src/sha2.c
+	src/sd_notify.c
 	src/tee_supp_fs.c
 	src/tee_supplicant.c
 	src/teec_ta_load.c
--- a/tee-supplicant/src/sd_notify.h
+++ b/tee-supplicant/src/sd_notify.h
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MIT-0
+/*
+ * The code below is imported from:
+ * https://www.freedesktop.org/software/systemd/man/devel/sd_notify.html#Standalone%20Implementations
+ */
+#ifndef SD_NOTIFY_H
+#define SD_NOTIFY_H
+
+int sd_notify_ready(void);
+
+#endif /* SD_NOTIFY_H */
--- a/tee-supplicant/src/tee_supplicant.c
+++ b/tee-supplicant/src/tee_supplicant.c
@@ -53,6 +53,7 @@
 #include <tee_socket.h>
 #include <tee_supp_fs.h>
 #include <tee_supplicant.h>
+#include <sd_notify.h>
 #include <unistd.h>
 
 #include "optee_msg_supplicant.h"
@@ -922,6 +923,11 @@ int main(int argc, char *argv[])
 			exit(EXIT_FAILURE);
 		}
 	}
+
+	/* we are set here notify systemd */
+	e = sd_notify_ready();
+	if(e < 0)
+		fprintf(stderr, "sd_notify_ready() failed: %s\n", strerror(-e));
 
 	if (daemonize) {
 		/* Release parent */
--- a/tee-supplicant/Makefile
+++ b/tee-supplicant/Makefile
@@ -18,7 +18,8 @@ TEES_SRCS	:= tee_supplicant.c \
 		   teec_ta_load.c \
 		   tee_supp_fs.c \
 		   rpmb.c \
-		   handle.c
+		   handle.c \
+		   sd_notify.c
 
 ifeq ($(CFG_GP_SOCKETS),y)
 TEES_SRCS 	+= tee_socket.c
--- a/tee-supplicant/src/sd_notify.c
+++ b/tee-supplicant/src/sd_notify.c
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: MIT-0
+/*
+ * The code below is imported from:
+ * https://www.freedesktop.org/software/systemd/man/devel/sd_notify.html#Standalone%20Implementations
+ */
+
+#define _GNU_SOURCE 1
+#include <string.h>
+#include <errno.h>
+#include <inttypes.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "sd_notify.h"
+
+#define _cleanup_(f) __attribute__((cleanup(f)))
+
+static void closep(int *fd)
+{
+    if (!fd || *fd < 0)
+        return;
+
+    close(*fd);
+    *fd = -1;
+}
+
+static int notify(const char *message)
+{
+    union sockaddr_union {
+        struct sockaddr sa;
+        struct sockaddr_un sun;
+    } socket_addr = {
+        .sun.sun_family = AF_UNIX,
+    };
+
+    ssize_t written = 0;
+    size_t path_length, message_length = 0;
+    _cleanup_(closep) int fd = -1;
+    const char *socket_path = NULL;
+
+    /* Verify the argument first */
+    if (!message)
+       return -EINVAL;
+
+    message_length = strlen(message);
+    if (message_length == 0)
+        return -EINVAL;
+
+    /* If the variable is not set, the protocol is a noop */
+    socket_path = getenv("NOTIFY_SOCKET");
+    if (!socket_path)
+       return 0; /* Not set? Nothing to do */
+
+    /* Only AF_UNIX is supported, with path or abstract sockets */
+    if (socket_path[0] != '/' && socket_path[0] != '@')
+        return -EAFNOSUPPORT;
+
+    path_length = strlen(socket_path);
+    /* Ensure there is room for NULL byte */
+    if (path_length >= sizeof(socket_addr.sun.sun_path))
+        return -E2BIG;
+
+    memcpy(socket_addr.sun.sun_path, socket_path, path_length);
+
+    /* Support for abstract socket */
+    if (socket_addr.sun.sun_path[0] == '@')
+        socket_addr.sun.sun_path[0] = 0;
+
+    fd = socket(AF_UNIX, SOCK_DGRAM|SOCK_CLOEXEC, 0);
+    if (fd < 0)
+        return -errno;
+
+    if (connect(fd, &socket_addr.sa, offsetof(struct sockaddr_un, sun_path) + path_length) != 0)
+        return -errno;
+
+    written = write(fd, message, message_length);
+    if (written != (ssize_t) message_length)
+        return written < 0 ? -errno : -EPROTO;
+
+    return 1; /* Notified! */
+}
+
+int sd_notify_ready(void)
+{
+    return notify("READY=1");
+}


flags.mk
```diff
@@ -9,6 +9,8 @@ PKG_CONFIG	?= $(CROSS_COMPILE)pkg-config
 
 C_COMPILER=$(shell readlink -f $$(which $(CC)))
 
+override LDFLAGS += -ldl
```
jforissier, line 12: `dl` needs to be added to `tee-supplicant/CMakeLists.txt`, too (it is already there but conditional)

tee-supplicant/src/tee_supplicant.c
```diff
@@ -493,6 +494,8 @@ static int usage(int status)
 	fprintf(stderr, "\t-h, --help: this help\n");
 	fprintf(stderr, "\t-d, --daemonize: run as a daemon (fork and return "
 			"after child has opened the TEE device or on error)\n");
+	fprintf(stderr, "\t-n, --sdnotify: signal READY=1 to systemd "
+			"after tee-supplicant is setup)\n");
```
jforissier, line 498: Is this option really needed? What if we notify always?

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,28 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if ((sdnotify) && (!daemonize)) {
```
jforissier, line 934: What prevents from notifying when `deamonize` is true?

tee-supplicant/src/tee_supplicant.c
```diff
@@ -835,6 +839,7 @@ int main(int argc, char *argv[])
 		/* long name      | has argument  | flag | short value */
 		{ "help",            no_argument,       0, 'h' },
 		{ "daemonize",       no_argument,       0, 'd' },
+		{ "sdnotify",        no_argument,       0, 'n' },
```
jenswi-linaro, line 842: `-n` usually means dry-run in other utilities. How about `-s` instead?

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,28 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if ((sdnotify) && (!daemonize)) {
+		/* we are set here notify systemd */
+		int(*__sd_notify__)();
```
jenswi-linaro: The `__` are ugly, I suppose you've added them to avoid conflicts with `sd_notify()` in `<systemd/sd-daemon.h>`. How about `my_sd_notify(int unset_environment, const char *state)`?
Please initialize all local variables.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,28 @@ int main(int argc, char *argv[])
...
+		void *systemd = dlopen("libsystemd.so", RTLD_LAZY);
+		if (systemd) {
+			*(int**)(&__sd_notify__) = dlsym(systemd, "sd_notify");
```
jenswi-linaro, line 939: Why do you need all this casting?

tee-supplicant/src/tee_supplicant.c
jenswi-linaro, line 934: The extra `()` are not needed.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,28 @@ int main(int argc, char *argv[])
...
+			if (__sd_notify__) {
+					int ret = __sd_notify__(0, "READY=1");
+					if (ret <= 0) {
```
jenswi-linaro, line 942: The extra `{}` are not needed.

embetrix:
@jforissier : I just figured out on the target testing that yocto doesn't ship libsystemd.so

but only versioned *so.*

```
# ls -l /lib/libsystemd.so*
lrwxrwxrwx 1 root root     20 Jul  9  2020 /lib/libsystemd.so.0 -> libsystemd.so.0.33.0
-rwxr-xr-x 1 root root 989248 Jul  9  2020 /lib/libsystemd.so.0.33.0
```

jforissier:
@embetrix `dlopen("libsystemd.so.0", RTLD_LAZY);` should work fine. Using `libsystemd.so` is actually not a good idea because you don't know which major version it is. It could lead to compatibility issues.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,27 @@ int main(int argc, char *argv[])
...
+ 		int (*sd_notify)(int unset_environment, const char *state) =  NULL;
+		sd_handle = dlopen("libsystemd.so", RTLD_LAZY);
+		if (sd_handle) {
+			sd_notify = dlsym(sd_handle, "sd_notify");
+			if (sd_notify) {
+					ret = sd_notify(0, "READY=1");
+					if (ret <= 0)
+						fprintf(stderr, "sd_notify failed: %d\n", ret);
+			}
+			else
```
jenswi-linaro, line 947: Please fold up `else` to make it `} else`

embetrix:
> @embetrix `dlopen("libsystemd.so.0", RTLD_LAZY);` should work fine. Using `libsystemd.so` is actually not a good idea because you don't know which major version it is. It could lead to compatibility issues.

done

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,27 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	/* we are set here notify systemd */
+	if (sdnotify && !daemonize) {
+		int ret = 0;
+		void *sd_handle = NULL;
+ 		int (*sd_notify)(int unset_environment, const char *state) =  NULL;
+		sd_handle = dlopen("libsystemd.so.0", RTLD_LAZY);
```
jenswi-linaro, line 939: Please leave an empty line between variables and code.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +931,27 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	/* we are set here notify systemd */
+	if (sdnotify && !daemonize) {
+		int ret = 0;
+		void *sd_handle = NULL;
+ 		int (*sd_notify)(int unset_environment, const char *state) =  NULL;
```
jenswi-linaro, line 938: This is bound to cause a compile error if `<systemd/sd-daemon.h>` is ever included, that's why I suggested `my_sd_notify` or such.

tee-supplicant/src/tee_supplicant.c
jenswi-linaro: Good question.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +926,20 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	/* we are set here notify systemd */
+	sd_handle = dlopen("libsystemd.so.0", RTLD_LAZY);
+	if (sd_handle) {
+		tee_sd_notify = dlsym(sd_handle, "sd_notify");
+		if (tee_sd_notify) {
+				e = tee_sd_notify(0, "READY=1");
```
mikkorapeli-linaro, line 934: While at it, could you add a "tee_sd_notify(0, "WATCHDOG=1"); into the main loops which call process_one_request()?

What could be the maximum time that a process_one_request() call takes? We could then double this and add to systemd service file to detect if tee-supplicant hangs. I've seen hangs related ftpm kernel module unloading which could affect tee-supplicant too, for example.

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +926,20 @@ int main(int argc, char *argv[])
...
+				if (e <= 0)
+					fprintf(stderr, "sd_notify failed: %d\n", e);
+		} else
+			fprintf(stderr, "Couldn't find sd_notify symbol: %s\n", dlerror());
+		dlclose(sd_handle);
```
mikkorapeli-linaro, line 939: If watchdog calls are done, I'd leave this open until exiting mainloops

tee-supplicant/src/tee_supplicant.c
jenswi-linaro: How long does it take to copy a file (of unknown size) from the file system into memory, worst case?

mikkorapeli-linaro: Ok, the watchdog target is application dependent but the sd_notify() call to kick the watchdog could be generic in the loop.

Users could set the application specific watchdog timeout in the systemd service file for tee-supplicant. Default could be that it's not set at all.

jforissier: The `ioctl()` call is blocking until a request comes, with no timeout. So we would at least need to add a timeout before adding support for such a watchdog. An alternative would be to ping from a separate thread but that's obviously a weaker health guarantee.

mikkorapeli-linaro: The watchdog call from tee-supplicant to systemd is just a health signal. It can block for ever. The timeout value is set in systemd service file, which can be unset by default, or a really large value. Dending on use cases and TAs, then this value can be reduced.

But it's ok to skip this for now too.

tee-supplicant/src/tee_supplicant.c
poettering: my recommendation (as systemd upstream): just always send the notification, it internally conditionalizes things anyway based on an env var that is either passed or not, hence a second level of conditionalization should really not be necessary

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +926,20 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	/* we are set here notify systemd */
+	sd_handle = dlopen("libsystemd.so.0", RTLD_LAZY);
```
poettering, line 930: sorry for parachuting in here, I was pointed to this PR from here: 

https://lists.freedesktop.org/archives/systemd-devel/2024-June/050385.html

I just wanted to mention that for a trivial call like sd_notify() you might just opt directly implementing this, rather than doing dlopen()/dlsym(). i.e. something like this:

https://www.freedesktop.org/software/systemd/man/devel/sd_notify.html#Standalone%20Implementations

tee-supplicant/src/tee_supplicant.c
poettering: also note that libsystemd.so is nowadays built with `-z nodelete`, i.e. we do not support being unloaded, and the dlclose() is hence without effect

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +926,20 @@ int main(int argc, char *argv[])
...
+	if (sd_handle) {
+		tee_sd_notify = dlsym(sd_handle, "sd_notify");
+		if (tee_sd_notify) {
+				e = tee_sd_notify(0, "READY=1");
+				if (e <= 0)
+					fprintf(stderr, "sd_notify failed: %d\n", e);
```
poettering, line 936: please do not consider `e == 0` an error, it's really not. It just means no message sent because $NOTIFY_SOCKET wasn't set, i.e. because it was invoked in a non-systemd context, which is entirely fine however.

tee-supplicant/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(SRC
 	src/hmac_sha2.c
 	src/rpmb.c
 	src/sha2.c
+	src/tee_sd_notify.c
```
jforissier, line 31: Let's drop the `tee_` prefix, it's distraction

tee-supplicant/src/tee_sd_notify.c
```diff
@@ -0,0 +1,81 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
jforissier, line 1:
```C
// SPDX-License-Identifier: MIT-0
/*
 * The code below is imported from:
 * https://www.freedesktop.org/software/systemd/man/devel/sd_notify.html#Standalone%20Implementations
 */

...
```

tee-supplicant/src/tee_sd_notify.h
```diff
@@ -0,0 +1,28 @@
...
+#include <inttypes.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <time.h>
+#include <unistd.h>
```
jforissier, line 24: Move to `sd_notify.c`

tee-supplicant/src/tee_sd_notify.c
```diff
@@ -0,0 +1,81 @@
...
+        return -errno;
+
+    written = write(fd, message, message_length);
+    if (written != (ssize_t) message_length)
+        return written < 0 ? -errno : -EPROTO;
+
+    return 1; /* Notified! */
+}
+
+int tee_sd_notify_ready(void)
```
jforissier, line 78: `sd_notify_ready()`

mikkorapeli-linaro:
Tested latest version of patch with TS and TRS in https://ledge.validation.linaro.org/scheduler/job/88692 and it works. tee-supplicant startup state is correctly reported to systemd.

tee-supplicant/src/sd_notify.h
```diff
@@ -0,0 +1,25 @@
...
+#include <inttypes.h>
+#include <signal.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <time.h>
+#include <unistd.h>
```
jforissier, line 21: Move the includes to `sd_notify.c`

tee-supplicant/src/tee_supplicant.c
```diff
@@ -923,6 +924,12 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	/* we are set here notify systemd */
+	e = sd_notify_ready();
+	if(e < 0) {
+		fprintf(stderr, "sd_notify_ready failed: %s\n", strerror(-e));
+	}
```
jforissier, line 931: Style:
- `{ }` not needed.
- "sd_notify_ready() failed" (parentheses after function names)

jforissier (commented):
- See comments below.
- Please use prefix in commit subject: "tee-supplicant: send READY=1 notification to systemd"
- Please add your `Signed-off-by:`

embetrix:
> * See comments below.
> * Please use prefix in commit subject: "tee-supplicant: send READY=1 notification to systemd"
fixed
> * Please add your `:`
added Signed-off-by

jforissier (approved):
Please wrap the commit description at 72 characters and add:

@poettering thanks for your recommendations

embetrix:
> Please wrap the commit description at 72 characters and add:
> 
> 
> @poettering thanks for your recommendations

done

jforissier:
Any more feedback? @mikkorapeli-linaro would you mind giving an Ack and/or Tested-by perhaps?

mikkorapeli-linaro:
> Any more feedback? @mikkorapeli-linaro would you mind giving an Ack and/or Tested-by perhaps?

Yes, latest patch version also

Tested-by: Mikko Rapeli <mikko.rapeli@linaro.org>

TS firmware with optee 4.1 and RPMB kernel patches from Jens, and TRS uki kernel and initramfs with systemd which starts tee-supplicant using this patch to create fTPM encrypted rootfs, on rockpi4b. The few remaining failures in TS/TRS test suite are not related to this patch.

Test run:
https://ledge.validation.linaro.org/scheduler/job/88800

jforissier:
Mikko's `Tested-by:` tag added and pushed to master. Thanks!

Issue #388, opened 2024-09-12
Null-Pointer Dereference in serialize_mecha_XXX()


hoyong2007:
Commit: [a5b1ffc](https://github.com/OP-TEE/optee_client/commit/a5b1ffcd26e328af0bbf18ab448a38ecd558e05c) (master)
libckteec/src/serialize_ck.c
Function: serialize_mecha_XXX()

There are lots of potential Null-Pointer deference spots in serialize_ck.c

```c
static CK_RV serialize_mecha_rsa_oaep_param(struct serializer *obj,
					    CK_MECHANISM_PTR mecha)
{
	CK_RSA_PKCS_OAEP_PARAMS *params = mecha->pParameter;
	CK_RV rv = CKR_GENERAL_ERROR;
	size_t params_size = 4 * sizeof(uint32_t) + params->ulSourceDataLen;

	if (mecha->ulParameterLen != sizeof(*params))
		return CKR_ARGUMENTS_BAD;
	
	...
	
	return serialize_buffer(obj, params->pSourceData,
				params->ulSourceDataLen);
}

```

For example, in the `serialize_mecha_rsa_oaep_param()` function, `mecha->pParameter` is referenced directly without proper pointer validation.

Additionally, the `params->pSourceData` pointer is also directly referenced in the `serialize_buffer()` function without proper pointer validation.

Directly referencing pointers without proper validation, as shown, can lead to null pointer dereference bugs, which may result in a DoS (Denial of Service).

Credit: @hoyong2007 @jch6637

etienne-lms:
Thanks for the investigation.

Using null pointers with libckteec will only crash the Linux client application that invokes the TA. It won't affect other Linux applications neither the pkcs11 TA itsleft. So I don't thinks these can lead to DoS attacks.

We could add non-NULL pointer verification for non-zero sized buffer references but that would only help for development/debugging. For example, there is no way to check pointers are well sized. It is the responsibility of the client application to contain programming errors. That said, it should not add much complexity to test some pointer here and there.

hoyong2007:
Thanks for your reply.

I'm focusing on the bugs via NULL pointer, not non-NULL pointer. I agreed with that the client cannot check the actual buffer size of the pointer, but I believe the client should at least verify whether the pointer is null.

As far as I know, the mechanism pointer is checked in the higher-level functions, but the pointer members of the mechanism are not checked for null. I believe every pointer variable should be checked at least once before being referenced.

[...]

PR #392, opened 2024-10-11
libckteec: fix null pointer dereference on template serialization

hoyong2007:
Fix potential null-pointer dereference on serialize_mecha_XXX() by checking every pointer value in under CK_MECHANISM_PTR.

Link: https://github.com/OP-TEE/optee_client/issues/388

### DIFF
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -393,6 +393,9 @@ static CK_RV serialize_mecha_aes_ctr(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	uint32_t size = 0;
 
+	if (!param)
+		return CKR_MECHANISM_PARAM_INVALID;
+
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
 		return rv;
@@ -420,12 +423,12 @@ static CK_RV serialize_mecha_aes_gcm(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_ULONG aad_len = 0;
 
+	if (!param || !param->pIv)
+		return CKR_MECHANISM_PARAM_INVALID;
+
 	/* AAD is not manadatory */
 	if (param->pAAD)
 		aad_len = param->ulAADLen;
-
-	if (!param->pIv)
-		return CKR_MECHANISM_PARAM_INVALID;
 
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
@@ -461,6 +464,9 @@ static CK_RV serialize_mecha_aes_iv(struct serializer *obj,
 	uint32_t iv_size = mecha->ulParameterLen;
 	CK_RV rv = CKR_GENERAL_ERROR;
 
+	if (iv_size && !mecha->pParameter)
+		return CKR_MECHANISM_PARAM_INVALID;
+
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
 		return rv;
@@ -478,6 +484,9 @@ static CK_RV serialize_mecha_key_deriv_str(struct serializer *obj,
 	CK_KEY_DERIVATION_STRING_DATA_PTR param = mecha->pParameter;
 	CK_RV rv = CKR_GENERAL_ERROR;
 	uint32_t size = 0;
+
+	if (!param || (param->ulLen && !param->pData))
+		return CKR_MECHANISM_PARAM_INVALID;
 
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
@@ -500,7 +509,14 @@ static CK_RV serialize_mecha_ecdh1_derive_param(struct serializer *obj,
 {
 	CK_ECDH1_DERIVE_PARAMS *params = mecha->pParameter;
 	CK_RV rv = CKR_GENERAL_ERROR;
-	size_t params_size = 3 * sizeof(uint32_t) + params->ulSharedDataLen +
+	size_t params_size = 0;
+	
+	if (!params ||
+	    (params->ulSharedDataLen && !params->pSharedData) ||
+	    (params->ulPublicDataLen && !params->pPublicData))
+		return CKR_MECHANISM_PARAM_INVALID;
+
+	params_size = 3 * sizeof(uint32_t) + params->ulSharedDataLen +
 			     params->ulPublicDataLen;
 
 	rv = serialize_32b(obj, obj->type);
@@ -539,6 +555,9 @@ static CK_RV serialize_mecha_aes_cbc_encrypt_data(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	uint32_t size = 0;
 
+	if (!param || (param->length && !param->pData))
+		return CKR_MECHANISM_PARAM_INVALID;
+	
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
 		return rv;
@@ -565,6 +584,9 @@ static CK_RV serialize_mecha_rsa_pss_param(struct serializer *obj,
 	CK_RSA_PKCS_PSS_PARAMS *params = mecha->pParameter;
 	CK_RV rv = CKR_GENERAL_ERROR;
 	uint32_t params_size = 3 * sizeof(uint32_t);
+
+	if (!params)
+		return CKR_MECHANISM_PARAM_INVALID;
 
 	if (mecha->ulParameterLen != sizeof(*params))
 		return CKR_ARGUMENTS_BAD;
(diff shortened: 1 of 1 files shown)


libckteec/src/serialize_ck.c
```diff
@@ -593,7 +613,12 @@ static CK_RV serialize_mecha_rsa_oaep_param(struct serializer *obj,
 {
 	CK_RSA_PKCS_OAEP_PARAMS *params = mecha->pParameter;
 	CK_RV rv = CKR_GENERAL_ERROR;
-	size_t params_size = 4 * sizeof(uint32_t) + params->ulSourceDataLen;
+	size_t params_size = 0;
+	
+	if (!params || !params->pSourceData)
```
etienne-lms, line 618: For consistency, prefer also to test `(params->ulSourceDataLen && !params->pSourceData)`.

libckteec/src/serialize_ck.c
```diff
@@ -630,9 +655,18 @@ static CK_RV serialize_mecha_rsa_aes_key_wrap(struct serializer *obj,
...
-	size_t params_size = 5 * sizeof(uint32_t) + aes_params->ulSourceDataLen;
+	size_t params_size = 0;
+
+	if (!params || !params->pOAEPParams)
+		return CKR_MECHANISM_PARAM_INVALID;
+
+	aes_params = params->pOAEPParams;
+	params_size = 5 * sizeof(uint32_t) + aes_params->ulSourceDataLen;
+
+	if (!aes_params->pSourceData)
```
etienne-lms, line 668: `aes_params->ulSourceDataLen && !aes_params->pSourceData`

libckteec/src/serialize_ck.c
```diff
@@ -675,6 +709,9 @@ static CK_RV serialize_mecha_eddsa(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_EDDSA_PARAMS *params = mecha->pParameter;
 
+	if (!params || !params->pContextData)
```
etienne-lms, line 712: `!params || (params->ulContextDataLen && !params->pContextData)`

libckteec/src/serialize_ck.c
```diff
@@ -700,6 +737,9 @@ static CK_RV serialize_mecha_mac_general_param(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_ULONG ck_data = 0;
 
+	if (!mecha->pParameter)
```
etienne-lms, line 740: `mecha->ulParameterLen && !mecha->pParameter`

libckteec/src/serialize_ck.c
```diff
@@ -539,6 +553,9 @@ static CK_RV serialize_mecha_aes_cbc_encrypt_data(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	uint32_t size = 0;
 
+	if (!param || !param->pData)
```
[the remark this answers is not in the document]
hoyong2007: Agreed. If the CA is implemented correctly, ensuring that the correct buffer size is passed, null-pointer dereferences may no longer occur even if malloc in the CA returns NULL. Let's leave the task of checking whether the parameter is valid to the TA :)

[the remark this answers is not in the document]
hoyong2007: > When the data is not well sized, the cryptoki API function are expected to return `CKR_DATA_LEN_RANGE`. Here also, I think we should let the TA bother with that and only ensures they is no unexpected segmentation faults:
> 
> ```c
> 	if (!param || (param->length && !param->pData))
> 		return CKR_MECHANISM_PARAM_INVALID;
> ```

Agreed. If the CA is implemented correctly, ensuring that the correct buffer size is passed, null-pointer dereferences may no longer occur even if malloc in the CA returns NULL. Let's leave the task of checking whether the parameter is valid to the TA :)

libckteec/src/serialize_ck.c
```diff
@@ -461,6 +464,9 @@ static CK_RV serialize_mecha_aes_iv(struct serializer *obj,
 	uint32_t iv_size = mecha->ulParameterLen;
 	CK_RV rv = CKR_GENERAL_ERROR;
 
+	if (!mecha->pParameter)
```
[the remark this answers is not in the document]
hoyong2007: Can we use `mecha->ulParameterLen` instead of `iv_size`? Since `mecha->ulParameterLen` is already used when `serialize_buffer()` is called.

libckteec/src/serialize_ck.c
```diff
@@ -500,7 +509,12 @@ static CK_RV serialize_mecha_ecdh1_derive_param(struct serializer *obj,
 {
 	CK_ECDH1_DERIVE_PARAMS *params = mecha->pParameter;
 	CK_RV rv = CKR_GENERAL_ERROR;
-	size_t params_size = 3 * sizeof(uint32_t) + params->ulSharedDataLen +
+	size_t params_size = 0;
+	
+	if (!params || !params->pSharedData || !params->pPublicData)
```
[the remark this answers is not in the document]
hoyong2007: It seems appropriate because if the length is 0, the memcpy called within serialize_buffer won't reference the data and will simply return.

libckteec/src/serialize_ck.c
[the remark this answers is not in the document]
hoyong2007: If you think using `iv_size` seems more appropriate, feel free to resolve this conversation since it’s already reflected in the commit.

libckteec/src/serialize_ck.c
```diff
@@ -420,13 +423,13 @@ static CK_RV serialize_mecha_aes_gcm(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_ULONG aad_len = 0;
 
+	if (!param || !param->pIv)
```
etienne-lms, line 426: I think test on `!param->pIv` was not really useful, testing `!param` should be enough. That said, for consistency, could you test `!param || (param->ulIvLen && !param->pIv)`

etienne-lms (commented):
Sorry, I think I answered too quickly on these changes.  For all cases where a buffer pointer is passed together with its buffer size, we should not have to strictly check the buffer is not NULL when the size is not 0. Note that we cannot verify that the passed pointer is really well sized regarding the passed size. That said, your changes add no regression and will have negligible impact on the API function execution performances so I'm fine we consider them. 

By the way, I see that few of the functions in _serialize_ck.c_ already return a error when `mecha->ulParameterLen != sizeof(*params)` but not all function do. For consistency, I think all these functions should. But that would need dedicated commit since your commit initially addresses potential NULL buffer references. Would you mind adding such a commit to your P-R?

By the way, I've not posted a `Review-by` explicit comment for this P-R so you should not add my R-b tag  in the commit message.

libckteec/src/serialize_ck.c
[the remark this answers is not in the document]
etienne-lms: Testing `iv_size` or `mecha->ulParameterLen` is equivalent. At your will.

libckteec/src/serialize_ck.c
```diff
@@ -500,7 +509,14 @@ static CK_RV serialize_mecha_ecdh1_derive_param(struct serializer *obj,
...
 	CK_RV rv = CKR_GENERAL_ERROR;
-	size_t params_size = 3 * sizeof(uint32_t) + params->ulSharedDataLen +
+	size_t params_size = 0;
+	
+	if (!params ||
+	    (params->ulSharedDataLen && !params->pSharedData) ||
+	    (params->ulPublicDataLen && !params->pPublicData))
+		return CKR_MECHANISM_PARAM_INVALID;
+
+	params_size = 3 * sizeof(uint32_t) + params->ulSharedDataLen +
```
etienne-lms, line 519: Could you fix indentation at line below?

libckteec/src/serialize_ck.c
etienne-lms: Looking again at PKCS#11 specifications, on one hand v2.40-errata01 says both AES-GCM [IV can be NULL](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850510) and must be between [1 and 256 byte](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850509). On the other hand v2.30 says AES-GCM IV must be between [1 and 2^32-1 bytes](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061261). In all one, I think the legacy implementation here in optee_client is enough, so OK for your proposal `if (!param || !param->pIv)`.

etienne-lms (commented):
@hoyong2007, sorry for this very late feedback.
Changes look good to me, just an indentation to fix.

PR #394, opened 2024-11-29
Validate EdDSA mechanism parameter length

mzachar:
I run into an segmentation fault when using following command:

`$ pkcs11-tool --module /usr/lib/libckteec.so.0 --login --sign -m EDDSA --slot 0 --id 01 --input-file /tmp/data --output-file /tmp/data.sig`

After bit of digging I found out that there is no parameterLen validation in EdDSA mechanism branch which result in null pointer dereference.

This fixes Segmentation fault when no parameter is provided as specified in pkcs11 v3.0 spec for Ed25519 Signature Scheme

### DIFF
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -674,6 +674,23 @@ static CK_RV serialize_mecha_eddsa(struct serializer *obj,
 {
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_EDDSA_PARAMS *params = mecha->pParameter;
+	CK_ULONG params_len = mecha->ulParameterLen;
+	/*
+	 * When no parameter is provided, the expected operation is
+	 * no-prehash and no-context.
+	 */
+	CK_EDDSA_PARAMS default_params = {
+		.phFlag = 0,
+		.ulContextDataLen = 0,
+	};
+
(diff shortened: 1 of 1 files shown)


libckteec/src/serialize_ck.c
```diff
@@ -674,6 +674,16 @@ static CK_RV serialize_mecha_eddsa(struct serializer *obj,
 {
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_EDDSA_PARAMS *params = mecha->pParameter;
+	CK_ULONG paramsLen = mecha->ulParameterLen;
```
etienne-lms, line 677: prefer snake_case labels, for conding style consistency in libckteec:
`paramsLen` => `params_len`
`defaultParams` => `default_params`

etienne-lms (commented):
Thanks a lot @mzachar for reporting and fixing this issue.

Could you add prefix `libckteec: ` to the commit message header line? With that addressed, 
`Reviewed-by: Etienne Carriere <etienne.carriere@foss.st.com>`.

etienne-lms:
@mzachar, could you add the [review tag i posted](https://github.com/OP-TEE/optee_client/pull/394#pullrequestreview-2470127143) into you commit message so that @jforissier can merge your change?

PR #397, opened 2025-04-09
cmake: raise mimimum required to 3.5

daniel-lang:
[CMake 4.0 was released](https://www.kitware.com/cmake-4-0-0-available-for-download/) which dropped compatibility with versions older than 3.5.
Bump the minimum required version to 3.5, which was released 9 years ago.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.5)
 project(optee_client C)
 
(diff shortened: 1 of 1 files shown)


jforissier:
> [CMake 4.0 was released](https://www.kitware.com/cmake-4-0-0-available-for-download/) which dropped compatibility with versions older than 3.5

...and causes `cmake_minimum_required(3.4)` to fail.

> . Bump the minimum required version to 3.5, which was released 9 years ago.

With the above added to the commit description:

Thanks!

daniel-lang:
I have updated the commit message and added your Reviewed-by tag

PR #407, opened 2025-12-02
tee-supplicant: add details for systemd startup procedure

h-assmann:
The current implementation lacks some details in order to work properly within the context of udev and systemd. The following patches aim to add these.


jforissier:
For "tee-supplicant: add priority prefix for udev rule":

jforissier:
For "tee-supplicant: add device instance as start parameter for service file":

h-assmann:
Thank you; ack-tags are applied now.

PR #409, opened 2025-12-04
tee-supplicant: fix uninitialized data access if stat() fails

jnippula:
do_mkdir() shall not continue to S_ISDIR check with 'st' struct fields in case the stat() call has returned error, because the struct content may be uninitialized.


jforissier:

Thanks!

jnippula:
Are there any issues preventing to merge this?

jnippula:
..after the updated "Reviewed-by" trailer?

jenswi-linaro:
All is good, I'm merging.

PR #411, opened 2025-12-17
ftrace: Fix not to write duplicate header when append to existing ftrace file

Leo-mediatek:
Currently, tee-supplicant appends to an existing prof file if OP-TEE provides a non-zero id in the rpc call parameter. However, for ftrace data, there is always a header in the buffer.

This patch makes tee-supplicant not to write duplicate header by locating the ftrace magic and discarding data before it when tee-supplicant receives a non-zero ftrace id.

This is required for https://github.com/OP-TEE/optee_os/pull/7657 to produce correct ftrace data.

### DIFF
--- a/tee-supplicant/src/prof.c
+++ b/tee-supplicant/src/prof.c
@@ -25,9 +25,11 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <assert.h>
 #include <errno.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
@@ -46,10 +48,11 @@ TEEC_Result prof_process(size_t num_params, struct tee_ioctl_param *params,
 {
 	char vers[5] = "";
 	char path[255] = { 0 };
+	const char ftrace_magic[] = { 'F', 'T', 'R', 'A', 'C', 'E', 0x00, 0x01 };
 	size_t bufsize = 0;
 	TEEC_UUID *u = NULL;
 	int fd = -1;
-	void *buf = NULL;
+	uint8_t *buf = NULL;
 	int flags = 0;
 	int id = 0;
 	int st = 0;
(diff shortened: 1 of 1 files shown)


jforissier:
This introduces some knowledge on the ftrace format into tee-supplicant (the "FTRACE\0\1" magic). I'd like to avoid that is possible. Could OP-TEE OS be updated to avoid sending the magic twice instead?

Leo-mediatek:
I cannot remember why I chose to remove the ftrace header here.

I think we can make the header not to be copied to the tee-supplicant by adding an if at optee_os to check whether `fbuf->dump_id == 0`.

https://github.com/OP-TEE/optee_os/blob/ce964642bf4580d8ab1914b6c4d4bcf7bd055928/ldelf/ftrace.c#L109-L110

jforissier:
> I cannot remember why I chose to remove the ftrace header here.

😁 
 
> I think we can make the header not to be copied to the tee-supplicant by adding an if at optee_os to check whether `fbuf->dump_id == 0`.
> 
> https://github.com/OP-TEE/optee_os/blob/ce964642bf4580d8ab1914b6c4d4bcf7bd055928/ldelf/ftrace.c#L109-L110

It should work, yes. Can you please try it and push an update to https://github.com/OP-TEE/optee_os/pull/7657 if it works well? Thanks!

Leo-mediatek:
I remembered that why I handle the header in tee-supplicant.

Since the header is written at many location:

- https://github.com/OP-TEE/optee_os/blob/2ff02bfa78ff26550259ddccc4d968f75aa00897/core/kernel/user_ta.c#L328
- https://github.com/OP-TEE/optee_os/blob/2ff02bfa78ff26550259ddccc4d968f75aa00897/ldelf/main.c#L106
- https://github.com/OP-TEE/optee_os/blob/2ff02bfa78ff26550259ddccc4d968f75aa00897/ldelf/ftrace.c#L109-L110

Therefore, handle this in tee-supplicant is easier. But I think it is ok to add an if statement to all of them.

jforissier:
> Therefore, handle this in tee-supplicant is easier. But I think it is ok to add an if statement to all of them.

Yes please fix in optee_os, it makes more sense.

Leo-mediatek:
Okay, I will try to modify that in optee_os and test if the ftrace output is correct.

Leo-mediatek:
I tested handling the header in optee_os and it works as expected now.

I think this PR is no longer needed.

PR #414, opened 2026-04-23
tee-supplicant: fix default udev path

jessetdy:
When an installation prefix is present e.g. /usr the install location will end up being /usr/etc/udev/rules.d instead of /etc/udev/rules.d When using CMAKE_INSTALL_FULL_SYSCONFDIR the installation prefix is ignored so it is installed to the correct location.

### DIFF
--- a/tee-supplicant/CMakeLists.txt
+++ b/tee-supplicant/CMakeLists.txt
@@ -143,7 +143,16 @@ endif()
 
 # Some sane defaults if discovering through pkgconfig fails or is disabled.
 set(SYSTEMD_UNIT_DIR "${CMAKE_INSTALL_LIBDIR}/systemd/system" CACHE PATH "Location of systemd unit files.")
-set(UDEV_UDEV_DIR "${CMAKE_INSTALL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
+
+if (POLICY CMP0192)
+	cmake_policy(GET CMP0192 _cmp0192)
+endif()
+
+if (_cmp0192 STREQUAL "NEW")
+	set(UDEV_UDEV_DIR "${CMAKE_INSTALL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
+else()
+	set(UDEV_UDEV_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
+endif()
 
 install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION ${CMAKE_INSTALL_SBINDIR})
 if (CFG_ENABLE_SYSTEMD)


tee-supplicant/CMakeLists.txt
```diff
@@ -143,7 +143,7 @@ endif()
 
 # Some sane defaults if discovering through pkgconfig fails or is disabled.
 set(SYSTEMD_UNIT_DIR "${CMAKE_INSTALL_LIBDIR}/systemd/system" CACHE PATH "Location of systemd unit files.")
-set(UDEV_UDEV_DIR "${CMAKE_INSTALL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
+set(UDEV_UDEV_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
```
jforissier: It may not be the proper fix because recent versions of CMake (>= 4.1) should work fine with `CMAKE_INSTALL_SYSCONFDIR` while `CMAKE_INSTALL_FULL_SYSCONFDIR` would not be recommended (policy CMP0192). Can you please check this? Thanks.

jessetdy: Hello Jerome,

I did check this, and you're right that `OLD` behavior is deprecated by definition, but the only way to make this work would be if you're able to enforce cmake minimum version 4.1 or set the policy, which is also only present from 4.1.

CMake does not tend to deprecate these things quickly hence I figured this would do for the time being, but if desired I could refactor this to perform a policy check and work from there

```
if (POLICY CMP0192)
    set(UDEV_UDEV_DIR "${CMAKE_INSTALL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
else()
    set(UDEV_UDEV_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
endif()
```

@jforissier I've now applied this change, it does seem to make most sense to introduce this policy check, as I didn't really find any other way around this if we want to use non-deprecated behavior from CMake 4.1+

jforissier: LGTM except you should check if the policy is NEW
```
if (POLICY CMP0192)
    cmake_policy(GET CMP0192 _cmp0192)
endif()

if (_cmp0192 STREQUAL "NEW")
    set(UDEV_UDEV_DIR "${CMAKE_INSTALL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
else()
    set(UDEV_UDEV_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}/udev/rules.d" CACHE PATH "Location of udev files.")
endif()
```

jessetdy: Huh right, I had misunderstood the if POLICY check, thanks for that :+1: 

Changes have been applied.

jforissier:
Please set the subject to `tee-supplicant: fix default udev path` and update the description to mention the behavior change with CMP0192 and CMake 4.1. With that:

`Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

jessetdy:
> Please set the subject to `tee-supplicant: fix default udev path` and update the description to mention the behavior change with CMP0192 and CMake 4.1. With that:
> 
> `Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

Reviewed-by added and commit message updated to mention the exact cause and usage of policy check.

Thank you very much Jerome for the speedy review.

PR #416, opened 2026-05-13
libckteec: fix integer overflow in ck_slot_get_list and ck_token_mechanism_ids

karnakarreddi:
*count value is multiplied by sizeof(uint32_t) to compute the size of a shared memory allocation. On 64-bit systems a *count of 0x4000000000000000 (or 0x40000000 on 32-bit) wraps this multiplication to zero, so ckteec_alloc_shm() ends up allocating an 8-byte buffer (the minimum that TEEC_AllocateSharedMemory substitutes for size 0). The subsequent loop then reads past the end of the shared memory buffer into adjacent heap memory.

Guard against the overflow by rejecting any *count that would make the size_t multiplication wrap, returning CKR_HOST_MEMORY in that case.

Reported as an integer overflow (CWE-190) leading to an out-of-bounds read (CWE-125).

### DIFF
--- a/libckteec/src/local_utils.h
+++ b/libckteec/src/local_utils.h
@@ -13,4 +13,190 @@
 		switch (0) { case 0: case ((x) ? 1: 0) : default : break; } \
 	} while (0)
 
+/*
+ * Checking overflow for addition, subtraction and multiplication. Result
+ * of operation is stored in res which is a pointer to some kind of
+ * integer.
+ *
+ * The macros return true if an overflow occurred and *res is undefined.
+ */
+#define ADD_OVERFLOW(a, b, res) __compiler_add_overflow((a), (b), (res))
+#define SUB_OVERFLOW(a, b, res) __compiler_sub_overflow((a), (b), (res))
+#define MUL_OVERFLOW(a, b, res) __compiler_mul_overflow((a), (b), (res))
+
+#define __GCC_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + \
+		       __GNUC_PATCHLEVEL__)
+
+#if __GCC_VERSION >= 50100 && !defined(__CHECKER__)
+#define __HAVE_BUILTIN_OVERFLOW 1
+#endif
+
+#if __GCC_VERSION >= 90100 && !defined(__CHECKER__)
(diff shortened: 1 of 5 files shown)


karnakarreddi:
@jenswi-linaro created this PR as per our discussion.

libckteec/src/local_utils.h
```diff
@@ -13,4 +13,190 @@
...
+/*
+ * Checking overflow for addition, subtraction and multiplication. Result
+ * of operation is stored in res which is a pointer to some kind of
+ * integer.
+ *
+ * The macros return true if an overflow occurred and *res is undefined.
+ */
+#define ADD_OVERFLOW(a, b, res) __compiler_add_overflow((a), (b), (res))
+#define SUB_OVERFLOW(a, b, res) __compiler_sub_overflow((a), (b), (res))
+#define MUL_OVERFLOW(a, b, res) __compiler_mul_overflow((a), (b), (res))
```
etienne-lms, line 25: IMHO, adding these helper macros would deserve a specific commit that says where they come from.

libckteec/src/pkcs11_processing.c
```diff
@@ -3,6 +3,7 @@
  * Copyright (c) 2017-2018, Linaro Limited
  */
 
+#include <limits.h>
```
etienne-lms, line 6: Can be removed, not needed.

libckteec/src/pkcs11_processing.c
```diff
@@ -1704,4 +1762,4 @@ CK_RV ck_unwrap_key(CK_SESSION_HANDLE session, CK_MECHANISM_PTR mechanism,
 	release_serial_object(&smecha);
 
 	return rv;
-}
+}
```
etienne-lms, line 1765: Could you discard this change. Note that it's likely done by your editor.

etienne-lms (commented):
The change in C source files could be squashed in a single commit. Preferably use a specific 1st commit for adding `*_OVERFLOW()` macros in **local_utils.h**.

Could you refine the commit messages? `libckteec: ` prefix in the header line and `Signed-off-by` tags at the end of the message body.
```
libckteec: <short-description>

<description>

Signed-off-by: Real Name <email@address>
```

karnakarreddi:
@etienne-lms created a clean PR https://github.com/OP-TEE/optee_client/pull/417

PR #417, opened 2026-05-14
libckteec: prevent integer overflow in shared memory allocations

karnakarreddi:
Fixes integer overflow in several libckteec callers of ckteec_alloc_shm() and in the PKCS#11 serializer.

### DIFF
--- a/libckteec/src/local_utils.h
+++ b/libckteec/src/local_utils.h
@@ -13,4 +13,190 @@
 		switch (0) { case 0: case ((x) ? 1: 0) : default : break; } \
 	} while (0)
 
+/*
+ * Checking overflow for addition, subtraction and multiplication. Result
+ * of operation is stored in res which is a pointer to some kind of
+ * integer.
+ *
+ * The macros return true if an overflow occurred and *res is undefined.
+ */
+#define ADD_OVERFLOW(a, b, res) __compiler_add_overflow((a), (b), (res))
+#define SUB_OVERFLOW(a, b, res) __compiler_sub_overflow((a), (b), (res))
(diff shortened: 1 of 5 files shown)


etienne-lms (commented):
For commit "libckteec: add {ADD|SUB|MUL}_OVERFLOW() helper macros":
* Could break the message message into several lines, 80 char/line max IIR?
(ditto for the 2nd commit)
* Could you add a reference from where the macros implementation were fetched from, e.g. OP-TEE OS Github URLs? It would be easier for one to find the source.
* With `Co-developed-by`, you need to also provide the co-developper `Signed-off-by` tag. Alternatively use `Co-authored-by`, that do not require an added S-o-b tag.

karnakarreddi:
> For commit "libckteec: add {ADD|SUB|MUL}_OVERFLOW() helper macros":
> 
> * Could break the message message into several lines, 80 char/line max IIR?
>   (ditto for the 2nd commit)
> * Could you add a reference from where the macros implementation were fetched from, e.g. OP-TEE OS Github URLs? It would be easier for one to find the source.
> * With `Co-developed-by`, you need to also provide the co-developper `Signed-off-by` tag. Alternatively use `Co-authored-by`, that do not require an added S-o-b tag.

Thanks for the review, @etienne-lms. I've addressed all three points and force-pushed the updated branch. Please take another look when you have a moment.

jenswikl:
@etienne-lms, are you OK with this PR?

etienne-lms:
@karnakarreddi. Your changes look good to me but need to be rebased. Could you update your P-R, I'll review it afterward, eventually posting my review tag.

PR #419, opened 2026-05-27
libckteec: fix NULL dereferences, overflow, and thread-safety in PKCS#11 serialization

manoj23:
This series fixes several issues in libckteec found by fuzzing (AFL++ with AddressSanitizer) and static analysis (cppcheck, clang-tidy):

1. **NULL pointer dereferences** in mechanism serialization functions -- `serialize_mecha_aes_ctr()`, `serialize_mecha_aes_gcm()`, `serialize_mecha_aes_cbc_encrypt_data()`, `serialize_mecha_eddsa()`, and `serialize_ck_mecha_params()` dereference `mecha->pParameter` without NULL checks. (CWE-476)

2. **Missing data pointer validation in `serialize()`** -- `memcpy(buf + *blen, data, len)` is called without validating `data`, causing invalid memory access when a caller passes a NULL or wild pointer with non-zero `len`.

3. **Integer overflow in `serialize_mecha_aes_cbc_encrypt_data()`** -- `param->length` (CK_ULONG, 64-bit on LP64) is added into `uint32_t size` without bounds checking. Values above UINT32_MAX silently truncate, producing an undersized buffer allocation. (CWE-190 leading to CWE-122)

4. **Redundant return value** in `serialize_mecha_aes_ctr()` -- the final `return rv` is reached only after early-exit guards guarantee `rv == CKR_OK`. Return the constant directly.

5. **Thread-unsafe `strerror()`** in `invoke_ta.c` -- replaced with `strerror_r()` since libckteec uses pthread_mutex for session management.

Tested with `xtest -t pkcs11`: 31 test cases, 7682 subtests, 0 failures.

### DIFF
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -559,6 +559,9 @@ static CK_RV serialize_mecha_aes_cbc_encrypt_data(struct serializer *obj,
 	if (!param)
 		return CKR_MECHANISM_PARAM_INVALID;
 
+	if (param->length > UINT32_MAX - sizeof(param->iv) - sizeof(uint32_t))
+		return CKR_MECHANISM_PARAM_INVALID;
+
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
 		return rv;
--- a/libckteec/src/serialize_ck.c
+++ b/libckteec/src/serialize_ck.c
@@ -393,6 +393,9 @@ static CK_RV serialize_mecha_aes_ctr(struct serializer *obj,
 	CK_RV rv = CKR_GENERAL_ERROR;
 	uint32_t size = 0;
 
+	if (!param)
+		return CKR_MECHANISM_PARAM_INVALID;
+
 	rv = serialize_32b(obj, obj->type);
 	if (rv)
 		return rv;
@@ -419,6 +422,9 @@ static CK_RV serialize_mecha_aes_gcm(struct serializer *obj,
 	CK_GCM_PARAMS_PTR param = mecha->pParameter;
 	CK_RV rv = CKR_GENERAL_ERROR;
 	CK_ULONG aad_len = 0;
+
+	if (!param)
+		return CKR_MECHANISM_PARAM_INVALID;
 
 	/* AAD is not manadatory */
 	if (param->pAAD)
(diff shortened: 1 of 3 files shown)


libckteec/src/serialize_ck.c
```diff
@@ -810,8 +822,45 @@ CK_RV serialize_ck_mecha_params(struct serializer *obj,
 	case CKM_AES_CBC:
 	case CKM_AES_CBC_PAD:
 	case CKM_AES_CTS:
+		if (!mecha.pParameter)
+			return CKR_MECHANISM_PARAM_INVALID;
 		return serialize_mecha_aes_iv(obj, &mecha);
 
+	/*
```
jenswikl: The changes in this function `serialize_ck_mecha_params()` are a bit inconsistent. For some values of `mecha.mechanism`, `mecha.pParameter`:
- is checked here and checked again in the next level serializer
- is checked here only
- is only checked at the next level serializer

Since this function is quite long, perhaps it's best to focus on routing to the next function here and do the parameter checking at the level that actually uses the parameter.

libckteec/src/serializer.c
```diff
@@ -44,8 +44,17 @@ void release_serial_object(struct serializer *obj)
...
+	size_t nlen = 0;
+	char *buf = NULL;
+
+	if (len && !data)
+		return CKR_ARGUMENTS_BAD;
+
+	if (!len)
+		return CKR_OK;
+
+	nlen = *blen + len;
```
jenswikl, line 60: How about checking for integer overflow while we're at it?

libckteec/src/serialize_ck.c
```diff
@@ -413,7 +413,7 @@ static CK_RV serialize_mecha_aes_ctr(struct serializer *obj,
 	if (rv)
 		return rv;
 
-	return rv;
+	return CKR_OK;
```
jenswikl: Same here as in https://github.com/OP-TEE/optee_os/pull/7824#discussion_r3312226847

manoj23: Ack, I'll discard this change.

jenswikl:
Looks good to me. Please apply:

manoj23:
Hello, I added `Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>`.

libckteec/src/serialize_ck.c
manoj23: Addressed

PR #421, opened 2026-07-03
libckteec: add and use {ADD|SUB|MUL}_OVERFLOW() macros

etienne-lms:
Add `ADD_OVERFLOW()`, `SUB_OVERFLOW()` and `MUL_OVERFLOW()` helper macros dumped from optee_os util.h [1] and compiler.h [2] and use them to guard against index/buffer overflows. 

[1] https://github.com/OP-TEE/optee_os/blob/4.10.0/lib/libutils/ext/include/util.h
[2] https://github.com/OP-TEE/optee_os/blob/4.10.0/lib/libutils/ext/include/compiler.h


libckteec/src/pkcs11_processing.c
```diff
@@ -3,6 +3,7 @@
  * Copyright (c) 2017-2018, Linaro Limited
  */
 
+#include <limits.h>
```
jforissier, line 6: I believe `<stdint.h>` should be included in `local_utils.h` instead

jforissier (commented):
Hi @etienne-lms,

Please see one comment below, and also:
s/Guard against overflow in when serializing/Guard against overflow when serializing/

With that fixed:

`Reviewed-by: Jerome Forissier <jerome.forissier@arm.com>`

Thanks!

libckteec/src/pkcs11_processing.c
etienne-lms: Would make sense.
That said, `limits.h` is not needed here, can be removed.

libckteec/src/pkcs11_processing.c
```diff
@@ -36,7 +38,9 @@ CK_RV ck_create_object(CK_SESSION_HANDLE session, CK_ATTRIBUTE_PTR attribs,
 		goto out;
 
 	/* Shm io0: (i/o) [session-handle][serialized-attributes] / [status] */
-	ctrl_size = sizeof(session_handle) + obj.size;
+	if (ADD_OVERFLOW(sizeof(session_handle), obj.size, &ctrl_size))
+               return CKR_ARGUMENTS_BAD;
```
etienne-lms, line 42: My mistake. `serialize_ck_attributes()` ( a few lines above) allocates memory that should be free before returning from this function. `goto out` would better apply here.

Note: I missed https://github.com/OP-TEE/optee_client/pull/417 that does the jobs nicely. I'll close my P-R.

PR #424, opened 2026-09-23
tee-supplicant: drop revoked RPC responses instead of exiting

### LINKED ISSUE #423: tee-supplicant should handle -EBADF from TEE_IOC_SUPPL_SEND without exiting
**Background**
Commit 70b0d6b0a199 ("tee: optee: Fix supplicant wait loop") changed the client wait to killable, allowing the client to terminate while the supplicant is still processing its request. A later fix (724d0caffd42) marks revoked requests as invalid; optee_supp_send() now returns -EBADF when the supplicant tries to send a response for a request whose client is gone.

**Kernel-side decision**
Jens Wiklander summarized the trade-off during review:
> To fix the possible use-after-free, we have two options:
> 1. Returning an error code: tee-supplicant will exit
> 2. Returning OK: worst case, tee-supplicant can leak memory
>
> I think we should do what you suggested and return an error. This will not happen during normal operation. We'll fix tee-supplicant to handle the return error properly. If there's an error in TEE_IOC_SUPPL_SEND, it knows that no one will receive whatever was sent, and cleanup is needed.

The kernel side chose to return -EBADF and expects tee-supplicant to handle that error path, including cleanup.

**Problem**
tee-supplicant currently treats any error from TEE_IOC_SUPPL_SEND as fatal and exits. It does not distinguish -EBADF, which the kernel uses specifically to signal a revoked request. As a result, a revoked request — expected during shutdown or when a privileged process kills the client — causes the supplicant to exit instead of simply discarding the response and cleaning up.

**Expected behavior**
When TEE_IOC_SUPPL_SEND returns -EBADF, tee-supplicant should:
1. Recognize that the request has been revoked (the client is gone; no one will receive the response).
2. Clean up all supplicant-side resources allocated for that request (buffers, shared memory, per-request state, etc.).
3. Not exit — continue serving other requests.

For any other error, the existing fatal behavior can remain.

**References**
Kernel commit: 724d0caffd42 ("tee: optee: prevent use-after-free when the client exits before the supplicant")
Mailing list discussion: Jens Wiklander's review comments on the v3 patch，[https://patchew.org/linux/20260128-fix-use-after-free-v3-1-b0786670d927@oss.qualcomm.com/](url)

zhangxiaohui1:
The kernel deliberately fails TEE_IOC_SUPPL_SEND with EBADF when the client that made the request exited while the supplicant was still processing it.  optee_supp_thrd_req() then replaces the request ID with INVALID_REQ_PTR (ERR_PTR(-EBADF)) so that the ID stays busy and a late response cannot be delivered to another client, and supp_pop_req() resets the kernel side bookkeeping (idr_remove() and req_id = -1) before returning the sentinel.

Dropping the response is therefore safe: it is a single-request event and the kernel has already re-synchronized.

write_response() currently treats every ioctl failure as fatal, so one client being killed (watchdog, killall, restart) tears down the whole tee-supplicant and aborts every other client's in-flight request with TEEC_ERROR_COMMUNICATION.  Distinguish EBADF and keep serving.

A dropped OPTEE_MSG_RPC_CMD_SHM_ALLOC response would however leak the shared memory allocated for it: cmd_alloc_suppl() only calls tee_shm_get_from_id() when optee_supp_thrd_req() succeeded, so the kernel never learns the ID and no SHM_FREE will ever arrive.  Reclaim it in process_one_request() when the response is dropped.

Closes #423


jenswikl:
`Reviewed-by: Jens Wiklander <jens.wiklander@oss.qualcomm.com>`

